PR #5, opened 2018-04-17
Cleaned up buttons and added colors

jake-robertson:
I brought the button styles up to spec by:
*adding `minor` variation
*changing font size per size
*adding simple transtions to all buttons
*matching padding to style guide

To support these changes I added our new color swatches and a missing thickness,`thickness-35`, to the shared-styles .less files.

I also added a folder and .md file for Icon buttons to preserve some documentation that no longer made sense in the main buttons .md file.


dustinsoftware:
Merged

PR #6, opened 2018-04-17
Switch to styled components

dustinsoftware:
TODO

- [x] Investigate using `babel-plugin-styled-components`
- [x] Ensure server-render still works
- [x] Fix stylelint configuration
- [x] Update docs


components/check-box/component.jsx
```diff
@@ -1,20 +1,20 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { forbidExtraProps } from 'airbnb-prop-types';
-import { themeClassNames } from '../utils';
-import baseTheme from './base-theme.less';
+import { ThemeProvider } from 'styled-components';
```
Imundy: Just kind of thinking out loud here, but it looks like styled-components is using the former version of the context API, so we'll probably just have to keep an eye on what version of React these components support if that changes. I see it's 16.3.1 right now so we should be safe since I don't think they have plans to remove the old context API any time soon.

stories/demo-button/story.jsx
```diff
@@ -12,51 +11,56 @@ export default class DemoContainer extends Component {
...
 
 	render() {
+		const { variations } = Button;
 		return (
-			<div className={styles.demos}>
+			<Styled.Demos>
 				{this.renderButtonsRow(marginVariation => (
 					<Button
-						onClick={this.props.onClick}
+						buttonProps={{ onClick: this.props.onClick }}
```
Imundy: This seems kind of unfortunate, but I don't really see a good way around it. I wonder if inverting is so that we passed `styleProps` would be a cleaner API? I imagine styleProps would be pretty similar between all the components and then we could do something like:

dustinsoftware: Yeah I'm not super happy with this approach but couldn't think of anything better. There are some downsides to using `...buttonProps` that I can think of, which led me to the current pattern:

- There is no distinction between props that we define, such as a custom `onChange`, and the props that need to be passed to the underlying control. For a more complicated example of this, see [demo-input](https://github.com/dustinsoftware/engagement-products-ui/blob/ebbb9273f7ca75b189a2a4831d46b0934f3d3167/components/demo-input/component.jsx#L11-L24)
- There are attributes on elements that we may be unintentionally overloading (`onChange` is one example)
- There is no support for warning if extra props are specified to the component

Imundy: 👍 I suppose that as long as the props of the underlying component are well defined, that's all that really matters here.

dustinsoftware: I've come around on this, most React UI libraries follow the approach you described, so I'm going to punt on this issue. If we need to deviate from that pattern, we can do that down the road.

PR #7, opened 2018-05-01
Updated Universal Styles

jake-robertson:
renamed existing thicknesses and colors, added missing styles.

Made a couple changes to the story documentation to make it more true :)


jake-robertson:
@dustinsoftware I just noticed this was never merged, which is making styling components difficult now.

PR #8, opened 2018-05-30
Improve validated input control

dustinsoftware:
The original implementation was too confusing to follow, this PR splits the validation logic into a separate component, which uses a render prop to render the styled input.


components/text-input/input-validation.jsx
```diff
@@ -0,0 +1,107 @@
...
+export default class InputValidation extends Component {
+	static propTypes = forbidExtraProps({
+		getIsValidInput: PropTypes.func,
+		help: PropTypes.node,
+		isRequiredField: PropTypes.bool,
+		onChange: PropTypes.func.isRequired,
+		onEnter: PropTypes.func,
+		renderInput: PropTypes.func.isRequired,
+		validationDelay: PropTypes.number,
+		validationErrorString: PropTypes.string,
```
Imundy: So, if I had an input that had a few possible error strings (e.g. too long vs invalid), would the expectation be that when `getIsValidInput` is called, the parent component would also update the props to pass the desired `validationErrorString`? I just want to make sure that wouldn't cause any weird flashes of content.

dustinsoftware: The original idea was that there's only one error message that the control supports, which is silly in hindsight. I'll update the API.

PR #11, opened 2018-06-21
Add modal component from giving

Imundy:


Imundy:
@dustinsoftware I was thinking about sensible/easy defaults for this lat night. What do you think of this for the modal footer:
Provide 2 props
* footerProps - PropTypes.shape({
  onCommit: PropTypes.func,
  onCancel: PropTypes.func,
  onDelete: PropTypes.func,
  commitText: PropTypes.string,
  cancelText: PropTypes.string,
  deleteText: PropTypes.string
})
* renderFooter - PropTypes.func

That way we provide a simple/easy to use interface to match the specs and provide a way to do "non-standard" footers.

Imundy:
Also should say that isn't fully my idea 😅 . It was inspired by a twitter thread I saw Ryan Florence (I think?) weigh in on.

Imundy:
I changed the prop stucture example to what I described. The case this still doesn't handle is stacking the buttons when the modal gets too small (but fwiw the reactstrap also doesn't yet do that).

components/modal/default-modal-footer.jsx
```diff
@@ -0,0 +1,61 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Button from '../demo-button/component.jsx';
+import { colors } from '../shared-styles';
+import ModalFooter from './modal-footer.jsx';
+import * as Styled from './styled.jsx';
+
+const DefaultModalFooter = ({ commit, cancel, delete: deletion }) => {
```
Imundy: `delete` is a keyword so maybe I shouldn't be using it as a prop. I couldn't think of another good name though so I'm open to suggestions.

Imundy: Maybe postifx all of them with `Button`? I think it would make sense to pass down optional button props in these objects as well so that consumers could control the color.

JordanSjodinFaithlife:
We already have the modal from Bootstrap. Is this one just better/nicer?

Imundy:
@JordanSjodinFaithlife 
2 reasons I PR'ed this:
* This is the modal from Faithlife Giving. So this was a good learning experience for me in working with `styled-components` and porting existing UI
* The bootstrap modal doesn't match our modal spec in zeplin which this one is striving to be closer to.

But using this as a learning experience was, I think, the most important part for me :)

components/utils/index.js
```diff
@@ -11,3 +11,12 @@ export function applyVariations(component, variationMap, props) {
 
 	return { component: wrappedComponent, filteredProps };
 }
+
+export const debouncedResize = callback => {
+	let frame;
+	return window.addEventListener('resize', () => {
```
ianfisk: I'm not in love with the fact that consumers of `debouncedResize` need to know to call `window.removeEventListener` like so:
```
componentDidMount() {
	this.resizeListener = debouncedResize(this.handleResize);
}

componentWillUnmount() {
	window.removeEventListener('resize', this.resizeListener);
}
```
I think it would be better if consumers of `debouncedResize` could call some `cancel` method on whatever is returned, and the cancel method could do the appropriate thing.

Aside from that, `addEventListener` actually returns `undefined` ([docs](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener)), so `window.removeEventListener('resize', this.resizeListener);` won't actually remove the listener.

Imundy: That's a good point. I could see a better implementation being
```js
export const debouncedResize = callback => {
	let frame;
	const listener = () => {
		if (frame) window.cancelAnimationFrame(frame);
		
		frame = window.requestAnimationFrame(callback);
	});

	window.addEventListener('resize', listener);

	return { 
		cancel: () => window.removeListener('resize', listener)
	}
};
```

How does that look to you @ianfisk?

PR #12, opened 2018-06-22
 Forward refs to underlying reactstrap components.

ianfisk:
This PR updates `styled-components` to version 3.2.6, which is when support for `React.forwardRef` was added (https://github.com/styled-components/styled-components/pull/1658).

Also, note this PR forwards refs to the `innerRef` prop when the component is a styled component. This prevents consumers from knowing the exported components are styled components, and enables code like this to "just work":
```
<Typeahead
	onChange={value => { setState({ selection: value })}}
	options={['Washington','California','Texas']}
	placeholder="Choose a state..."
	ref={e => {
		console.log('Typeahead ref:', e);
	}}
/>
```
Without the ref forwarding, consumers would need something like
```
<Typeahead
	onChange={value => { setState({ selection: value })}}
	options={['Washington','California','Texas']}
	placeholder="Choose a state..."
	innerRef={e => {
		console.log('Typeahead ref:', e);
	}}
/>
```
to get a ref to the underlying `TypeAhead` component.

### DIFF
--- a/catalog/bootstrap/components.md
+++ b/catalog/bootstrap/components.md
@@ -164,7 +164,7 @@ showSource: false
 ```
 
 ```react
-showSource: flase
+showSource: false
 ---
 <Row>
 	<div className="text-center">0%</div>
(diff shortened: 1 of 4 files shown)


ianfisk:
Actually, I noticed in `reactstrap`'s source code that they [forward the `innerRef` prop](https://github.com/reactstrap/reactstrap/blob/e6a131304236a301b95338b8c228623c9021d34c/src/Button.js#L76) just like styled-components, so maybe we don't need this PR since consumers can use `innerRef={e => { this.ref = e; }}` to get refs.

PR #14, opened 2018-07-24
Group selector

michael-gamlem-faithlife:
This component was based on the one that is currently deployed on Faithlife Sites

This component was written in Faithlife Music and then ported to styled-ui.  To see the full history of commits, go to [Faithlife Music PR 13](https://git/Logos/FaithlifeMusic/pull/13), [Faithlife Music PR 14](https://git/Logos/FaithlifeMusic/pull/14), and [Faithlife Music PR 16](https://git/Logos/FaithlifeMusic/pull/16).
### Screen Shots
screen shot 2018-07-24 at 11 21 36 am
screen shot 2018-07-24 at 11 21 29 am
screen shot 2018-07-24 at 11 20 59 am
screen shot 2018-07-24 at 11 21 12 am


catalog/group-selector/documentation.md
```diff
@@ -0,0 +1,7 @@
+This documentation is automatically generated from jsdoc comments.
+
+```react
+noSource: true
+---
+<DocgenTable component={GroupSelector} />
+```
```
michael-gamlem-faithlife: I am wondering if there is anything else that should be documented here to make the component easier to use

components/group-selector/component.jsx
```diff
@@ -0,0 +1,144 @@
...
+		// groups: PropTypes.arrayOf(
+		// 	PropTypes.shape({
+		// 		name: PropTypes.string,
+		// 		groupId: PropTypes.number,
+		// 		kind: PropTypes.string,
+		// 		avatarUrl: PropTypes.string,
+		// 		membershipKind: PropTypes.string,
+		// 	}),
+		// ).isRequired,
+		groups: PropTypes.array.isRequired,
```
michael-gamlem-faithlife: I simplified to `groups: PropTypes.array.isRequired` because I was receiving linter errors.  However, if the shape of the array is not like what I originally defined, the component will either error or operate incorrectly.  Is there a better way to use `PropTypes` to validate the shape when the props may not be explicitly used in this file?

components/group-selector/icons/church-group-kinds-icons.g.jsx
```diff
@@ -0,0 +1,869 @@
+import { createElement } from 'react';
```
michael-gamlem-faithlife: This file was taken from Faithlife Sites.  I did not generate it.

components/group-selector/icons/index.js
```diff
@@ -0,0 +1,60 @@
...
+	RadioStation,
+	School,
+	SmallGroup,
+	SpecialInterest,
+	SportsTeam,
+	WorshipTeam,
+	YouthGroup,
+} from './church-group-kinds-icons.g.jsx';
+
+/* eslint-disable */
```
michael-gamlem-faithlife: Is it appropriate to disable `eslint` here to prevent it from displaying an error about default exports?

components/group-selector/modal/search-result.jsx
```diff
@@ -0,0 +1,147 @@
...
+		handleGetStartedClick: PropTypes.func.isRequired,
+		handleRequestClick: PropTypes.func.isRequired,
+		handleEditClick: PropTypes.func.isRequired,
+		handleJoinGroupClick: PropTypes.func.isRequired,
+	};
+
+	state = {
+		message: <p />,
+		membershipLine: <p />,
+		button: <p />,
```
ianfisk: Although possible to store React elements in state, I believe it is considered an anti-pattern. Can we store the message strings or some other state that can be used to determine what to render?

components/group-selector/modal/component.jsx
```diff
@@ -0,0 +1,284 @@
...
+												)}
+											</button>
+										</Styled.SearchInputAddon>
+									</Styled.SearchInputGroup>
+									<Styled.SearchResultsContainer>
+										{this.getSearchResults()}
+									</Styled.SearchResultsContainer>
+								</div>
+								<Styled.CreateGroupWrapper
+									property={this.state.isCreateGroupFocused ? '0' : '-109px'}
```
ianfisk: Wasn't the name of this prop going to change?

components/group-selector/modal/create-group.jsx
```diff
@@ -0,0 +1,76 @@
...
+		newGroupName: '',
+		newGroupLocation: '',
+	};
+
+	handleNewGroupNameChange = event => {
+		this.setState({ newGroupName: event.target.value });
+	};
+
+	handleCreateGroupInsideClick = () => {
+		this.props.setCreateGroupFocused(true);
```
ianfisk: Boolean parameters are notoriously hard to reason about at call sites (what does passing `true` here mean?). I kind of feel like this component is coupled in a weird way to the consuming component because of the `isCreateGroupFocused`, `setCreateGroupFocused`, and `searchInputValue` props. Once I solidify my thinking we can work together to decouple it.

components/group-selector/component.jsx
```diff
@@ -0,0 +1,144 @@
...
+		// searchedGroups: PropTypes.arrayOf(
+		// 	PropTypes.shape({
+		// 		name: PropTypes.string,
+		// 		groupId: PropTypes.number,
+		// 		kind: PropTypes.string,
+		// 		avatarUrl: PropTypes.string,
+		// 		membershipKind: PropTypes.string,
+		// 	}),
+		// ),
+		searchedGroups: PropTypes.array,
```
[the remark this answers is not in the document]
michael-gamlem-faithlife: Yeah, I haven't been pleased with the name `searchedGroups` either.  I like `groupSearchResults` better.  What about something like `returnedGroupSearchResults`?  Would that be better/too long?

components/group-selector/dropdown.jsx
```diff
@@ -0,0 +1,105 @@
...
+		selectedGroup: PropTypes.object.isRequired,
+		handleSelectionChange: PropTypes.func.isRequired,
+		handleFindChurchButtonClick: PropTypes.func.isRequired,
+	};
+
+	state = {
+		isDropdownOpen: false,
+	};
+
+	makeStoredIcon = group => {
```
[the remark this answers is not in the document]
michael-gamlem-faithlife: Ian and I worked through this in person.  We decided that it was best to factor this out into its own component that will handle all icons.

components/group-selector/dropdown.jsx
```diff
@@ -0,0 +1,105 @@
...
+	handleGroupSelection = groupId => {
+		this.props.handleSelectionChange(groupId);
+		this.setState({ isDropdownOpen: false });
+	};
+
+	render() {
+		const groups = this.props.groups.map(group => (
+			<SimpleGroup
+				key={group.groupId}
+				id={group.groupId}
```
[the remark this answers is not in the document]
michael-gamlem-faithlife: This is actually the `groupId` it is just poorly named.  I'll change it to `groupId`

components/group-selector/modal/component.jsx
```diff
@@ -0,0 +1,284 @@
...
+		searchInputValue: '',
+		isCreateGroupFocused: false,
+		modalContent: 'main',
+	};
+
+	clearSearchInput = () => {
+		this.setState({
+			searchInputValue: '',
+		});
+		this.executeSearch(this.state.searchInputValue);
```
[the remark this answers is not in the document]
michael-gamlem-faithlife: My intention was to call it with the empty string to force the search results to clear and have the component re-render with the groups of which the user is already a member.  There is probably a better way to do this.  Faithlife Music is currently returning the wrong data to the group selector in this case, so I am working through that issue as well.

components/group-selector/modal/component.jsx
```diff
@@ -0,0 +1,284 @@
...
+	};
+	makeStoredIcon = group => {
```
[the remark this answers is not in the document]
michael-gamlem-faithlife: See above comment about factoring this code out

components/group-selector/modal/component.jsx
```diff
@@ -0,0 +1,284 @@
...
+	handleJoinGroup = () => {
+		this.props.handleJoinGroupClick(this.state.selectedGroupId);
+	};
+
+	handleCancel = () => {
+		this.setState({ modalContent: 'main' });
+	};
+
+	render() {
+		return (
```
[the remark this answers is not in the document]
michael-gamlem-faithlife: I fixed this problem.  The reason that I did not use `less` was that there was no `less` being used in `styled-ui`.  Maybe this is a decision that @dustinsoftware has made?

components/group-selector/modal/search-result.jsx
michael-gamlem-faithlife: The way that the component is structured right now, I do not think that would be possible.  If we do not want to store elements in the state a different approach to creating `search-result`s would need to be taken.

components/group-selector/modal/create-group.jsx
michael-gamlem-faithlife: Ian and I spoke about this and I will be re-writing this as a functional component to fix this problem

dustinsoftware:
Thanks so much for the pull request. It might be a little bit before I can get to reviewing and merging. We also need a sample component checked in that mocks the different components states, see [carousel-demo](https://github.com/Faithlife/styled-ui/blob/master/catalog/bootstrap/carousel-demo.jsx) as an example.

components/group-selector/component.jsx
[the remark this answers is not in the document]
ianfisk: I like `groupSearchResults` more than `returnedGroupSearchResults`.

components/group-selector/modal/component.jsx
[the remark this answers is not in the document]
ianfisk: >My intention was to call it with the empty string

Ok, it is important to note that state may be updated asynchronously so you don't want to access `this.state` right after calling `this.setState`.

michael-gamlem-faithlife:
@ianfisk and I are wanting input on a discussion we had today...
Faithlife Music has a use case where we would like to show just the modal without the dropdown portion of the `GroupSelector` component.  This got us to thinking about if it still makes sense to ship the modal baked in with the `GroupSelector` component.  If we did so, then to accomplish what we want to do in Faithlife Music, we would need to somehow hide the dropdown portion and programmatically open the modal when we wanted.  This seems to be a hack and not desirable.
We are thinking that it makes sense to ship the dropdown and the modal as two separate components.  If this were to happen, the `Find or add church` button would require a function that the application would need to supply.  This could be a function to open the modal which is now a separate component.  This would free the modal to be placed both with the dropdown and on its own if desired.

components/group-selector/modal/component.jsx
```diff
@@ -0,0 +1,274 @@
...
+	redirectToGroup = () => {
+		this.setModalState('main');
+		window.open(`https://www.faithlife.com/${this.state.selectedGroupId}`, 'noopener, noreferrer');
+	};
+
+	handleSearchInput = event => {
+		this.setState({
+			searchInputValue: event.target.value,
+		});
+		this.props.executeSearch(this.state.searchInputValue);
```
ianfisk: We talked about this in person, but I am commenting here so it is not forgotten. This call to `this.props.executeSearch` needs to be debounced, and `event.target.value` should be used instead of `this.state.searchInputValue`. Using `searchInputValue` from state is why the first request has an empty query. Updating state is asynchronous so when this line executes `this.state.searchInputValue` is not the updated value (empty string initially).

ianfisk: Also, it would be great if the cursor is placed in the search input when the modal opens.

JordanSjodinFaithlife:
@dustinsoftware last I heard, Amber signed off on the design. Anything still blocking merging? Michael is gone now so I can fix anything that is left

catalog/group-selector/variations.md
```diff
@@ -0,0 +1,113 @@
+### Group Selector with Groups
+
+```react
+showSource: true
+state: {
+	groups: [{
+		name: Michael's Test Church,
```
dustinsoftware: This string is not quoted

catalog/group-selector/variations.md
```diff
@@ -0,0 +1,113 @@
...
+		onCreateGroup={() => {alert("handled by application")}}
+		onSelectionChange={() => {alert("handled by application")}}
+		groups={state.groups}
+		selectedGroupId={state.selectedGroupId}
+		groupSearchResults={state.groups}
+		groupSelectorView={"no-groups"}
+		onGetStartedClick={() => {alert("handled by application")}}
+		onClaimGroupClick={() => {alert("handled by application")}}
+		isMobile={false}
+		style={{zIndex:1000}}
```
dustinsoftware: Is this prop used? It's not documented

components/group-selector/component.jsx
```diff
@@ -0,0 +1,118 @@
...
+		/** returns data to application for group creation */
+		onCreateGroup: PropTypes.func.isRequired,
+		/** function that is called every time user chooses a different group */
+		onSelectionChange: PropTypes.func.isRequired,
+		/** returns search string to application */
+		executeSearch: PropTypes.func.isRequired,
+		/** -1 if no previous selected group */
+		selectedGroupId: PropTypes.number.isRequired,
+		/** all groups that should be displayed on the dropdown can be empty but must be defined */
+		// groups: PropTypes.arrayOf(
```
dustinsoftware: Commented out code

components/group-selector/component.jsx
```diff
@@ -0,0 +1,118 @@
...
+	groupId: -1,
+	kind: '',
+};
+
+/** Styled group selector control */
+export class GroupSelector extends React.Component {
+	static propTypes = {
+		/** Selects view state of group selector */
+		groupSelectorView: PropTypes.oneOf(['groups', 'no-groups']).isRequired,
+		/** returns data to application for group creation */
```
dustinsoftware: Seems like this should use sentence casing in the comment

components/group-selector/component.jsx
```diff
@@ -0,0 +1,118 @@
...
+		// 		kind: PropTypes.string,
+		// 		avatarUrl: PropTypes.string,
+		// 		membershipKind: PropTypes.string,
+		//		claimable: PropTypes.string,
+		//		location: PropTypes.string
+		// 	}),
+		// ).isRequired,
+		groups: PropTypes.array.isRequired,
+		/** undefined or empty if no search has been executed yet */
+		// searchedGroups: PropTypes.arrayOf(
```
dustinsoftware: Commented out code

components/group-selector/component.jsx
```diff
@@ -0,0 +1,118 @@
...
+export class GroupSelector extends React.Component {
+	static propTypes = {
+		/** Selects view state of group selector */
+		groupSelectorView: PropTypes.oneOf(['groups', 'no-groups']).isRequired,
+		/** returns data to application for group creation */
+		onCreateGroup: PropTypes.func.isRequired,
+		/** function that is called every time user chooses a different group */
+		onSelectionChange: PropTypes.func.isRequired,
+		/** returns search string to application */
+		executeSearch: PropTypes.func.isRequired,
```
dustinsoftware: `onSearchInputChange`? Since this is a callback prop and the consumer is expected to make the search, using `onFoo` naming makes it more clear. `handleSearch` could also work if there's more complicated logic involved beyond just an input changing.

components/group-selector/dropdown.jsx
```diff
@@ -0,0 +1,177 @@
...
+import { Bootstrap } from '../../components/main.js';
+import * as Styled from './styled.jsx';
+import { SimpleGroup } from './simple-group.jsx';
+import { Avatar } from './avatar.jsx';
+
+const { Button } = Bootstrap;
+
+export class GroupDropdown extends React.PureComponent {
+	static propTypes = {
+		// groups: PropTypes.arrayOf(
```
dustinsoftware: Commented out code

components/group-selector/dropdown.jsx
```diff
@@ -0,0 +1,177 @@
...
+	};
+
+	setHovered = group => {
+		if (this.state.hoveredGroupIndex !== -1)
+			if (group.groupId === this.props.groups[this.state.hoveredGroupIndex].groupId) return true;
+		return false;
+	};
+
+	handleClick = event => {
+		if (this.dropdownRef.current.contains(event.target)) return;
```
dustinsoftware: We should be using curly braces with `if` statements, I'll see if there's a lint rule we can enable for this

components/group-selector/dropdown.jsx
```diff
@@ -0,0 +1,177 @@
...
+		window.removeEventListener('mousedown', this.handleClick, false);
+	};
+
+	closeDropdown = () => {
+		this.setState({ isDropdownOpen: false });
+	};
+
+	setHovered = group => {
+		if (this.state.hoveredGroupIndex !== -1)
+			if (group.groupId === this.props.groups[this.state.hoveredGroupIndex].groupId) return true;
```
dustinsoftware: Curly braces will help here

components/group-selector/dropdown.jsx
```diff
@@ -0,0 +1,177 @@
...
+				name={group.name}
+				onClick={this.handleGroupSelection}
+				avatar={<Avatar group={group} />}
+			/>
+		));
+
+		const selectedGroupContents = (
+			<div>
+				<Styled.SelectedGroupAvatar>
+					{<Avatar group={this.props.selectedGroup} />}
```
dustinsoftware: Are the curlies needed here?

components/group-selector/dropdown.jsx
```diff
@@ -0,0 +1,177 @@
...
+		const groups = this.props.groups.map(group => (
+			<SimpleGroup
+				key={group.groupId}
+				isSelected={this.props.selectedGroup.groupId === group.groupId}
+				isHovered={this.setHovered(group)}
+				groupId={group.groupId}
+				kind={group.kind}
+				name={group.name}
+				onClick={this.handleGroupSelection}
+				avatar={<Avatar group={group} />}
```
dustinsoftware: Are the curlies needed here?

catalog/group-selector/variations.md
```diff
@@ -0,0 +1,113 @@
...
+		executeSearch={() => {alert("handled by application")}}
+		onCreateGroup={() => {alert("handled by application")}}
+		onSelectionChange={() => {alert("handled by application")}}
+		groups={state.groups}
+		selectedGroupId={state.selectedGroupId}
+		groupSearchResults={state.groups}
+		groupSelectorView={"no-groups"}
+		onGetStartedClick={() => {alert("handled by application")}}
+		onClaimGroupClick={() => {alert("handled by application")}}
+		isMobile={false}
```
dustinsoftware: Let's confirm that we have to have a distinction between mobile and desktop devices. If we have to have it, drawing a distinction between small and large viewport would be better, since you can have a desktop form-factor that is a touch device (iPad or Surface for example)

components/group-selector/modal/component.jsx
```diff
@@ -0,0 +1,337 @@
...
+		this.setState({ selectedGroupId: id });
+	};
+
+	resetModalState = () => {
+		this.setState({ modalContent: 'main' });
+	};
+
+	joinGroup = groupId => {
+		this.setModalState('main');
+		window.open(`https://www.faithlife.com/${groupId}`, 'noopener, noreferrer');
```
dustinsoftware: Faithlife does not have `www` in the prefix

dustinsoftware: We should link to the group token, not the group ID

components/group-selector/modal/component.jsx
```diff
@@ -0,0 +1,337 @@
...
+					{this.state.modalContent === 'main' && (
+						<Styled.MainModalContent>
+							<Styled.GroupSelectorModalTopGradient />
+							<Styled.ModalTitle>Find Your Church</Styled.ModalTitle>
+							<Styled.ModalSubtitle>in the Faithlife Church Directory</Styled.ModalSubtitle>
+							{this.props.showAlert && (
+								<Styled.ModalAlert color="primary">{this.props.alertText}</Styled.ModalAlert>
+							)}
+							<Styled.CreateGroupWrapper
+								style={{
```
dustinsoftware: Since we're using styled components already, it would better to create an `isPinnedToTop` prop, and then in the styled component definition, switch the style based on the prop!

[Example](https://github.com/Faithlife/styled-ui/blob/44ecd6f9f58eb3904ddd5d5ffc3cfa4e3159fb74/components/loading-spinner/styled.jsx#L17)

components/group-selector/modal/search-result.jsx
```diff
@@ -0,0 +1,210 @@
...
+		return string.charAt(0).toUpperCase() + string.slice(1);
+	}
+
+	componentDidMount() {
+		this.getSearchResultDetails();
+	}
+
+	getSearchResultDetails = () => {
+		if (this.props.claimable === 'true') {
+			// message = <p className={styles.searchResultMessage}>This is an empty group</p>;
```
dustinsoftware: More commented out code

components/group-selector/modal/search-result.jsx
dustinsoftware: Let's prefer to not store JSX in state. It looks like this state is derived from props and multiple back-to-back setState calls are made.. If we can do this all in a render without using state let's do that, otherwise store some boolean state values and switch what JSX gets rendered based off of that.

components/group-selector/modal/search-result.jsx
```diff
@@ -0,0 +1,210 @@
...
+					<Styled.SearchResultButton size="sm" color="primary" outline onClick={this.getStarted}>
+						Get Started
+					</Styled.SearchResultButton>
+				),
+			});
+		}
+	};
+
+	formatGroupKind = () => {
+		let string = this.props.kind.replace(/([A-Z])/g, ' $1');
```
dustinsoftware: This won't work for group kinds like `MinistryOrganization`, which need a space. We probably need to maintain a map of group kind to human-readable string.

components/group-selector/styled.jsx
```diff
@@ -0,0 +1,356 @@
...
+const { Button, Label } = Bootstrap;
+
+export const GroupSelector = styled.div`
+	${resetStyles};
+	max-width: 100%;
+	max-height: 800px;
+	text-align: center;
+	position: relative;
+	padding-bottom: 6px;
+	font-family: Source Sans Pro, Arial, Helvetica, sans-serif;
```
dustinsoftware: Fonts with spaces in the name should be surrounded by quotes per [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/font-family#Values)

components/group-selector/styled.jsx
```diff
@@ -0,0 +1,356 @@
...
+	background-color: white;
+	margin: 1px -10px 7px -10px;
+	box-shadow: 0 3px 2px 0 rgba(0, 0, 0, 0.12), 0 0 18px 0 rgba(0, 0, 0, 0.12);
+	border-radius: 3px;
+`;
+
+export const SearchResultAvatar = styled.div`
+	width: 40px;
+	height: 40px;
+	float: left;
```
dustinsoftware: It's a bit weird that float is getting used here, maybe flex would be better? You could run into cases where non-floated content wraps directly below the floated element.

components/modal/modal-header.jsx
```diff
@@ -3,11 +3,14 @@ import PropTypes from 'prop-types';
 import { Close } from '../icons';
 import * as Styled from './styled.jsx';
 
-export const ModalHeader = ({ title, subtitle, onClose }) => (
-	<Styled.ModalHeader>
+export const ModalHeader = ({ title, subtitle, onClose, style, moveCloseButton }) => (
+	<Styled.ModalHeader style={style}>
```
dustinsoftware: I wish I could enforce this in lint rules, but we shouldn't be mixing style props and styled components.

components/modal/modal.jsx
```diff
@@ -19,8 +19,16 @@ export class Modal extends React.Component {
 		subtitle: PropTypes.string,
 		/** Callback function for when the modal is close  */
 		onClose: PropTypes.func.isRequired,
+		/** Allow padding to be overridden */
+		modalPadding: PropTypes.string,
```
dustinsoftware: It's rare that the consumer actually needs to care about the internal padding. It seems like we should have a version of the Modal that has no inner content and no padding, since in this case the header and the footer aren't getting used at all. In this new version, the close icon would still be controlled by the Modal, but be absolutely positioned. It's up to the consumer to make sure the modal content does not overlap with the close icon.

We could call this new component `SimpleModal` or something. (Naming things is hard..)

JordanSjodinFaithlife:
FYI fo anyone reading I'm picking up this PR and will resolve these most recent comments.

components/group-selector/avatar.jsx
```diff
@@ -0,0 +1,35 @@
...
+	kind: PropTypes.string,
+	size: PropTypes.string,
+};
+
+const storedIcons = new Map();
+
+function getIconForGroupKind(kind) {
+	let icon = storedIcons.get(kind);
+	if (!icon) {
+		/* eslint import/namespace: ['error', { allowComputed: true }] */
```
dustinsoftware: Is this needed?

components/group-selector/avatar.jsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import * as icons from './icons.jsx';
+
+export function Avatar({ avatarUrl, name, kind, size = '32px' }) {
+	if (avatarUrl) {
+		return (
+			<img style={{ borderRadius: '3px', width: size, height: size }} src={avatarUrl} alt={name} />
```
dustinsoftware: Inline styles

components/group-selector/modal/search-result.jsx
```diff
@@ -0,0 +1,174 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Avatar } from '../avatar.jsx';
+import { Bootstrap } from '../../main.js';
+import * as Styled from '../styled.jsx';
+
+const { Button } = Bootstrap;
```
dustinsoftware: Let's use this button instead (here and below):
https://github.com/Faithlife/styled-ui/blob/master/components/main.js#L7

PR #19, opened 2018-08-10
Display children inside check-box component

awesomebob:
Would you consider merging this?
Here's the particular use case that I have in mind:

image

I'm passing in a <span> with a CSS class as a child component.

My alternative would be to include the text in the wrapper component, but then it wouldn't be as clickable as inside the button.


dustinsoftware:
Nice! Can you add an example? (see the catalog folder)

awesomebob:
@dustinsoftware added an example

PR #20, opened 2018-08-13
Add triangle icon to typeahead.

sportnak:
Add triangle icon for dropdown.

screen shot 2018-08-13 at 3 47 55 pm

### DIFF
--- a/components/text-input/inferred-typeahead.jsx
+++ b/components/text-input/inferred-typeahead.jsx
@@ -5,7 +5,7 @@ import { BootstrapContainer } from '../utils';
 import { Typeahead } from './typeahead.jsx';
 import { InferredBase } from './inferred-base.jsx';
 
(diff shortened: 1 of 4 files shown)


components/text-input/inferred-typeahead.jsx
```diff
@@ -5,7 +5,7 @@ import { BootstrapContainer } from '../utils';
 import { Typeahead } from './typeahead.jsx';
 import { InferredBase } from './inferred-base.jsx';
 
-const StyledTypeahead = styled(({ inferred, ...props }) => <Typeahead {...props} />)`
```
dustinsoftware: The inferred prop should not be passed to the underlying typeahead (that’s what this was for)

sportnak: Addressed: https://github.com/Faithlife/styled-ui/pull/20/files#diff-2e142b972dda92cca75cd76836f34707R63

PR #23, opened 2018-08-22
Match the line height to the radio button height

JeremyEinfeld:
The radio button input has a min-height of `1.5em`, but the line height is was still at a `normal` value resulting in an offset label:

alignment

This adjusts the line height to `1.5`, which should generally match the height of `1.5em`.


dustinsoftware:
Closing and re-opening to trigger a build

dustinsoftware:
Netlify does not like closed PR's re-opened. Opening a new one.

PR #25, opened 2018-08-27
Fix unclickable checkbox in Firefox

Imundy:
This fixes an issue where if children/title are not provided, the checkbox gets a width of 0 and becomes unclickable in firefox.


dustinsoftware:
Thanks. Would you mind adding a docs update to demo this behavior?

PR #29, opened 2018-09-06
Add files section component.

ianfisk:
- [x] Basic files section list with drop target.
- [ ] Different view styles (column option). Punting on this for now.

### DIFF
--- a/catalog/button/variations.md
+++ b/catalog/button/variations.md
@@ -37,7 +37,7 @@ showSource: true
 ```
 
 ### Supported style customizations
-Do not use the `style` prop to style this component (it will be ignored). Instead, if there is missing customization that you need for this component, ask to have it added to the theme prop.
+Do not use the `style` prop to style this component (it will be ignored). Instead, if there is missing customization that you need for this component, ask to have it added. The `theme` prop can be used to control color variations while the `styleOverrides` prop can be used for other styles.
 
 ```react
 showSource: true
@@ -46,10 +46,10 @@ showSource: true
 	<Button primary large theme={{ defaultColor: 'plum', hoverColor: 'darkslateblue' }}>
 		Plum
 	</Button>
-	<Button primary large theme={{ width: '200px' }}>
+	<Button primary large styleOverrides={{ width: '200px' }}>
 		200px Wide
 	</Button>
-	<Button primary theme={{ fontSize: '12px', padding: '8px' }}>
+	<Button primary styleOverrides={{ fontSize: '12px', padding: '8px' }}>
 		12px Font, 8px Padding
 	</Button>
 </ButtonDemo>
--- a/catalog/files-section/variations.md
+++ b/catalog/files-section/variations.md
@@ -97,9 +97,11 @@ state: {
 		renderFileActions={file => (
 			<div style={{ paddingTop: 4 }}>
 				{file.mediaType.startsWith('image') && (
-					<Button style={{ paddingRight: 16 }} onClick={() => console.log('Print file', file)} title="Print" link>
-						<svg stroke="currentColor" fill="none" strokeWidth="2" viewBox="0 0 24 24" strokeLinecap="round" strokeLinejoin="round" height="18" width="18"><polyline points="6 9 6 2 18 2 18 9" /><path d="M6 18H4a2 2 0 0 1-2-2v-5a2 2 0 0 1 2-2h16a2 2 0 0 1 2 2v5a2 2 0 0 1-2 2h-2" /><rect x="6" y="14" width="12" height="8" /></svg>
-					</Button>
+					<span style={{ padding: '0 16px 0 0' }}>
+						<Button onClick={() => console.log('Print file', file)} title="Print" link>
+							<svg stroke="currentColor" fill="none" strokeWidth="2" viewBox="0 0 24 24" strokeLinecap="round" strokeLinejoin="round" height="18" width="18"><polyline points="6 9 6 2 18 2 18 9" /><path d="M6 18H4a2 2 0 0 1-2-2v-5a2 2 0 0 1 2-2h16a2 2 0 0 1 2 2v5a2 2 0 0 1-2 2h-2" /><rect x="6" y="14" width="12" height="8" /></svg>
+						</Button>
+					</span>
 				)}
 				<Button onClick={() => console.log('Download file', file)} title="Download" link>
 					<svg width="16px" height="16px" viewBox="0 0 16 16"><g stroke="none" strokeWidth="1" fill="none" fillRule="evenodd"><g fill="currentColor"><path d="M9,7.79516537 L9,1 L7,1 L7,7.56670377 L4.71728604,5.35377997 L3.32519442,6.78977472 L7.48559482,10.8229778 L8.21010405,11.5253364 L8.90589288,10.7945156 L12.86268,6.63850972 L11.4141755,5.25943949 L9,7.79516537 Z M14.5,16 L16,16 L16,10 L14,10 L14,14 L2,14 L2,10 L0,10 L0,16 L1.5,16 L14.5,16 L14.5,16 Z" /></g></g></svg>
--- a/catalog/files-section/variations.md
+++ b/catalog/files-section/variations.md
@@ -1,0 +1,268 @@
+## Component Variations
+
+### With Drop Target
+
+```react
+showSource: true
+state: {
+	files: [
+		{
+			id: 'one',
+			mediaType: 'image/png',
+			name: 'File one',
+			byteCount: 1244124,
+			isProcessing: false,
+		},
+		{
+			id: 'two',
+			mediaType: 'video/mp4',
+			name: 'File two',
+			byteCount: 123,
+			isProcessing: false,
+		},
+		{
+			id: 'three',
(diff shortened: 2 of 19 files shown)


ianfisk:
CC @sportnak have a look and let me know if you want to make any API changes.

components/files-section/component.jsx
```diff
@@ -0,0 +1,141 @@
+/* eslint-disable react/no-unused-prop-types */
```
ianfisk: ESLint would not stop complaining about unused prop types (false positives).

components/files-section/component.jsx
```diff
@@ -0,0 +1,141 @@
...
+						<DropZone onDrop={onUploadFiles}>
+							<Styled.DropZoneText>{dropZoneText}</Styled.DropZoneText>
+
+							<Button link>
+								<Styled.FileInputLabel style={fileInputLabelStyle} htmlFor="fileSectionUploadFile">
+									{browseFilesButtonText}
+									<input
+										onChange={onUploadFiles}
+										type="file"
+										id="fileSectionUploadFile"
```
dustinsoftware: You shouldn't need `htmlFor` since the input is wrapped within the label.

components/utils/index.js
```diff
@@ -30,6 +30,29 @@ export const debouncedResize = callback => {
 	};
 };
 
+export const memoize = getValueForKey => {
+	const memoizedValues = new Map();
```
dustinsoftware: This map will expand indefinitely and cached values will never get evicted, `memoize-one` solves this problem elegantly by caching only one thing. It's fine to add that as a dependency.

components/files-section/component.jsx
```diff
@@ -0,0 +1,141 @@
...
+			browseFilesButtonText,
+			renderLoadingSpinner,
+			mediaTypeLabels,
+			theme,
+		} = this.props;
+		const files = this.props.files || [];
+		const lastFileIndex = files.length - 1;
+
+		return (
+			<ThemeProvider theme={theme || {}}>
```
dustinsoftware: `ThemeProvider` is best used for color variations.. I'd recommend specifying the `theme` prop directly to `Styled.Title` instead, and rename the prop to `titleFontSize`. Was this pattern copied from another component? If so, we should probably fix it there as well.

components/files-section/component.jsx
dustinsoftware: Does the problem go away if you avoid destructuring props? I'd prefer not to disable the linter.

```
// try this.props.foo.bar or props.foo.bar instead of:
const { foo } = this.props;
const { bar } = foo;
```

components/files-section/file-item.jsx
```diff
@@ -0,0 +1,106 @@
...
+				{onFileClicked ? (
+					<Button onClick={this.handleFileClick} link>
+						{fileIcon}
+					</Button>
+				) : (
+					fileIcon
+				)}
+				<Styled.FileInformation>
+					{onFileClicked ? (
+						<Button theme={fileLabelButtonTheme} onClick={this.handleFileClick} link>
```
dustinsoftware: This would be a good time to revisit the `theme` prop on button, and rename it to `styleOverrides` or something more descriptive, since adjusting the width has nothing to do with color

components/files-section/file-item.jsx
```diff
@@ -0,0 +1,106 @@
...
+			word: PropTypes.string,
+			powerpoint: PropTypes.string,
+			pdf: PropTypes.string,
+		}).isRequired,
+		onFileClicked: PropTypes.func,
+		renderFileActions: PropTypes.func,
+		renderLoadingSpinner: PropTypes.func,
+	};
+
+	convertBytesToFriendlyString = memoize(convertBytesToFriendlyString);
```
sportnak: Instead of memoizing, we could pass the file into a pure component to avoid calling the `convertBytesToFriendlyString` unless the file itself changes. We could do the same with the media label.

ianfisk: >we could pass the file into a pure component

I think you're right that it would be ok to remove memoization, and `FileItem` is already a pure component. I initially added it because Faithlife Music renders files in a component that renders _a lot_, causing a noticeable performance degradation, but I solved that issue in Faithlife Music a better way.

sportnak: Ah yeah, I hadn't event noticed that this was already a pure component.

components/utils/file-utils.js
```diff
@@ -0,0 +1,55 @@
+import { FileIcon, AudioIcon, VideoFileIcon } from '../icons';
+
+const mediaTypeLabelSelectors = {
+	video: mediaTypeLabels => mediaTypeLabels.video,
```
sportnak: Do we need to return a function instead of the value by itself?

ianfisk: Using selectors in this object allows it to live outside the scope of `mapMediaTypeToLabel` so we can create a regex from it to be used in `mapMediaTypeToIcon` as well.

components/files-section/component.jsx
dustinsoftware: Ok, so setting `skipShapeProps` fixes it? Would you mind pushing that up as a fix?

components/utils/file-utils.js
ianfisk: I could this object into `mapMediaTypeToLabel`, use the direct values, and then do something like:
```
let assetIcon = FileIcon;
if (recognizedMediaType.startsWith('audio')) {
	assetIcon = AudioIcon;
} else if (recognizedMediaType.startsWith('video')) {
	assetIcon = VideoFileIcon;
}
```
in `mapMediaTypeToIcon` instead.

ianfisk: But `startsWith` is not supported on IE.

components/files-section/component.jsx
ianfisk: Sounds good.

catalog/button/variations.md
```diff
@@ -37,7 +37,7 @@ showSource: true
 ```
 
 ### Supported style customizations
-Do not use the `style` prop to style this component (it will be ignored). Instead, if there is missing customization that you need for this component, ask to have it added to the theme prop.
+Do not use the `style` prop to style this component (it will be ignored). Instead, if there is missing customization that you need for this component, ask to have it added. The `theme` prop can be used to control color variations while the `styleOverrides` prop can be used for other styles.
```
ianfisk: >Do not use the `style` prop to style this component (it will be ignored)

@dustinsoftware currently `style` isn't actually ignored. Should we enforce this?

ianfisk: Also, I'm not in love with `styleOverrides` but can't decide if there is anything better. Is there a reason we are ignoring the `style` prop and only allowing certain overrides?

dustinsoftware: The docs are a bit stale here, I changed my mind later since there were too many complaints :) I added some reasoning [here](https://faithlife.github.io/styled-ui/#/?a=guidelines-for-use) about when to adjust styles directly

catalog/files-section/variations.md
```diff
@@ -0,0 +1,270 @@
...
+---
+<FilesSectionDemo>
+	<FilesSection
+		files={state.files}
+		onFileClicked={file => console.log(file)}
+		renderFileActions={file => (
+			<div style={{ paddingTop: 4 }}>
+				{file.mediaType.startsWith('image') && (
+					<span style={{ padding: '0 16px 0 0' }}>
+						<Button onClick={() => console.log('Print file', file)} title="Print" link>
```
dustinsoftware: Just FYI we are planning on changing the `link` prop to `primaryTransparent`, since `link` doesn't accurately describe the styling and is going to be more confusing when we add an upcoming `ButtonAnchor` component to wrap anchor tags

components/drop-zone/component.jsx
```diff
@@ -0,0 +1,59 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import * as Styled from './styled.jsx';
+
+export class DropZone extends PureComponent {
```
dustinsoftware: Missing documentation for this component and proptypes

components/files-section/component.jsx
```diff
@@ -0,0 +1,140 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import { Button } from '../demo-button/component.jsx';
+import { DropZone } from '../drop-zone/component.jsx';
+import { FileItem } from './file-item.jsx';
+import * as Styled from './styled.jsx';
+
+// margin is being overwritten when setting in the styled component and rendering this component in a bootstrap container
+const fileInputLabelStyle = { marginBottom: 0 };
```
dustinsoftware: You can use a more specific style to resolve this instead.
https://github.com/styled-components/styled-components-website/issues/187

components/drop-zone/component.jsx
ianfisk: 👍 I didn't export this component to consumers of the library so I didn't add documentation, but I was thinking it is a useful for component. I'll add the export and docs.

catalog/files-section/variations.md
ianfisk: Ok. If the change happens before this is merged I can fix this PR.

PR #30, opened 2018-09-11
Add transparent border to `link` button variation

awesomebob:
This allows a `small square link` to be 32x32 instead of 30x32.


awesomebob:
@dustinsoftware @etihwddot

etihwddot:
Could we get a sample added to the Catalog page so I can confirm that it works?

awesomebob:
Sample added

PR #31, opened 2018-09-11
Add close functionality to HelpBox

ELanning:
This adds a close icon, and a function to call when that icon is clicked to the `HelpBox` component.

- [ ] Possibly use `onChange` instead of `onClick`?


components/help-box/styled.jsx
```diff
@@ -20,6 +20,30 @@ export const BulbIcon = styled(LightBulbH)`
 	flex: none;
 `;
 
+export const CloseButton = styled.button`
+	margin: auto;
+	margin-right: 16px;
+	height: 18px;
+
+	/* Removes all the browser default button styling */
```
dustinsoftware: Are all of these resets actually needed? If not, can we just stick with removing border and padding?

ELanning: Some of the obvious ones _probably_ can be removed but it's very hard to make any assertions when trying to support 4 or so browsers and different versions of those browsers. I imagine the text ones can be safely removed, but browsers are weird. I'll look into trimming some of the obvious ones, but it's a bit scary on windows where I can't test it on Safari.

Note we probably should not do this `outline: none;` for accessibility reasons-
source for the styles: https://gist.github.com/MoOx/9137295

ELanning: It looked like a lot of them could be safely removed (provided Safari doesn't throw me any curve balls). Good call 👍

dustinsoftware:
Missing cursor: pointer and a hover style. It also looks a bit weird that the X is center aligned, was that called for in a spec?

ELanning:
>It also looks a bit weird that the X is center aligned, was that called for in a spec?

That was a mistake on my part, I didn't see the larger boxes in the spec had the "x" in the top right corner. Should be an easy change, fortunately.

>Missing cursor: pointer and a hover style.

Working on this now, thanks for the reminder 👍

PR #36, opened 2018-09-13
Fix button prop types.

ianfisk:
Rendering the `Button` or `AnchorButton` component yields the following warnings:
```
Warning: Failed prop type: The prop `baseComponent` is marked as required in `Button`, but its value is `undefined`.
```
and
```
Warning: Failed prop type: Invalid prop `baseComponent` of type `function` supplied to `BaseButton`, expected a single ReactElement.
```

### DIFF
--- a/components/button/base-button.jsx
+++ b/components/button/base-button.jsx
@@ -3,47 +3,51 @@ import PropTypes from 'prop-types';
 import { applyVariations } from '../utils';
 import * as Styled from './styled.jsx';
 
+export const SharedButtonPropTypes = {
+	/** The contents of the button (can be text, svg, or other element) */
+	children: PropTypes.node,
+	/** See the docs for how to override styles properly  */
+	className: PropTypes.string,
+	/** Condensed button padding. Uses same padding for horizontal and vertical. */
+	condensed: PropTypes.bool,
+	/** An optional theme */
+	theme: PropTypes.shape({
+		defaultColor: PropTypes.string,
+		hoverColor: PropTypes.string,
+		activeColor: PropTypes.string,
+		disabledColor: PropTypes.string,
+	}),
+	/** Style overrides */
+	styleOverrides: PropTypes.shape({
+		width: PropTypes.string,
+		fontSize: PropTypes.string,
+		padding: PropTypes.string,
+	}),
+	/** Primary button variation */
+	primary: PropTypes.bool,
+	/** Primary outline variation */
+	primaryOutline: PropTypes.bool,
+	/** Medium variation */
+	medium: PropTypes.bool,
+	/** Small variation */
+	small: PropTypes.bool,
+	/** Large variation */
+	large: PropTypes.bool,
+	/** Extra large variation */
+	extraLarge: PropTypes.bool,
+	/** Transparent with primary text variation */
+	primaryTransparent: PropTypes.bool,
+	/** Transparent with minor text variation */
+	minorTransparent: PropTypes.bool,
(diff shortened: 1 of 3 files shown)


components/button/base-button.jsx
```diff
@@ -3,47 +3,51 @@ import PropTypes from 'prop-types';
...
-		large: PropTypes.bool,
-		/** Extra large variation */
-		extraLarge: PropTypes.bool,
-		/** Transparent with primary text variation */
-		primaryTransparent: PropTypes.bool,
-		/** Transparent with minor text variation */
-		minorTransparent: PropTypes.bool,
-		/** Enables rendering a display: flex span, needed for rendering SVG icons */
-		renderIcon: PropTypes.node,
+		baseComponent: PropTypes.func.isRequired,
```
ianfisk: It may be technically more correct to use `PropTypes.oneOfType([PropTypes.func, PropTypes.element]).isRequired`, but our base components are currently styled components which are functions.

ianfisk:
That is a major downside. We could just make `baseComponent` not required in `BaseButton` and know it will always be provided.

awesomebob:
Let's go with that. If we wanted, we could throw an error in BaseButton if that prop is not supplied, along with a comment explaining why we went that route instead of adding `isRequired` to the prop.

ianfisk:
On second thought, that means the `baseComponent` prop type now shows in the button documentation table, and `type` (from `Button`) is missing.

ianfisk:
We might want to exclude `baseComponent` in the defined prop types and move
```
/** The type of button (for instance, submit). Note this is used by Button only. */
type: PropTypes.string,
```
to `ButtonBase.propTypes`.

awesomebob:
I suppose we could leave `baseComponent` out of the propTypes. We looked around for a `hidden` or `private` attribute to hide it from the docs, but were unsuccessful.

One note about the `type` attribute is that it means *different* things on a `<a>` tag verses a `<button>` tag. We want a default `type` prop for the Button, but not for the AnchorButton component.

ianfisk:
>We want a default type prop for the Button, but not for the AnchorButton component.

Right. Moving the prop type definition would only add it to the docgen table, but `defaultProps` would stay the same on `Button`.

ianfisk:
Oh I just realized we could do a total hack like this to hide it from the docgen table:
```
const hiddenPropTypes = {
	baseComponent: PropTypes.func.isRequired,
};

export class BaseButton extends PureComponent {
	static propTypes = {
		...hiddenPropTypes,
		// other prop types
	};
```

awesomebob:
You can try it, but I don't think that works. I think it would still be imported by the child components (and therefore wrongly required), and would still appear in the docs, just without a comment.

ianfisk:
Oh you're right. My tab refreshed but I still had a cached version of the page.

ianfisk:
I've excluded `baseComponent` from the defined prop types here https://github.com/ianfisk/styled-ui/commit/c52a6e52eeea65cbca4974de4ea0ce6ce18cec94.

I also think it would be good to document both the shared prop types (like we are doing currently) as well as differences between `Button` and `AnchorButton` like this https://github.com/ianfisk/styled-ui/commit/1bda19905cbec7e7a89801db1a8ea61e8aaaade8.

awesomebob:
I like those changes. Could you open a PR?

ianfisk:
Opened https://github.com/Faithlife/styled-ui/pull/37. Closing this.

PR #39, opened 2018-09-13
Increase label font size to 16px.

ianfisk:
CC @awesomebob


ianfisk:
screen shot 2018-09-13 at 2 42 04 pm
screen shot 2018-09-13 at 2 42 19 pm

PR #43, opened 2018-09-14
Add support for justify content to stretched buttons

etihwddot:


etihwddot:
@dustinsoftware this implementation uses css grid instead of flexbox in the content of the button because I liked the `justifyContent: '[left, center, right]'` better than `justifyContent: '[flex-start, center, flex-end]`. Though that means this will be break `styled-ui` for IE11 users. Is that something we are ok with?

If not we could move it back to `flex` and map `left` and `right` to `flex-start` and `flex-end`.

PR #44, opened 2018-09-17
Add tooltip component

korbinancell:
Add a not-bootstrap tooltip

### DIFF
--- a/catalog/index.js
+++ b/catalog/index.js
@@ -17,6 +17,7 @@ import {
 	DropZone,
 	GroupSelector,
 	GroupSelectorModal,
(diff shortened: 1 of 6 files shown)


korbinancell:
I'm not sure why it is erroring. It seems to not like the template strings that I use.

dustinsoftware:
Hey @korbinancell thanks for the PR. The component in its current form is missing a lot of the functionality provided by the reactstrap tooltip, we'll need to make sure that whatever tooltip we add here follows a similar API before we add one here. Let me know if you'd like to meet to work through how to do this.

PR #49, opened 2018-09-20
Without footer

awesomebob:
I added a description to the `renderFooter` prop as well. I'm open to suggestions about how to word this better.

### DIFF
--- a/components/modal/modal.jsx
+++ b/components/modal/modal.jsx
@@ -106,7 +106,7 @@ export class Modal extends React.Component {
 					>
 						<ModalHeader title={title} subtitle={subtitle} onClose={onClose} />
 						<Styled.ModalContent> {children} </Styled.ModalContent>
-						{withoutFooter ||
+						{!withoutFooter &&
 							(renderFooter ? (
 								renderFooter()
 							) : (
--- a/components/modal/modal.jsx
+++ b/components/modal/modal.jsx
@@ -39,7 +39,10 @@ export class Modal extends React.Component {
 				text: PropTypes.string.isRequired,
 			}),
 		}),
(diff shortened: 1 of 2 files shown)


components/modal/modal.jsx
```diff
@@ -102,7 +106,7 @@ export class Modal extends React.Component {
 					>
 						<ModalHeader title={title} subtitle={subtitle} onClose={onClose} />
 						<Styled.ModalContent> {children} </Styled.ModalContent>
-						{renderFooter ? (
+						{withoutFooter ? null : renderFooter ? (
```
JordanSjodinFaithlife: Nested ternary can be a little hard to read. Would something like this be more clear?

```
{withoutFooter ||
    (renderFooter ? (
        renderFooter()
    ) : (
        <DefaultModalFooter useFullWidthButtons={verticalButtons} {...footerProps} />
))}
```

components/modal/modal.jsx
```diff
@@ -101,11 +106,12 @@ export class Modal extends React.Component {
 					>
 						<ModalHeader title={title} subtitle={subtitle} onClose={onClose} />
 						<Styled.ModalContent> {children} </Styled.ModalContent>
-						{renderFooter ? (
-							renderFooter()
-						) : (
-							<DefaultModalFooter useFullWidthButtons={verticalButtons} {...footerProps} />
-						)}
+						{withoutFooter ||
```
dustinsoftware: I'm not sure that this is clearer. `true` will be returned from this expression now which seems incorrect.

dustinsoftware: && is fine since it will return the right hand side of the expression, || will return the left-hand side if it evaluates to truthy. Can you update this to use `&&` or go back to the nested ternary?

PR #51, opened 2018-09-24
Change how we are styling SimpleModal to fix issue with reset styles

JordanSjodinFaithlife:
Due to how I was styling SimpleModal, I was actually not getting the styling I thought I was. Thus, the reset styles was not working and the `box-sizing` was set incorrectly. This also means that during development, some sizes were done with `content-box` instead of `border-box`.

This fixes the reset styles, and cleans up the css for the affected components.


JordanSjodinFaithlife:
@dustinsoftware could you take a look?

dustinsoftware:
Whoops, I missed that when I merged the original PR. Thanks.

PR #52, opened 2018-09-25
Cleanup group Selector

JordanSjodinFaithlife:
This PR contains a number of fixes:

One is it removed some duplicated code and switched from using float and hardcoded padding to center things to using flex box.

Also, if removed some styles that were being set directly on components instead of using styled components.

Lastly, making a complicated component that works on everyone's web page is hard. 'p' elements were used in this component and were getting useragent styles with margins such that it looked nice in the catalog, but could look bad elsewhere. Switching to divs works around this and seems better since the 'p' elements weren't really being used for paragraphs of text.

@dustinsoftware

### DIFF
--- a/components/group-selector/styled.jsx
+++ b/components/group-selector/styled.jsx
@@ -48,6 +48,19 @@ export const SelectedGroup = styled.button`
 	align-items: center;
 `;
 
+export const AvatarWrapper = styled.div`
+	border-radius: 3px;
+	width: ${props => props.size}px;
+	height: ${props => props.size}px;
+	overflow: hidden;
+`;
+
+export const AvatarImage = styled.img`
+	display: block;
+	width: ${props => props.size}px;
+	height: ${props => props.size}px;
+`;
+
 export const SimpleGroup = styled.div`
 	text-align: left;
 	padding: 6px;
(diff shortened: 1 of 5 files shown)


components/group-selector/styled.jsx
```diff
@@ -27,91 +27,67 @@ export const GroupInputButton = styled.div`
...
-	vertical-align: middle;
-	max-width: 200px;
-	width: 76%;
-`;
-
-export const SimpleGroupName = styled.div`
-	font-size: 13px;
-	color: #585250;
-	white-space: nowrap;
+	width: ${props => props.size}px;
```
dustinsoftware: Is the width and height being overridden anywhere? Or is this old code?

dustinsoftware:
There are some issues with scrollbar fading in Firefox:
image

dustinsoftware:
When mousing over Create, the button does not change state and appears to be unclickable:
image

components/group-selector/styled.jsx
JordanSjodinFaithlife: should be set here 
https://github.com/Faithlife/styled-ui/pull/52/files/926526371bfd2bbbdc9c57ad2d9c42d620306cdb#diff-9739e43a5a0288393f4da68ba14cfefcR16

JordanSjodinFaithlife:
The two bugs you mentioned are present in the current version. Could I address those in a future PR? (i.e. not have them block this one)

PR #55, opened 2018-10-02
Reset the file section's input so uploading the same file twice works.

ianfisk:
Currently trying to upload the same file twice in a row results in a no-op.

### DIFF
--- a/components/files-section/component.jsx
+++ b/components/files-section/component.jsx
@@ -80,6 +80,13 @@ export class FilesSection extends PureComponent {
 		this.fileInputRef.current.click();
 	};
 
+	handleUploadWithFileInput = event => {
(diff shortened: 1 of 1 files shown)


dustinsoftware:
What if onUploadFiles is undefined?

ianfisk:
We conditionally render the upload button only if `onUploadFiles` is defined, so I didn't add it to the event handler. That being said, I did initially write
```javascript
const { onUploadFiles } = this.props;
if (onUploadFiles) {
	onUploadFiles(event);
}

this.fileInputRef.current.value = '';
```
so I could add that back if you'd like.

dustinsoftware:
Yeah let’s add the condition back

PR #61, opened 2018-10-08
Fix group name truncation when name is larger than selector

JordanSjodinFaithlife:
This PR fixes truncation of the group name. It was going beyond the bounds of the selector and overlapping the triangle icon.

Plus:
Fixes an issue where when the selector was really constrained, the avatar would get cut off, which we never want.

Adds a sample avatarUrl to the sample data in the catalog to better show off the selector.

### DIFF
--- a/catalog/group-selector/variations.md
+++ b/catalog/group-selector/variations.md
@@ -7,7 +7,7 @@ state: {
 		name: 'Michaels Test Church',
 		groupId: 1,
 		kind: 'church',
-		avatarUrl: '',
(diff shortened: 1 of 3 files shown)


catalog/group-selector/variations.md
```diff
@@ -7,7 +7,7 @@ state: {
 		name: 'Michaels Test Church',
 		groupId: 1,
 		kind: 'church',
-		avatarUrl: '',
+		avatarUrl: 'https://avatars.logoscdn.com/451/6830451_f896e509e7504b33812f0f6bf086ffa8.jpg',
```
dustinsoftware: Can't merge this change in 😆

PR #62, opened 2018-10-09
Use custom scrollbar for GroupSelector

JordanSjodinFaithlife:
@dustinsoftware 

The group selector has some complicated scrolling behavior, and not all browses displayed the scrollbar above the content.

Using css to attempt to alter the scrollbar was not a viable solution due to Firefox not supporting the css required.

Using a custom scrollbar allows us to have control over the location of the scrollbar and get a consistent look accross browsers.

Plus, fixed some small buttons on secondary modal content and fixed a gap in the header.


components/group-selector/styled.jsx
```diff
@@ -122,7 +123,20 @@ export const DropdownButtonContainer = styled.div`
...
+
+export const ModalScrollViewContentClass = 'modal-scroll-view-content';
+export const ModalScrollView = styled(ScrollArea)`
 	position: relative;
+	height: 60vh;
+	width: 375px;
+	background: white;
+	border-radius: 4px;
+
+	.${ModalScrollViewContentClass} {
```
dustinsoftware: It would be better to pass in a reference to another component than to create a class name for it, can we do that here? [Docs](https://www.styled-components.com/docs/advanced#referring-to-other-components)

yarn.lock
```diff
@@ -8167,6 +8197,14 @@ react-router@^3.2.0:
     prop-types "^15.5.6"
     warning "^3.0.0"
 
+react-scrollbar@^0.5.4:
+  version "0.5.4"
+  resolved "https://registry.yarnpkg.com/react-scrollbar/-/react-scrollbar-0.5.4.tgz#270576a28d0f7783b2304b284dbd38d82b36b0e0"
```
dustinsoftware: I'm a bit worried that this dependency (and react-motion) are going to bloat the main bundle. If we need this dependency, let's move anything that uses it to a separate bundle. [This is how we did it for text-input](https://github.com/Faithlife/styled-ui/blob/master/webpack.config.js#L11)

components/group-selector/styled.jsx
JordanSjodinFaithlife: I can get around using this class name

yarn.lock
JordanSjodinFaithlife: I'll try this

PR #63, opened 2018-10-09
Improve close logic for backdrop

JordanSjodinFaithlife:
Using `onClick`, a `mouseDown` in a child , then dragging your mouse, and a `mouseUp` in the background causes `onClick` to be called, and the `target` is the backdrop, thus it closes. This is no good when dragging a scrollbar as it is very easy to be dragging with the mouse down and then release while over the backdrop.


dustinsoftware:
This change breaks closing the modal on iOS..

JordanSjodinFaithlife:
@dustinsoftware thats unfortunate. So iOS gets the onClick but not the mouseDown/ups? Think there are touchDown🏈/Ups that I could use?

Currently it is pretty easy to close the modals on accident..

JordanSjodinFaithlife:
@dustinsoftware updated, and tested on a real iPhone.

PR #64, opened 2018-10-12
Create styled radio button

kbsletten:


dustinsoftware:
Version updates need to be done outside the PR

Also, alignment seems off:
image

JordanSjodinFaithlife:
ezgif-5-6ab104902028

Should this be flashing white like this? Safari on Mac 10.13

kbsletten:
@JordanSjodinFaithlife That does seem to also happen on Checkbox (which I copied), I'll see if I can't figure out what's causing it.

components/check-box/styled.jsx
```diff
@@ -22,6 +22,10 @@ export const CheckboxContainer = styled.button`
 	min-height: 16px;
 	background: transparent;
 
+	&:active {
+		color: buttontext;
```
kbsletten: This is my solution to the "white flash" that Safari has for these. By default Safari applies "activebuttontext" color to buttons with the ":active" pseudo-class. I'm not convinced this is the best idea, however because we're now enforcing a "buttontext" color on these controls. This makes one additional step to coloring the text of these elements, though it's arguably more discoverable than the Safari thing because it will act the same in all browsers.

kbsletten: @dustinsoftware is there a more "styled-ui" way of doing this, in your opinion?

PR #66, opened 2018-10-18
Split input onChange from typeahead onchange in inferred-typeahead

Imundy:
The typeahead onChange returns and array while the input onChange returns a string. It's also worth noting the inferred typeahead doesn't really work with `multiple` set to true but I have not fixed that here.


dustinsoftware:
This control is being deprecated in favor of `text-input-v2`.

PR #68, opened 2018-10-26
Add Popover component

korbinancell:
Create a non bootstrap popover that conforms to FL style guides using the newest version of react-popper.

### DIFF
--- a/catalog/popover/variations.md
+++ b/catalog/popover/variations.md
@@ -1,0 +1,106 @@
+## Popover
+
+```react
+showSource: true
+state: { isOpen: false }
+---
+<PopoverDemo>
+	<PopoverManager>
+		<PopoverReference>
+			<Button primary medium onClick={() => setState({ isOpen: !state.isOpen })}>Show a Popover!</Button>
+		</PopoverReference>
+		<Popover isOpen={state.isOpen} placement="top">Hello!</Popover>
+	</PopoverManager>
+</PopoverDemo>
+```
+
+## Placement
+
+```react
+showSource: true
+state: { isOpen: false }
+---
+<PopoverDemo>
+	<PopoverManager>
+		<PopoverReference>
+			<Button primary medium onClick={() => setState({ isOpen: !state.isOpen })}>Show a Popover!</Button>
+		</PopoverReference>
+		<Popover isOpen={state.isOpen} placement="top">Hello!</Popover>
+	</PopoverManager>
+	<PopoverManager>
+		<PopoverReference>
+			<Button primary medium onClick={() => setState({ isOpen: !state.isOpen })}>Show a Popover!</Button>
+		</PopoverReference>
+		<Popover isOpen={state.isOpen} placement="bottom">Hello!</Popover>
+	</PopoverManager>
+	<PopoverManager>
+		<PopoverReference>
+			<Button primary medium onClick={() => setState({ isOpen: !state.isOpen })}>Show a Popover!</Button>
+		</PopoverReference>
+		<Popover isOpen={state.isOpen} placement="left">Hello!</Popover>
+	</PopoverManager>
+	<PopoverManager>
+		<PopoverReference>
+			<Button primary medium onClick={() => setState({ isOpen: !state.isOpen })}>Show a Popover!</Button>
+		</PopoverReference>
+		<Popover isOpen={state.isOpen} placement="right">Hello!</Popover>
+	</PopoverManager>
+</PopoverDemo>
+```
+
+## Options
+
+```react
+showSource: true
+state: { isOpen: false }
+---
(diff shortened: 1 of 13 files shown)


dustinsoftware:
Thanks!! I'll get to this soon.

@jake-robertson please review: https://deploy-preview-68--faithlife-styled-ui.netlify.com/#/popover/variations

jake-robertson:
@korbinancell @dustinsoftware 

This looks pretty good, but what is it exactly? Is this the "tooltip" we'd use for on hover/tap of iconography? Is this the "Menu" we'd use on "dropdown buttons"? Is this merely the little popover that appears next to buttons occasionally to confirm an action happened (e.g., copied to clipboard!)?

korbinancell:
@jake-robertson 
I just added a new section about tooltips. This component would be the tooltip or the little popover. It is set up to be used both ways.

components/popover/component.jsx
```diff
@@ -0,0 +1,166 @@
...
+		]),
+		styleOverrides: PropTypes.shape({
+			hideShadow: PropTypes.bool,
+			width: PropTypes.string,
+			padding: PropTypes.string,
+		}),
+		theme: PropTypes.shape({
+			backgroundColor: PropTypes.string,
+			textColor: PropTypes.string,
+			boder: PropTypes.string,
```
Imundy: typo: `boder`

components/popover/component.jsx
```diff
@@ -0,0 +1,166 @@
...
+		}),
+	};
+
+	static defaultProps = {
+		placement: 'top',
+		theme: {
+			backgroundColor: '#ffffff',
```
Imundy: It's probably slightly safer to use white from the shared styles:
https://github.com/Faithlife/styled-ui/blob/master/components/shared-styles/index.js#L38

components/popover/component.jsx
```diff
@@ -0,0 +1,166 @@
...
+		this.handleIsPopoverShowing(prevProps.isOpen, this.props.isOpen, this.props.delay);
+	}
+
+	handleIsPopoverShowing = (prevIsOpen, isOpen, delay) => {
+		if (prevIsOpen !== isOpen && !delay) {
+			this.setState({ showPopper: isOpen });
+		} else if (prevIsOpen && !isOpen && !!delay) {
+			setTimeout(
+				() => this.setState({ showPopper: false }),
+				typeof delay === 'object' ? delay.hide : delay,
```
Imundy: A diverging change, but I think it would make the API slightly simpler if we just always require delay to be an object containing `show` and `hide` instead of also allowing it to be a string.

components/popover/tooltip.jsx
```diff
@@ -0,0 +1,38 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import { Popover } from './component.jsx';
+import { PopoverManager, PopoverReference } from './popper-helpers.jsx';
+
+export class Tooltip extends PureComponent {
```
Imundy: I think the `PureComponent` rendering *may* break if `ref.current` changes. In that case we'll still have the same `ref` so `oldProps.container === newProps.container`, but the `current` property would change.

components/popover/component.jsx
```diff
@@ -0,0 +1,161 @@
...
+
+	componentDidUpdate(prevProps) {
+		this.handleIsPopoverShowing(prevProps.isOpen, this.props.isOpen, this.props.delay);
+	}
+
+	handleIsPopoverShowing = (prevIsOpen, isOpen, delay) => {
+		if (prevIsOpen !== isOpen && !delay) {
+			this.setState({ showPopper: isOpen });
+		} else if (prevIsOpen && !isOpen && !!delay) {
+			setTimeout(() => this.setState({ showPopper: false }), delay.hide || 0);
```
dustinsoftware: Possible crash if the component unmounts before the timer fires. Let's make sure to cancel this work.

catalog/popover/variations.md
```diff
@@ -0,0 +1,129 @@
...
+```react
+showSource: true
+state: { isOpen: false }
+---
+<PopoverDemo>
+	<PopoverManager>
+		<PopoverReference>
+			<Button primary medium onClick={() => setState({ isOpen: !state.isOpen })}>Show a Popover!</Button>
+		</PopoverReference>
+		<Popover isOpen={state.isOpen} placement="top" theme={{ backgroundColor: '#ebf7ff', border: 'black solid 1px' }}>Hello!</Popover>
```
dustinsoftware: `theme` should be used for just colors, let's move `border` into `styleOverrides`

dustinsoftware:
Merging as-is, please send a follow up PR with the fixes proposed above and we'll get this shipped.

PR #71, opened 2018-10-31
Date picker

Imundy:
Mostly opening this to get feedback on the approach. This is a direct port of the date picker currently in Giving. I tried to abstract away dependence on a specific date library a bit by passing in the date functions it's going to use as a props object. That *should* let consumers choose but it might need some tuning to be a better API (right now it's pretty one to one with date-fns).


components/date-picker/calendar-input/calendar-date/component.jsx
```diff
@@ -0,0 +1,61 @@
...
+		this.props.dateFunctions.getDate(date1) === this.props.dateFunctions.getDate(date2) &&
+		this.props.dateFunctions.getMonth(date1) === this.props.dateFunctions.getMonth(date2) &&
+		this.props.dateFunctions.getYear(date1) === this.props.dateFunctions.getYear(date2);
+
+	render() {
+		const currentDate = new Date();
+		const selectedDate = this.props.selectedDate;
+		const date = this.props.date;
+		let currentDayDot = null;
+		let CalendarWeekday = Styled.calendarWeekDay;
```
dustinsoftware: The exports in `Styled` should be PascalCase so that this variable is not necessary.

dustinsoftware:
Thanks! I like the initial approach and will be able to look at this in depth soon.

package.json
```diff
@@ -59,8 +59,10 @@
 		"babel-runtime": "^6.26.0",
 		"bootstrap": "^4.1.1",
 		"catalog": "^3.5.4",
+		"chrono-node": "^1.3.5",
 		"cross-env": "^5.2.0",
 		"css-loader": "^1.0.0",
+		"date-fns": "^2.0.0-alpha.25",
```
Imundy: Just as a note - these are both just dev dependencies for using it in catalog.

dustinsoftware:
Looks like the styles on Faithlife.com have changed a bit: 
image

SpencerAlexander:
Oo this looks interesting. I think I could use this component in the Skeletor payment plans feature I'm working on. 
screen shot 2018-11-14 at 8 55 56 am

https://app.zeplin.io/project/5b511c28eb657422b8b29957/screen/5b6b2fc2a9b4bd1bc73a085d
@Imundy, @dustinsoftware is this a learning exercise or could this get in to styled-ui in the near future?

Imundy:
I think this is mostly good to go after updating the styles and adding some documentation how to use it. I'm not sure what the timeline would be for it to be in a published package, but I could probably wrap up those 2 things later today.

Imundy:
This updated to the [official (?) date picker UI](https://app.zeplin.io/project/58a23c2d4597afb248c13a43/screen/5b97faeaaef2c264ed60c775).
It still only implements the most basic version (input with calendar icon). In order to support those other date picker variations, I've split it into two components - the actual date picker and the date picker input. I'm currently exporting the date-picker, but with no documentation. I imagine we'll use it as a building block in other controls and we might not want to explicitly export it in the future once we have all of then/

components/date-picker-input/component.jsx
```diff
@@ -0,0 +1,153 @@
...
+import { DatePicker } from '../date-picker/component.jsx';
+import * as Styled from './styled.jsx';
+
+export class DatePickerInput extends PureComponent {
+	static propTypes = {
+		defaultSelectedDate: PropTypes.instanceOf(Date),
+		/** Functions that operate on a JS Date obejct.
+		 * The following functions must be provided:
+		 *
+		 * startOfWeek, startOfMonth, endOfWeek, endOfMonth, getYear, getMonth,getDate, addWeeks, addMonths,s ubMonths, isBefore, format,isValid,parseUserDateString
```
dustinsoftware: Typo: `s ubMonths`. There are also some spacing errors.

components/date-picker-input/component.jsx
```diff
@@ -0,0 +1,153 @@
...
+					<Styled.Input
+						type="text"
+						onBlur={this.handleBlur}
+						onChange={this.handleChange}
+						onFocus={this.handleFocus}
+						value={value}
+						disabled={disabled}
+					/>
+					<Styled.CalendarButton
+						innerRef={this.icon}
```
dustinsoftware: What is this ref used for?

components/date-picker-input/styled.jsx
```diff
@@ -0,0 +1,43 @@
...
+	font-family: inherit;
+	color: @font-color;
+	border: 1px solid ${colors.borderColor};
+	border-radius: 3px;
+	height: 32px;
+	width: 100%;
+	margin: 0;
+	padding: 0.25rem 0.5rem;
+	transition: color 0.2s linear;
+	box-sizing: border-box;
```
dustinsoftware: Let's import the global reset instead of re-applying border-box.
https://github.com/Faithlife/styled-ui/blob/d1c29ee8e805d98c50eb8f9e72c74dbf258e3c01/components/button/styled.jsx#L3

components/date-picker-input/styled.jsx
```diff
@@ -0,0 +1,43 @@
...
+`;
+
+export const CalendarButton = styled.div`
```
dustinsoftware: Can we instead `onClick` handlers to `button` elements instead of `div`?

components/date-picker/calendar-date/component.jsx
```diff
@@ -0,0 +1,57 @@
...
+		}
+
+		if (this.areDatesEqual(currentDate, date)) {
+			currentDayDot = <Styled.CalendarWeekDayCurrentDay />;
+		}
+
+		return (
+			<CalendarWeekday
+				onClick={this.setSelectedDate}
+				tabIndex="-1"
```
dustinsoftware: Why is this tabIndex necessary? Can we use a button instead?

dustinsoftware:
Merging as-is, please send a follow up PR with improvements

SpencerAlexander:
@Imundy, fwiw I think Skeletor would be interested in consuming just the `date-picker` without the input wrapper that you mentioned splitting out. Essentially just the calendar icon that pops up the picker and sends back the selected date. Bonus points if we could pass in an optional override icon, but the calendar would be fine as well. 

Thanks for your work on this! 🎉

Imundy:
Oh I've already split it in the PR @SpencerAlexander, so this now exports `DatePicker` and `DatePickerInput`  That being said, if you have another use case, it might be worthwhile to build an additional input control on top of the DatePicker calendar view in styled UI, rather than build that control in Skeletor (if it seems like it's going to be something we want to reuse at least). There are at least a few other types of controls in that zeplin spec already.

SpencerAlexander:
> Oh I've already split it in the PR

🎊nice, excited to check it out! If you don't mind I'll probably ping you on the future Skeletor PR with it's usage so you can tell me if I'm using it right =)

SpencerAlexander:
screen shot 2018-11-15 at 3 06 21 pm

A working date-picker from styled-ui 🎉 
Thanks for adding this Ian, Dustin 👍

SpencerAlexander:
Is `parseUserDateString` really required if I'm just using the DatePicker? 
I had to do this:
```
// Don't support user date input parsing in date-picker
export const parseUserDateString = dateString => null;
```

Imundy:
@SpencerAlexander That's a good catch. I extracted the propTypes so I could reuse them when they get passed through the different components, but I should have removed that one from the actual date picker. That function is just for turning some string into an actual date and it doesn't ever get called by the DatePicker at the moment (just from the input), so you're correct that it's not required.

PR #72, opened 2018-11-01
Fix modals with really long content overflowing the screen size

EthanRutherford:
Sets a max-width on the modal, so that it can no longer overflow the page on the x axis. Also adds a few other css properties to make sure the content doesn't overflow the modal.

### DIFF
--- a/components/modal/styled.jsx
+++ b/components/modal/styled.jsx
@@ -1,5 +1,6 @@
 import styled from 'styled-components';
 import { thickness, fonts, colors } from '../shared-styles';
+import { resetStyles } from '../utils';
 
 export const Label = styled.label`
 	display: block;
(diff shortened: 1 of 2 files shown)


components/modal/styled.jsx
```diff
@@ -8,9 +8,11 @@ export const Label = styled.label`
 `;
 
 export const Modal = styled.div`
+	box-sizing: border-box;
```
dustinsoftware: This should be using `resetStyles` to get `border-box` set. @Imundy do you remember offhand why that was not done here?

Imundy: I'm not sure if resetStyles was in place when this was component was originally merged? Or it might have been in flux still. But I agree that makes sense.

dustinsoftware:
I'll merge it as is, although in the future I'd like to add support for rendering as a child node of `body`. Copy [this utility](https://github.com/reactstrap/reactstrap/blob/214da8c6fb02ccb1198444a378e85851f366b87a/src/Portal.js) from Reactstrap and follow how it's being used in [this modal](https://github.com/reactstrap/reactstrap/blob/07ec4b52e1bebcbb268a7c283b562bc85c0cf020/src/Modal.js#L363).

EthanRutherford:
Why not just use the built-in react portal to do it?

dustinsoftware:
Take a look at what I linked to in depth. It handles edge cases we may not
have considered.

PR #75, opened 2018-11-07
Update to styled-components v4

korbinancell:
As well as update button component to take advantage of v4 functionality.

### DIFF
--- a/components/input/component.jsx
+++ b/components/input/component.jsx
@@ -54,7 +54,7 @@ export const Input = forwardClassRef(
 				autoFocus,
 				onClick,
 				disabled,
-				forwardedRef,
+				forwardedRef, // eslint-disable-line react/prop-types
 				...inputProps
 			} = this.props;
 
--- a/components/input/component.jsx
+++ b/components/input/component.jsx
@@ -1,80 +1,84 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
-import { applyVariations } from '../utils';
+import { applyVariations, forwardClassRef } from '../utils';
 import * as Styled from './styled';
 
 /** Standard text input with no validation */
-export class Input extends PureComponent {
-	static propTypes = {
-		value: PropTypes.string,
-		placeholder: PropTypes.string,
-		type: PropTypes.string,
-		readOnly: PropTypes.bool,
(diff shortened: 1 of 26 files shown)


components/button/anchor-button.jsx
```diff
@@ -1,16 +1,13 @@
-import React, { PureComponent } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
-import { BaseButton } from './base-button.jsx';
-import * as Styled from './styled.jsx';
+import { Button } from './component.jsx';
```
dustinsoftware: Let's separate out the restructuring of `base-button` out from the v4 changes. I'm not sold on this approach but don't want to block the update.

dustinsoftware:
Since this is a breaking change I'd like to wait until the next time we need to push out a breaking change (preferably at least a month from now).

components/button/anchor-button.jsx
korbinancell: Sure thing. Do you want a separate pr for updating the button or just remove the commit and keep it out for now?

dustinsoftware: Remove it for now, open a separate PR for button after the v4 update gets merged

dustinsoftware:
@awesomebob this should be good to go, please review

components/input/component.jsx
```diff
@@ -1,80 +1,84 @@
...
+					type={type || 'text'}
+					autoFocus={autoFocus}
+					readOnly={readOnly}
+					disabled={disabled}
+					value={value || ''}
+					placeholder={placeholder || ''}
+					onChange={this.handleChange}
+					onClick={onClick}
+					onKeyPress={this.handleKeyPress}
+					ref={forwardedRef}
```
awesomebob: To be clear: this `<Input>` component was formerly *not* forwarding the ref, but we are adding that functionality in this PR, correct?

awesomebob:
Is this a usage of `innerRef` that we'd like to remove as well?
https://github.com/Faithlife/styled-ui/blob/master/components/collapse/component.jsx#L33

components/input/component.jsx
awesomebob: I don't see that we previously exposed `innerRef` in this component, might that have been for a different component?

PR #76, opened 2018-11-07
Use portal to render modals outside of any stacking contexts

EthanRutherford:
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context

### DIFF
--- a/components/modal/modal.jsx
+++ b/components/modal/modal.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { createPortal } from 'react-dom';
 import { ThemeProvider } from 'styled-components';
 import { debouncedResize } from '../utils';
 import { ModalBackdrop } from '../modal-backdrop/component.jsx';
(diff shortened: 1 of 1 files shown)


components/modal/modal.jsx
```diff
@@ -53,11 +54,13 @@ export class Modal extends React.Component {
 
 	state = {
 		modalWidth: null,
+		canUseDom: false,
 	};
 
 	componentDidMount() {
 		const { cancel } = debouncedResize(this.handleResize);
 		this.cancelResizeListener = cancel;
+		this.setState({ canUseDom: true }); // eslint-disable-line
```
dustinsoftware: It would be cleaner to remove use of state and check `typeof window !== 'undefined'` in render

dustinsoftware: Edit: IIRC this strategy doesn't work anymore in React 16, so let's keep state around.

EthanRutherford: Right, that strategy now can cause hydration issues.

PR #77, opened 2018-11-08
Add Popover component

EthanRutherford:
Based primarily on the work already done by Korbin [here](https://github.com/Faithlife/styled-ui/pull/68), but using `material-ui/popper` instead of  `react-popper`. This allows for a simpler API that still provides all the functionality of the original.

### DIFF
--- a/catalog/index.js
+++ b/catalog/index.js
@@ -17,6 +17,7 @@ import {
 	FilesSection,
 	DropZone,
 	SimpleModal,
+	Popover,
 } from '../components/main.js';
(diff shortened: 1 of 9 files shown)


dustinsoftware:
Per offline conversation I’d like to see this use a render prop instead.

EthanRutherford:
No longer uses a handrolled wrapper.

dustinsoftware:
We may need a simpler API in the future, but we are going to roll with #68 first.

PR #81, opened 2018-11-14
Cancel timeouts on popover

korbinancell:


components/popover/component.jsx
```diff
@@ -95,13 +97,24 @@ export class Popover extends React.Component {
 		this.handleIsPopoverShowing(prevProps.isOpen, this.props.isOpen, this.props.delay);
 	}
 
+	componentWillUnmount() {
+		if (this._timeout) clearTimeout(this._timeout);
```
dustinsoftware: Sorry for not adding a lint rule for this. `if` statements should always have curly braces.

components/popover/component.jsx
```diff
@@ -95,13 +97,24 @@ export class Popover extends React.Component {
...
+	}
+
+	handleTimeout = showPopper => () => {
+		this.setState({ showPopper });
```
dustinsoftware: Abstracting this out is a bit overkill, let's just inline it below.

PR #85, opened 2018-11-15
Fix group selector modal box-sizing.

rmjohnson:


rmjohnson:
CC: @dustinsoftware Since the group selector modal is now outside of the catalog's `AppLayout`, it was no longer getting `box-sizing: border-box;` (which it relies on).

components/group-selector/styled.jsx
```diff
@@ -136,6 +136,11 @@ export const LargeScrollView = styled(ScrollArea)`
 	width: 375px;
 	background: white;
 	border-radius: 4px;
+	box-sizing: border-box;
```
dustinsoftware: Why isn’t this coming from the reset styles already?

rmjohnson:
Combined with https://github.com/Faithlife/styled-ui/pull/86

PR #86, opened 2018-11-15
Make title on group selector optional and up z-index of modal-backdrop.

rmjohnson:
CC: @dustinsoftware design has asked us to omit the "Find your Church in the Faithlife Church Directory" title. As well, the modal fix didn't work because the z-index on our elements was still higher than the modal-backdrop's so I increased the z-index there.

### DIFF
--- a/catalog/group-selector/variations.md
+++ b/catalog/group-selector/variations.md
@@ -44,101 +44,99 @@ state: {
 	mobile: false,
 }
 ---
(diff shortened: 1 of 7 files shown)


rmjohnson:
@dustinsoftware, should I just combine this PR with [this one](https://github.com/Faithlife/styled-ui/pull/85) ?

PR #87, opened 2018-11-16
Make updates to match design review

Imundy:
Split parseUserDateString from date-picker props

### DIFF
--- a/components/date-picker-input/component.jsx
+++ b/components/date-picker-input/component.jsx
@@ -13,13 +13,16 @@ export class DatePickerInput extends PureComponent {
 		/** Functions that operate on a JS Date obejct.
 		 * The following functions must be provided:
 		 *
-		 * startOfWeek, startOfMonth, endOfWeek, endOfMonth, getYear, getMonth,getDate, addWeeks, addMonths,s ubMonths, isBefore, format,isValid,parseUserDateString
+		 * startOfWeek, startOfMonth, endOfWeek, endOfMonth, getYear, getMonth,getDate, addWeeks, addMonths,s ubMonths, isBefore, format,isValid
 		 *
 		 * For details on how these function should behave see date-fns documentation (v2) https://date-fns.org
 		 *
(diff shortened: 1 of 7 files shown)


components/date-picker-input/component.jsx
```diff
@@ -13,13 +13,16 @@ export class DatePickerInput extends PureComponent {
...
 		 *
 		 * For details on how these function should behave see date-fns documentation (v2) https://date-fns.org
 		 *
-		 * In addition, parseUserDateString should be a function that takes a string and returns a js date. For example: https://www.npmjs.com/package/chrono-node (depends on momentjs)
 		 */
 		dateFunctions: dateFunctionProps,
+		/**
+		 * parseUserDateString should be a function that takes a string and returns a js date. For example: https://www.npmjs.com/package/chrono-node (depends on momentjs)
+		 */
+		parseUserDateString: PropTypes.func.isRequired,
```
Imundy: @dustinsoftware this is a breaking change, but I'd rather just eat the cost of this now and split these since (afaik) Spencer is the only one using the `DatePicker` right now, and he is *not* using the `DatePickerInput` so it won't affect them.

Imundy: But also we're about to release a major version change soon anyways, so we could just roll this into there.

components/date-picker-input/component.jsx
```diff
@@ -13,13 +13,16 @@ export class DatePickerInput extends PureComponent {
 		/** Functions that operate on a JS Date obejct.
 		 * The following functions must be provided:
 		 *
-		 * startOfWeek, startOfMonth, endOfWeek, endOfMonth, getYear, getMonth,getDate, addWeeks, addMonths,s ubMonths, isBefore, format,isValid,parseUserDateString
+		 * startOfWeek, startOfMonth, endOfWeek, endOfMonth, getYear, getMonth,getDate, addWeeks, addMonths,s ubMonths, isBefore, format,isValid
```
dustinsoftware: There are still typos here

PR #89, opened 2018-11-22
Add experimental table from giving

Imundy:
This is pretty much a direct port of what we have in giving. I'm open to dramatically altering the API, but it was copied over into stores recently so I thought it made sense to add it to styled-ui. I would see this not as a replacement for ag-grid, but as a simpler alternative.

### DIFF
--- a/components/experimental-table/component.jsx
+++ b/components/experimental-table/component.jsx
@@ -3,9 +3,15 @@ import PropTypes from 'prop-types';
 import { ArrowUp } from '../../components/icons';
 import * as Styled from './styled.jsx';
 
+/**
+ * ExperimentalTable:
+ * Simple table component (not meant as a replacement for ag-grid). It takes a set an arry of headings and an arry of rows.
+ * Order matters for headings and rows (data will be rendered in the order given)
+ */
 export class ExperimentalTable extends Component {
 	/* eslint-disable react/no-unused-prop-types */
 	static propTypes = {
+		/** Table headings. This can be null or empty for no headings */
 		headings: PropTypes.arrayOf(
 			PropTypes.shape({
 				heading: PropTypes.string.isRequired,
@@ -14,13 +20,16 @@ export class ExperimentalTable extends Component {
 				sortOrder: PropTypes.string,
 			}),
 		),
+		/** Table data. Each row should be provided with a key for proper list handling in react. See https://reactjs.org/docs/lists-and-keys.html */
 		data: PropTypes.arrayOf(
 			PropTypes.shape({
 				key: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
 				rowData: PropTypes.arrayOf(
 					PropTypes.shape({
 						content: PropTypes.oneOfType([PropTypes.string, PropTypes.number, PropTypes.node]),
+						/** override/cell-specific styling */
 						style: PropTypes.object,
+						className: PropTypes.string,
 					}),
 				),
 				onClickRow: PropTypes.func,
@@ -60,7 +69,11 @@ export class ExperimentalTable extends Component {
 			? []
 			: data.map(({ key, rowData, onClickRow }) => {
 					const tds = rowData.map((cellData, i) => (
-						<Styled.TableCell key={`${key}-${i}`} style={cellData.style}>
+						<Styled.TableCell
+							key={`${key}-${i}`}
+							className={cellData.className}
+							style={cellData.style}
+						>
 							{cellData.content}
 						</Styled.TableCell>
 					));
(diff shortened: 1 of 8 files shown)


Imundy:
@SpencerAlexander You might be interested in this.

SpencerAlexander:
💥 It's used several times in Skeletor in the new payment plans page. I could also see it being used in a future/new Skeletor react cart/checkout flow for simple tabular layouts.

components/experimental-table/component.jsx
```diff
@@ -0,0 +1,99 @@
...
+							{content.onClick ? orderIcon : ''} {content.heading}
+						</Styled.Heading>
+					);
+			  });
+
+		const tableData = !data
+			? []
+			: data.map(({ key, rowData, onClickRow }) => {
+					const tds = rowData.map((cellData, i) => (
+						<Styled.TableCell key={`${key}-${i}`} style={cellData.style}>
```
dustinsoftware: It would be nice to avoid setting `style` directly, but maybe that is unavoidable here.

components/main.js
```diff
@@ -19,3 +19,4 @@ export { Modal, ModalFooter } from './modal';
 export { Popover, PopoverManager, PopoverReference, Tooltip } from './popover';
 export { Radio } from './radio';
 export { SimpleModal } from './simple-modal';
+export { ExperimentalTable } from './experimental-table';
```
dustinsoftware: It would be nice to keep this sorted (and add a comment indicating exports should be sorted)

components/experimental-table/styled.jsx
```diff
@@ -0,0 +1,54 @@
+import styled from 'styled-components';
+import { thickness, colors } from '../shared-styles';
+
+export const Table = styled.table`
```
dustinsoftware: Missing reset styles

components/experimental-table/component.jsx
```diff
@@ -0,0 +1,99 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { ArrowUp } from '../../components/icons';
+import * as Styled from './styled.jsx';
+
+export class ExperimentalTable extends Component {
```
dustinsoftware: Missing documentation

components/experimental-table/component.jsx
```diff
@@ -0,0 +1,99 @@
...
+	/* eslint-disable react/no-unused-prop-types */
```
dustinsoftware: Can this lint rule be re-enabled?

Imundy: No, for some reason the lint rule does not work properly because it is used in the `static getDervivedStateFromProps`

SpencerAlexander:
Just looked at the preview deploy (haven't played around with it in depth yet).
Supposing I'd want the table styles to more closely match this spec:
https://app.zeplin.io/project/5b511c28eb657422b8b29957/screen/5b6b2fc2a9b4bd1bc73a085d

Is that something I could do?. ie
- Header row have white background instead of gray
- remove top header border 
- have right column be `justify-content: flex-end` instead of center

Imundy:
@SpencerAlexander 

I'm not sure about the first couple. Providing hooks to override those styles is definitely doable (though I have not yet), but my questions would just be about what design wants tables to look like. If there's a couple different "themes" for tables, then we might be able to do better by providing a prop that handled all the styles to help guarantee better consistency.

As for the last one you can definitely do that (we do it in giving all the time as well be cause of dollar values being right aligned). You would just need to pass in a style override with the data for that specific cell. Or you can pass in a full react component as the content which we also do occasionally for icons, etc.

SpencerAlexander:
Okay thanks @Imundy! In that case it might be worth roping in @jake-robertson, to see if design would be okay with replacing the existing payment plans tables with this styled-ui version (which has gray header background and top header border) for consistency across our projects or if we'd want to create a separate theme variant as Ian mentioned above as a possible solution.

screen shot 2018-12-07 at 10 50 43 am

jake-robertson:
@SpencerAlexander @Imundy I'm not tracking with what problem we're trying to solve, or what the implications are of the proposed solution, but am happy to help!

Imundy:
@jake-robertson I think we're looking at the differences in tables between the giving specs and the ecommerce sites specs. The notable differences in the ecommerce spec are that the header row is white and there is no top border. So questions are:

* Should we prefer to use one of these everywhere?
* Is the ecommerce table spec a different table "theme", or does it make sense to just let ecommerce override what the header row looks like?

The distinction with themes vs. overrides is one of consistency. We want the components to be flexible, but we also want them to fall back to looking standardardized. If we can encapsulate common table style variations in themes, then I think we do ourselves some favors for helping people do the right thing when they're building stuff out of styled-ui.

dustinsoftware:
Let's re-open this when we have a project actively blocked on this feature.

PR #94, opened 2018-12-03
Fix group creation for in place group selector.

rmjohnson:


components/group-selector/large/component.jsx
```diff
@@ -37,20 +37,23 @@ export class LargeGroupSelector extends React.Component {
...
-	searchResultsRef = React.createRef();
-	fixedCreateWrapper = false;
+		this.state = {
+			searchInputValue: '',
+			newChurchName: '',
+			newChurchLocation: '',
+			modalContent: 'main',
+			selectedGroupId: -1,
+			createGroupFixed: false,
+			resultsTopMargin: props.showInPlace ? 0 : defaultResultsTopMargin,
```
dustinsoftware: I'm pretty sure `this.props` can be referenced using class field syntax.

components/group-selector/large/component.jsx
```diff
@@ -173,6 +176,8 @@ export class LargeGroupSelector extends React.Component {
 	};
 
 	handleScroll = scrollData => {
+		if (this.props.showInPlace) return;
```
dustinsoftware: Note to self: Add a lint rule requiring curly braces.

components/group-selector/large/component.jsx
rmjohnson: i.e. I don't need to put this in a `constructor`? I'll try that. I always thought you had to do a constructor if initial state was based off of props.

components/group-selector/large/component.jsx
```diff
@@ -44,11 +44,10 @@ export class LargeGroupSelector extends React.Component {
 		modalContent: 'main',
 		selectedGroupId: -1,
 		createGroupFixed: false,
-		resultsTopMargin: defaultResultsTopMargin,
+		resultsTopMargin: this.props.showInPlace ? 0 : defaultResultsTopMargin,
 		scrollWidthDelta: 0,
 	};
 
-	modalRef = React.createRef();
```
rmjohnson: This *appeared* to be unused.

rmjohnson:
@dustinsoftware, updated!

PR #98, opened 2018-12-07
Fix overflow bug for churches with long names. SITES-5029

rmjohnson:


components/group-selector/styled.jsx
```diff
@@ -271,6 +271,7 @@ export const SearchResultContent = styled.div`
 `;
 
 export const SearchResultNameText = styled.div`
+	max-width: 275px;
```
rmjohnson: @dustinsoftware, I tried all manner of `max-width: 100%` (on this and on the parent) to try to make this more flexible but this is the only way I could get it to work. Am I missing something?

awesomebob: The `width: 100%` on the grandparent `SearchResultContent` div seems incorrect to me.
Changing the `width: 100%` on that element to `min-width: 100px` (and removing this new max-width line) seems to achieve what you're looking for.

https://github.com/Faithlife/styled-ui/blob/fb915f05d707cda7b11febf73a586d2773efb0ed/components/group-selector/styled.jsx#L268

I'm not 100% clear on *why* that works, but I know that setting a `min-width` changes things as opposed to having a `width: auto`

rmjohnson: @awesomebob, @dustinsoftware, I'm not sure that is what we want (unless I'm misunderstanding you). Changing that `width: 100%` to `min-width: 100px` gives us this:
image

What's the rationale behind the `100px`?

awesomebob: Looks like my suggestion was missing the `flex-grow: 1` part.
Here's a PR where you can check it out (once the preview builds):
https://github.com/Faithlife/styled-ui/pull/109

I still don't understand exactly why it works, but the `min-width` somehow affects the flexbox calculation.

rmjohnson: Thanks @awesomebob ! I'll close this PR in favor of yours.

rmjohnson:
Closing in favor of https://github.com/Faithlife/styled-ui/pull/109

PR #99, opened 2018-12-07
Update button padding to match input heights

awesomebob:
WIP, still unsure of how to properly handle buttons with icons


awesomebob:
@dustinsoftware is there something I can do to trigger a preview deploy?

PR #100, opened 2018-12-07
Scale button icons to match font size

awesomebob:
Uses 1em height and width. Are our icons always guaranteed to be square?


awesomebob:
Confirmed with Shiloh that we're ok with the current icon centering.

dustinsoftware:
@awesomebob go ahead with a merge and ship then

PR #103, opened 2018-12-12
Fix server-render in group selector

dustinsoftware:
@JordanSjodinFaithlife please test locally with `yarn link`


JordanSjodinFaithlife:
Works great! (besides the style lint errors)

PR #104, opened 2018-12-13
Remove extraLarge, set explicit button heights

awesomebob:
The old "large" button size was not in the UI specs, and has now been
removed. The former "extraLarge" is now just "large".

There were no existing usages of "extraLarge" anywhere, and all the
existing usages of "large" have been migrated to be "medium".
All of those usages were in the Giving.Web project.

Firefox scales button border widths in unexpected ways:
https://bugzilla.mozilla.org/show_bug.cgi?id=1427391

> Yeah - this is intended behavior (though it can be a bit surprising).
> We snap border-widths to be exact multiples of device pixels, as noted
> in bug 477157 comment 7.

Since we expect the combination of line-height, padding, and border to
exactly equal certain pixel heights, setting that height explicitly is
not really a major change.

Discussion in FLCOM team chat:
https://beta.faithlife.com/messages/12492?threadId=504803

### DIFF
--- a/components/button/styled.jsx
+++ b/components/button/styled.jsx
@@ -178,15 +178,15 @@ export const variationMap = {
 		}
 `,
 	small: component => component.extend`
+		height: 32px;
 		padding: 7px ${props => (props.condensed ? '7px' : '9px')};
 `,
 	medium: component => component.extend`
+		height: 40px;
 		padding: 11px ${props => (props.condensed ? '11px' : '15px')};
 `,
 	large: component => component.extend`
-		padding: 14px ${props => (props.condensed ? '14px' : '20px')};
-`,
-	extraLarge: component => component.extend`
+		height: 56px;
 		padding: 15px ${props => (props.condensed ? '15px' : '23px')};
 		font-size: 24px;
 `,
--- a/components/button/styled.jsx
+++ b/components/button/styled.jsx
@@ -9,7 +9,7 @@ const buttonColors = {
 	disabled: '#bedcf2',
 };
 
-export const ButtonContents = styled.span`
+export const ButtonContentWrapper = styled.div`
 	display: grid;
 	grid-auto-flow: column;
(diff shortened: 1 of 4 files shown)


Imundy:
> as the 4.2.0 release?

It would be a 5.0 release (major version change). I think even if we weren't removing a button style, it would still be a breaking change as well due to the changes in the heights of existing button styles (just fwiw).

components/button/styled.jsx
```diff
@@ -178,15 +178,15 @@ export const variationMap = {
 		}
 `,
 	small: component => component.extend`
+		height: 32px;
```
Imundy: 🤔 I'm wondering if `min-height` wouldn't be more appropriate here? I'm not sure what the answer to this would be, but what if the content we are putting in the button is larger than what is expected for that font size? This seems to happen occasionally with SVGs and it seems like the more important thing is for the button retain the correct padding than to retain the explicit height.

awesomebob: That's a fair consideration. At this point, we do want SVG icons to match 1em height, and I haven't seen any designs that account for button text wrapping, if the button text is too long that it wraps I believe we're doing something wrong.
My inclination is to keep it as height until we have good reason to move it to min-height, but I could be convinced otherwise.

awesomebob:
@dustinsoftware without objection, I'll release this as a 5.0 major version update

components/button/styled.jsx
awesomebob: Dustin and Eric confirmed that buttons should never wrap, and the text should be elided instead of clipped. I'll get that added to this PR.

Discussion: https://beta.faithlife.com/messages/12492?threadId=504803

dustinsoftware:
Merge and publish as 5.0 when you are ready, let's make sure to get all consumers updated to 5.0 as well so that no new uses of `extraLarge` are introduced.

PR #109, opened 2019-01-15
Fix overflow bug for churches with long names. SITES-5029

awesomebob:


dustinsoftware:
Thanks, this is a much more elegant solution.

PR #119, opened 2019-01-24
Fix checkbox text alignment when text wraps

JeremyEinfeld:
The checkbox component inherits its text-alignment value (center) from the agent style for buttons, a value that isn't apparent until the label is long enough to wrap:

image

This changes allows the checkbox component to inherit its alignment value from parent elements.


dustinsoftware:
That's unfortunate. Thanks for the fix.

PR #122, opened 2019-02-04
Add a simple toast component.

korbinancell:
[Link to spec](https://app.zeplin.io/project/58a23c2d4597afb248c13a43/screen/5bedfd6811d4fc12b2f92d4a)

### DIFF
--- a/components/shared-styles/index.js
+++ b/components/shared-styles/index.js
@@ -25,6 +25,11 @@ export const fonts = {
 		font-weight: normal;
 		line-height: 16px;
 	`,
+	ui18: `
+		font-size: 18px;
+		font-weight: normal;
+		line-height: 18px;
+	`,
 	h4: `
 		font-size: 22px;
 		font-weight: bold;
@@ -84,4 +89,10 @@ export const inputColors = {
 	inputBorderColor: colors.borderColor,
 	inputFocusedBorderColor: '#278ed4',
 	inputFocusedShadowColor: '#d0e6f6',
+};
+
+export const mediaSizes = {
+	desktop: '980px',
+	tablet: '736px',
+	phone: '480px',
 };
--- a/components/shared-styles/index.js
+++ b/components/shared-styles/index.js
@@ -92,7 +92,7 @@ export const inputColors = {
 };
 
 export const mediaSizes = {
-	desktop: '980px',
-	tablet: '736px',
-	phone: '480px',
+	desktop: '992px',
+	tablet: '768px',
+	phone: '579px',
 };
--- a/components/simple-toast/styled.jsx
+++ b/components/simple-toast/styled.jsx
@@ -62,7 +62,7 @@ export const ToastContainer = styled.div`
 
 	${fonts.ui18};
 
-	background-color: ${colors.white};
+	background-color: ${({ theme }) => theme.backgroundColor || colors.white};
 	border-radius: 3px;
 	box-shadow: 0 19px 38px 0 rgba(0, 0, 0, 0.12), 0 15px 12px 0 rgba(0, 0, 0, 0.12);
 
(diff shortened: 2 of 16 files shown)


awesomebob:
retest this please

components/toast/styled.jsx
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	10%, 100% {
+		opacity: 1;
+	}
+`;
+
+const hiddenBottomValue = ({ styleOverrides }) => css`
+	bottom: calc(-1 * (${styleOverrides.height || toastHeight.desktop} + ${toastPadding} * 2));
+`;
```
awesomebob: I like this helper function. I had just finished figuring out what the calculation was supposed to do, and then I see you refactored it. 👍 

I might even go so far as to suggest adding a *comment* saying something like "On desktop, the toast will initially render just offscreen and slide up into view."

components/toast/styled.jsx
```diff
@@ -0,0 +1,130 @@
+import styled, { keyframes, css } from 'styled-components';
+import { fonts, thickness, colors, mediaSizes } from '../shared-styles';
+
+const flcomMobileHeaderHeight = 47; // px
```
awesomebob: I have mixed opinions about including this constant.
This is a public repo that in theory could be used by any project.
In practice, _most_ of its usages will probably be within the context of a faithlife.com view.

components/toast/component.jsx
```diff
@@ -0,0 +1,90 @@
...
+		}
+	};
+
+	render() {
+		const { icon, theme, styleOverrides } = this.props;
+		const { messages, toastState } = this.state;
+
+		return (
+			<Styled.ToastContainer
+				state={toastState}
```
awesomebob: I'll have to admit, seeing a `prop` called _state_ really threw me for a loop when I first saw it, but it makes total sense to me now.

korbinancell:
@jake-robertson could you take a look a look at this? https://deploy-preview-122--faithlife-styled-ui.netlify.com/#/toast/variations

components/toast/styled.jsx
```diff
@@ -0,0 +1,130 @@
...
+
+export const ToastContainer = styled.div`
+	/** Shared Styles */
+	display: grid;
+	grid-auto-flow: column;
+	grid-column-gap: ${thickness.eight};
+	grid-template-columns: min-content max-content;
+
+	position: fixed;
+	z-index: 999999;
```
awesomebob: I'm not a fan of having such a high z-index.
I'd prefer to use a more reasonable number, with an available `styleOverride` in case a particular consumer really needs a higher number.
https://github.com/Faithlife/styled-ui/blob/master/components/modal/modal.jsx#L61

components/toast/styled.jsx
```diff
@@ -0,0 +1,130 @@
...
+
+	${fonts.ui18};
+
+	background-color: ${colors.white};
+	border-radius: 3px;
+	box-shadow: 0 19px 38px 0 rgba(0, 0, 0, 0.12), 0 15px 12px 0 rgba(0, 0, 0, 0.12);
+
+	pointer-events: none;
```
awesomebob: Will this `pointer-events: none` cause problems if a `<Button>` is included as a child?
Could we please add a demo with a button to test?

korbinancell: There is currently no way to attach a button, using the current setup it would need it's own prop. Toast does not respect children.

components/toast/component.jsx
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * A mobile first Toast. A toast indicates that an action is being taken that does not require the user's
+ * specific attention.
+ */
+export class Toast extends PureComponent {
+	static propTypes = {
+		/** SVG Icon */
+		icon: PropTypes.node,
+		/** Toast Content */
+		message: PropTypes.string,
```
awesomebob: Would it be possible to make this `message` prop into a `PropTypes.node` to allow passing in arbitrary contents?
I'm looking in particular at the examples that have buttons and titles in the content:
https://app.zeplin.io/project/58a23c2d4597afb248c13a43/screen/5bedfd6811d4fc12b2f92d4a

components/toast/component.jsx
```diff
@@ -0,0 +1,90 @@
...
+		theme: PropTypes.shape({
+			backgroundColor: PropTypes.string,
+		}),
+		styleOverrides: PropTypes.shape({
+			width: PropTypes.string,
```
awesomebob: The `width` styleOverride isn't used for anything. I might suggest turning it into an override for `minWidth` though.

components/toast/component.jsx
korbinancell: That is an option, but we might loose the protection of a toast with the same message appearing several times in a row. I would also argue that adding a button that isn't a close button would go against our definition of a toast.
> They’re feedback for users to indicate that an action is being taken that doesn’t require their specific attention

components/shared-styles/index.js
```diff
@@ -25,6 +25,11 @@ export const fonts = {
 		font-weight: normal;
 		line-height: 16px;
 	`,
+	ui18: `
```
awesomebob: Let's double-check with design that the 18px font-size was intentional. FLCOM standard is now 16px.

korbinancell: cc @jake-robertson

jake-robertson: Yep, its intentional. I'm not sure the rational (I didn't design it), but it was deliberate.

components/shared-styles/index.js
```diff
@@ -85,3 +90,9 @@ export const inputColors = {
 	inputFocusedBorderColor: '#278ed4',
 	inputFocusedShadowColor: '#d0e6f6',
 };
+
+export const mediaSizes = {
```
awesomebob: These values aren't quite Bootstrap values, do we have these in a design spec somewhere?

We do want to be careful about using these for both min-width and max-width though, you generally do not want `min-width: tablet` and `max-width: tablet` styles to *both* be active at exactly 736px width.
Bootstrap typically handles this by having a separate variable for `max-width: tablet-max` that is half a pixel less than `min-width: desktop`, for example.

I also find it to be generally cleaner if you don't have both `min-width` and `max-width` media queries in the same file. If the file is intended to be mobile-first, as this one is, then the mobile styles can be outside of a media query, and the desktop-specific rules can be in a `min-width` query.

components/toast/component.jsx
awesomebob: Our specs do have uses cases where we'll have a specific single action available in the toast, often an "undo" feature. I have an upcoming use case for it.

components/simple-toast/styled.jsx
```diff
@@ -0,0 +1,120 @@
...
+
+	justify-items: center;
+
+	${props => {
+		switch (props.state) {
+			case TransitionStatuses.ENTERING:
+				return css`
+					transition: opacity ${transitionTime}ms linear;
+					opacity: 0;
+					opacity: 1;
```
awesomebob: Looks like we have some conflicting opacity rules here, unless I'm missing something tricksy with the `css` helper?

PR #123, opened 2019-02-07
Popover click away

korbinancell:
- Add a new component ClickAwayHandler as a wrapper for [react-event-listener](https://www.npmjs.com/package/react-event-listener).
- Add the new component to the popover as well as a new prop to handle click away.

### DIFF
--- a/components/popover/component.jsx
+++ b/components/popover/component.jsx
@@ -3,6 +3,7 @@ import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
 import { ThemeProvider } from 'styled-components';
 import { Popper } from 'react-popper';
+import { ClickAwayHandler } from '../utils';
 import { colors } from '../shared-styles';
 import * as Styled from './styled';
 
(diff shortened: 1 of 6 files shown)


components/utils/click-away-handler.jsx
```diff
@@ -0,0 +1,44 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import ReactEventListener from 'react-event-listener';
+
+export class ClickAwayHandler extends Component {
+	static propTypes = {
+		/** Which mouse events should trigger the handler */
```
awesomebob: I'm not clear on what the use case is for picking a different mouseEvent or touchEvent would be.
If mouseDown and onTouchStart are the clear primary use-case, could we make those into defaultProps?

components/popover/component.jsx
```diff
@@ -135,35 +138,42 @@ export class Popover extends React.Component {
...
-								...(!hideArrow ? Styled.margins[Styled.getPlacement(placement)] : {}),
-							}}
-							delay={delay}
-							onAnimationEnd={this.handleTransition}
-							hideArrow={hideArrow}
-							styleOverrides={styleOverrides}
+						<ClickAwayHandler
+							mouseEvent="onMouseDown"
+							touchEvent="onTouchStart"
+							{...(showPopper && onClickAway ? { onClick: onClickAway } : {})}
```
awesomebob: I realize that you're being careful to *not* add an `onClick: null` prop, but is that necessary?
It looks to me that the ClickAwayHandler checks whether `onClick` is truthy before using it.

components/utils/click-away-handler.jsx
korbinancell: I think there's a use case for wanting to listen to different events depending on the order of things. onMouseDown is before onBlur is before onClick. It has mattered before. I agree on the default props though,

awesomebob: Just need those defaultProps and I'm ready to merge.

PR #129, opened 2019-02-12
Add product drawer control SITES-4522

dustymack:


awesomebob:
1. Looks like there were a few `yarn lint` errors on the build.
2. One of the stated goals for this project is to include [*simple* UI components](https://github.com/Faithlife/styled-ui#goals). At what point would we consider a component "complex" enough to split into a separate project? I would consider the `GroupSelector` component in this category as well. @bryanrsmith what do you think?

components/product-drawer/animated-image/styled.jsx
```diff
@@ -0,0 +1,172 @@
...
+	transform-origin: 0% 50%;
+	animation-delay: 40ms;
+`;
+
+export const AnimatedImage = styled.div`
+	margin: 0 auto;
+	height: 30px;
+	width: 70%;
+
+	${AnimatedImageTextFirst} {
```
dustymack: This syntax appears to be the cause of the lint errors.
`Unexpected duplicate selector ...` followed by `Expected empty line before rule`.

I am depending on the prop passed to the parent (or below, the hover over the parent) to effect change in the child. Is there a better way to handle these situations?  This was a port from a `.less` file so I am still learning Styled Components syntax.

Auk14HP:
I'm seeing an underline on the "More" link at the bottom of the drawer, is that intentional?  It's not there on ebooks.faithlife.com

components/product-drawer/animated-image/styled.jsx
awesomebob: I'm a little stumped as well. It looks like you're using the correct syntax for referring to another component: https://www.styled-components.com/docs/advanced#referring-to-other-components

There might be a slight issue with how you're referencing the keyframe rule, it looks like you'll need to use the `css` helper along with the `keyframes` helper now:
https://www.styled-components.com/docs/faqs#what-do-i-need-to-do-to-migrate-to-v4 (point 6)

bryanrsmith:
> One of the stated goals for this project is to include [_simple_ UI components](https://github.com/Faithlife/styled-ui#goals). At what point would we consider a component "complex" enough to split into a separate project? I would consider the `GroupSelector` component in this category as well.

I'm a little unsure about this as well. I asked the same question, and my impression was that it hadn't been answered because it hadn't come up. I'm inclined to say that so long as the component has a separate output bundle and is generally applicable to a wide range of Faithlife projects I don't see any harm in putting it here. Separate repos for medium-sized components has management overhead that I'd rather avoid paying until there's a good reason.

Some things that I'd say should probably not live in this repo:

* Things that are only applicable to a specific subset of projects within the organization.
* Things that have large or many dependencies that aren't shared with the rest of the components in this library, or aren't widely used by projects with the organization. Things that have large dependencies that are likely to change (e.g., flavor-of-the-month rich text libraries).
* Things that are likely to require frequent breaking changes.

components/product-drawer/animated-image/styled.jsx
awesomebob: Looks like the stylelint processor isn't smart enough to know that the string interpolation will be getting a *different* class name for each different component, it just puts 'dummyValue' in for each one, which results in duplicates.

There is an escape hatch you can use to trick the stylelint processor into knowing they are distinct values, if you add a unique string to a special comment such as this one:
https://github.com/styled-components/stylelint-processor-styled-components/issues/193

awesomebob:
> my impression was that it hadn't been answered because it hadn't come up

This is probably true. This might be a good time to talk about it though.

> Separate repos for medium-sized components has management overhead

There is certainly *some* overhead, although a large part of that will be reduced as we move to embrace Sheepdog.

I would be in favor of at least setting up one new package for complex shared components, and maybe that repo wouldn't need to be on public GitHub as the implementation is that much more Faithlife-specific.

To maybe expand a little on our stated goals in the README, `styled-ui` gives us the official UI building blocks that teams need to implement new specs in their various projects. The proposed new `shared-complex-components` are for complex components that aren't themselves building blocks, they are completely-implemented components that can be re-used across projects.

components/product-drawer/animated-image/styled.jsx
dustymack: I like your solutions better than the `// eslint-disable-next-line no-dupe-keys` hack I just pushed. Making that change now.

components/product-drawer/component.jsx
```diff
@@ -0,0 +1,257 @@
...
+} from './svgs';
+import * as Styled from './styled';
+
+export class ProductDrawer extends React.PureComponent {
+	static propTypes = {
+		isBrandBarEnabled: PropTypes.bool.isRequired,
+		resources: PropTypes.object,
+		customClassNames: PropTypes.shape({
+			toggleButton: PropTypes.string,
+		}),
```
bryanrsmith: This method of styling customization often leads to specificity wars, and makes future styling changes harder because you have no way of knowing what customizations consumers have applied. Consider using either the `theme`, or `styleOverrides` pattern used elsewhere in this repo (e.g., https://github.com/Faithlife/styled-ui/blob/408f8780b0221a39d3e9721f5f500b1fe64db9bf/components/button/base-button.jsx)

components/product-drawer/component.jsx
```diff
@@ -0,0 +1,257 @@
...
+			<Styled.ProductDrawer>
+				<Styled.ProductDrawerToggle
+					className={(customClassNames && customClassNames.toggleButton) || null}
+					onClick={this.handleToggleClick}
+					ref={el => {
+						this.toggle = el;
+					}}
+					tabIndex="-1"
+				>
+					<svg
```
bryanrsmith: Why is this inlined instead of stored in a svg file, as the other icons are?

components/product-drawer/product-link/styled.jsx
```diff
@@ -0,0 +1,122 @@
+import styled, { keyframes } from 'styled-components';
+import { mediaSizes } from '../../shared-styles';
+
+const reveal = keyframes`
```
bryanrsmith: Why not use transitions for these simpler animations?

components/product-drawer/product-link/product-link.jsx
```diff
@@ -0,0 +1,32 @@
...
+		</Styled.ProductLink>
+	</Styled.ProductLinkListItem>
+);
+
+ProductLink.propTypes = {
+	title: PropTypes.string.isRequired,
+	description: PropTypes.string.isRequired,
+	href: PropTypes.string.isRequired,
+	autoFocus: PropTypes.boolean,
+	children: PropTypes.array,
```
bryanrsmith: Should this be `PropTypes.node`?

components/product-drawer/styled.jsx
```diff
@@ -0,0 +1,268 @@
...
+export const ProductDrawerToggle = styled.button`
+	display: flex;
+	padding: 0;
+	align-items: center;
+	cursor: pointer;
+	background: transparent;
+	border: none;
+
+	&:focus {
+		outline: none;
```
bryanrsmith: Is there an option that doesn't completely disable focus indication?

components/product-drawer/styled.jsx
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	100% {
+		top: 20px;
+	}
+`;
+
+export const ProductDrawer = styled.div`
+	position: relative;
+	font-family: 'source_sans', sans-serif;
```
bryanrsmith: Where did this font name come from? I'd expect `'Source Sans Pro'`.

components/product-drawer/product-link/styled.jsx
```diff
@@ -0,0 +1,122 @@
...
+	&:hover {
+		background-color: #f5f5f5;
+
+		${/* sc-custom 'ProductLink' */ ProductLink} {
+			${/* sc-custom 'ProductLinkTitle' */ ProductLinkTitle} {
+				color: #1e91d6;
+			}
+
+			${/* sc-custom 'ProductLinkDescription' */ ProductLinkDescription} {
+				color: #7a7a7a;
```
bryanrsmith: There are a lot of colors, drop shadows, and timing functions hard coded and repeated. Should they use values from https://github.com/Faithlife/styled-ui/blob/408f8780b0221a39d3e9721f5f500b1fe64db9bf/components/shared-styles/index.js, and/or be moved to variables within this component so they're not repeated and can be easily changed?

components/product-drawer/component.jsx
```diff
@@ -0,0 +1,257 @@
...
+							fillRule="evenodd"
+						/>
+					</svg>
+					<Styled.ProductDrawerToggleText
+						className={(customClassNames && customClassNames.toggleButtonText) || null}
+					>
+						{resources.products}
+					</Styled.ProductDrawerToggleText>
+				</Styled.ProductDrawerToggle>
+				{isOpen ? (
```
bryanrsmith: Should we consider lazy loading the popup contents? How large is the output bundle?

bryanrsmith:
> I would be in favor of at least setting up one new package for complex shared components

For what purpose? What benefits do you anticipate?

catalog/product-drawer/variations.md
```diff
@@ -0,0 +1,38 @@
...
+```react
+showSource: true
+---
+<ProductDrawerDemo>
+	<ProductDrawer
+		isBrandBarEnabled={true}
+		resources={({products: 'Alternate Product text'})}
+	/>
+</ProductDrawerDemo>
+```
```
bryanrsmith: Does this component make any assumptions about the layout of the hosting page? If so, we should document them.

components/product-drawer/animated-image/styled.jsx
```diff
@@ -0,0 +1,171 @@
...
+	transform-origin: 0% 50%;
+	animation-delay: 40ms;
+`;
+
+export const AnimatedImage = styled.div`
+	margin: 0 auto;
+	height: 30px;
+	width: 70%;
+
+	${/* sc-custom 'AnimatedImageTextFirst' */ AnimatedImageTextFirst} {
```
bryanrsmith: Forgive my ignorance, but what are these comments doing?

awesomebob: See this earlier comment on this PR: https://github.com/Faithlife/styled-ui/pull/129#discussion_r256088945

awesomebob:
> > I would be in favor of at least setting up one new package for complex shared components
> 
> For what purpose? What benefits do you anticipate?

I guess my main questions are:
1. How do we define what a "simple" component is?
2. To what extent is it a worthwhile goal for this project to have only "simple" UI components?

components/product-drawer/animated-image/styled.jsx
dustymack: They are keeping the stylelint processor from getting confused. See [conversation above](https://github.com/Faithlife/styled-ui/pull/129#discussion_r256088945)

components/product-drawer/component.jsx
```diff
@@ -0,0 +1,257 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import defaultResources from './locales/en-US/resources.json';
```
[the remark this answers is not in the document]
dustymack: I was thinking that having the defaults there in English would smooth the initial (and most common) use path.  Otherwise `<ProductDrawer />` will be unusable. What are the risks to providing default strings I am missing?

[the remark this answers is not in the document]
bryanrsmith: 1. It bloats the bundle with code that may never be used.
2. It allows consumers to remain blissfully unaware of localization requirements or concerns, creating technical debt. If they want to *explicitly* take on that debt, they can statically import the en-US strings themselves and pass them in to this component.

components/product-drawer/component.jsx
dustymack: It looks like it was originally done this way in order to change the fill color via css.

components/product-drawer/product-link/styled.jsx
dustymack: I'm not sure why it was done this way originally. Is there a disadvantage here? The readability is nice.

components/product-drawer/product-link/product-link.jsx
dustymack: Reading that `PropTypes.node` is less specific and represents "Anything that can be rendered". Is there additional meaning I am missing?

components/product-drawer/styled.jsx
dustymack: Good question for design, I can ask for something.

components/product-drawer/styled.jsx
dustymack: Possibly from Skeletor? Making the change.

components/product-drawer/component.jsx
dustymack: Where can I check the size of the output bundle?

catalog/product-drawer/variations.md
dustymack: The `brandBarEnabled` was built on an assumption of the hosting page layout. I broke that out into the `mobileTopOffset` styleoverride

components/product-drawer/component.jsx
bryanrsmith: Run a release build, and look at the generated files in dist/.

components/product-drawer/product-link/product-link.jsx
bryanrsmith: No; but isn't `PropTypes.node` a more appropriate representation of what's supported here? It seems like this component supports any children.

components/product-drawer/product-link/styled.jsx
bryanrsmith: Not really, other than the added complexity and dependency on the keyframes helper. keyframe animations are usually only used for complex animations that can't be done with transitions.

components/product-drawer/component.jsx
dustymack: 59 KB

dustymack:
@awesomebob @bryanrsmith Any final thoughts? I have some folks waiting on the package so I would love to get this pushed forward ASAP. I appreciate the GREAT review notes!

components/product-drawer/component.jsx
bryanrsmith: Sorry; I was being rhetorical. I think we should treat images consistently, and move this to a file. The title is also inappropriate on this element, and should move to the parent button.

components/product-drawer/component.jsx
bryanrsmith: That's pretty large for something that's likely to never be opened. I think it's worth code splitting and lazy loading the popup contents.

components/product-drawer/styled.jsx
bryanrsmith: Has this been answered? We shouldn't ship inaccessible shared components.

PR #133, opened 2019-02-22
Prevent clicking on descendants of Buttons

awesomebob:
There is no good reason for a _child_ of a button component to be clickable, or for the click event to take on the `value` attribute of that child element. I'm suggesting that we add `pointer-events: none;` to the ButtonContentWrapper to address this situation.

Preview link demonstrating the problem:
https://deploy-preview-134--faithlife-styled-ui.netlify.com/#/button/variations?a=the-problem

### DIFF
--- a/components/button/styled.jsx
+++ b/components/button/styled.jsx
@@ -14,6 +14,7 @@ export const ButtonContentWrapper = styled.div`
 	grid-auto-flow: column;
 	grid-column-gap: 6px;
 	align-items: center;
+	pointer-events: none;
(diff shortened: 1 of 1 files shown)


awesomebob:
Any objection to this @bryanrsmith?

bryanrsmith:
That's pretty standard behavior. Why wouldn't this code use `event.currentTarget.value`?

awesomebob:
Thanks Bryan, that's exactly what I should use instead.

PR #135, opened 2019-02-22
Add Typescript support

korbinancell:
There are a few issues to look into.
* The tsconfig will definitely need to be expanded
* The docgen tables are a little ugly. A lot of the assumptions we made when we were using prop-types are not true for the typescript docgen.
* Updates need to be made to linters to include typescript rules/practices.
* Testing with mocha also seems to be broken atm.
* I'm not completely sure what the trade off are for using babel to parse so additional typechecking/ compilation might be neccesary.
* I'm sure there are additional things that I have missed.


.babelrc
```diff
@@ -1,28 +1,85 @@
...
+			{
+				"include": "components.*\\.tsx$",
+				"exclude": "catalog\\.tsx$"
+			}
+		],
 		"react-docgen",
-		"babel-plugin-transform-flow-strip-types"
+
+		// Stage 0
+		"@babel/plugin-proposal-function-bind",
```
bryanrsmith: We should stick to more stable transforms. Probably just `preset-env`, `preset-react`, and `preset-typescript`.

.babelrc
```diff
@@ -1,28 +1,85 @@
...
 			{
 				"modules": false,
 				"targets": { "browsers": [">1%", "not op_mini all"] }
 			}
-		],
-		"stage-0"
+		]
 	],
 	"plugins": [
+		"@babel/plugin-transform-flow-strip-types",
```
bryanrsmith: What's this for?

catalog.config.js
```diff
@@ -0,0 +1,22 @@
+const extensions = ['.ts', '.tsx', '.js', '.json', '.jsx'];
+const tsRule = {
+	test: /\.(ts|tsx)$/,
+	include: ['C:\\Code\\Styled-Ui', 'C:\\Code\\Styled-Ui\\catalog'],
```
bryanrsmith: Is this necessary? If so, we need to make it relative, and not windows-specific.

catalog.config.js
```diff
@@ -0,0 +1,22 @@
...
+	exclude: /node_modules/,
+	loader: 'babel-loader',
+	options: { babelrc: true, presets: [], cacheDirectory: true },
+};
+
+module.exports = {
+	webpack: catalogWebpackConfig => {
+		const modifiedConfig = { ...catalogWebpackConfig };
+		modifiedConfig.resolve.extensions = extensions;
+		modifiedConfig.module.rules[0].oneOf.splice(
```
bryanrsmith: A comment would be helpful here.

bryanrsmith: This is pretty fragile. Can we locate the rule by load name, or something?

components/popover/component.tsx
```diff
@@ -1,87 +1,65 @@
-import React from 'react';
+import React, { RefObject, Component, MouseEvent, TouchEvent } from 'react';
```
bryanrsmith: This is personal preference, but I prefer not to expand the import to all the types because it gets hard to keep track of everything while reading through the code. e.g., `TouchEvent` is a global constructor, and when reading code I have to think about which `TouchEvent` we're referring to. I'd revert this line, and use `React.TouchEvent` instead.

components/popover/component.tsx
```diff
@@ -1,87 +1,65 @@
...
 import { ThemeProvider } from 'styled-components';
+import { Placement, Modifiers } from 'popper.js';
 import { Popper } from 'react-popper';
 import { ClickAwayHandler } from '../utils';
 import { colors } from '../shared-styles';
 import * as Styled from './styled';
 
+type PopoverPlacements = Exclude<Placement, 'auto' | 'auto-start' | 'auto-end'>;
+
+export type Props = {
```
bryanrsmith: We typically prefer interfaces over types. We should migrate to `eslint-config-faithlife`, which will help with stuff like this.

components/popover/component.tsx
```diff
@@ -1,87 +1,65 @@
...
+	/** Is the popover open */
+	isOpen?: boolean;
```
bryanrsmith: Are these really all optional? I'd expect this prop to be required. (I do see it was optional in the proptypes).

components/popover/component.tsx
```diff
@@ -1,87 +1,65 @@
...
+	/** Where on the target the popover renders */
+	placement?: PopoverPlacements;
+	/** Refer to https://popper.js.org/popper-documentation.html#modifiers for documentation*/
+	modifiers?: Modifiers;
+	/** Where to inject the popover. Defaults to inline */
+	container?: string | RefObject<any> | (() => RefObject<any>);
```
bryanrsmith: Maybe `React.RefObject<HTMLElement>`?

components/popover/component.tsx
```diff
@@ -1,87 +1,65 @@
...
+		backgroundColor?: string;
+		textColor?: string;
+	};
+};
+
+const initialState = {
+	showPopper: false,
+};
+
+type State = Readonly<typeof initialState>;
```
bryanrsmith: It's weird to reference the same initial state object from every instance of the component. Instead, define a type (interface) for the state that describes its shape, and then create a new object in the property initializer.

components/popover/component.tsx
```diff
@@ -1,87 +1,65 @@
...
+
 /** Positioning helper used to display content above another element.
  *  Refs are not supported, please use PopoverManager and PopoverReference to handle positioning.
  */
-export class Popover extends React.Component {
-	static propTypes = {
```
bryanrsmith: Sadly, removing these is a significant breaking change, since we have existing projects that rely on them. Some day hopefully we can either generate them, or decide it's worth it to remove them in a major release.

components/popover/component.tsx
```diff
@@ -1,87 +1,65 @@
...
 	};
 
-	state = {
-		showPopper: this.props.isOpen || false,
-	};
+	readonly state: State = initialState;
 
-	_timeout = null;
+	private timeout: number | null = null;
+	private targetContainer: Element | null = null;
```
bryanrsmith: Probably an `HTMLElement`

components/popover/component.tsx
```diff
@@ -102,29 +80,29 @@ export class Popover extends React.Component {
 	}
 
 	componentWillUnmount() {
-		if (this._timeout) {
-			clearTimeout(this._timeout);
+		if (this.timeout) {
```
bryanrsmith: I know it was already here, but this check is unnecessary.

components/popover/popper-helpers.tsx
```diff
@@ -1,10 +1,17 @@
-import React from 'react';
-import PropTypes from 'prop-types';
+import React, { SFC, MouseEvent } from 'react';
```
bryanrsmith: `SFC` is deprecated, since hooks allow function components to be stateful. Use `React.FunctionComponent` instead.

components/popover/popper-helpers.tsx
```diff
@@ -1,10 +1,17 @@
...
 import { Manager, Reference } from 'react-popper';
 import * as Styled from './styled';
 
+type referenceProps = {
```
bryanrsmith: This confusingly overlaps with local variable names, and it appears unnecessary. Just define a Props type whose properties are optional.

components/popover/component.tsx
bryanrsmith: Naming it `PopoverProps` would save every consumer of this component from having to rename the import. I generally always name the prop and state types with the name of the component they belong with.

components/utils/click-away-handler.tsx
```diff
@@ -0,0 +1,52 @@
...
+	children?: ReactNode;
+};
+
+export class ClickAwayHandler extends Component<Props> {
+	static defaultProps: Props = {
+		mouseEvent: 'onMouseDown',
+		touchEvent: 'onTouchStart',
+	};
+
+	private targetRef: RefObject<any> = React.createRef();
```
bryanrsmith:
```js
private targetRef = React.createRef<HTMLDiv>();
```

Should give you a more specific type with less code.

package.json
```diff
@@ -100,9 +121,13 @@
...
 		"stylelint-processor-styled-components": "^1.3.1",
 		"svg-colorize-loader": "^0.1.2",
+		"typescript": "^3.2.4",
 		"url-loader": "^1.0.1",
 		"webpack": "^4.15.1",
 		"webpack-cli": "^3.0.8",
 		"webpack-node-externals": "^1.7.2"
+	},
+	"resolutions": {
+		"babel-core": "7.0.0-bridge.0"
```
bryanrsmith: What uses this? I see that you added `@babel/core` as well, and we shouldn't have both.

tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+		"sourceMap": true,
+		"strictNullChecks": true,
+		"module": "es6",
+		"jsx": "react",
+		"target": "es5",
+		"allowJs": true,
+		"moduleResolution": "node",
+		"allowSyntheticDefaultImports": true,
+		"declaration": true
+	},
```
bryanrsmith: Looks like we're still using babel to actually build (👍), so we can probably remove `allowJs. It'd be great to set `strict` as well.

.babelrc
korbinancell: https://github.com/Faithlife/styled-ui/blob/master/components/text-input-v2/react-select-async.jsx is forked from an unmerged branch of react-select-async and is flow typed.

catalog.config.js
korbinancell: 🤦‍♂️ It is not.

package.json
korbinancell: I forget which but a couple of the older dependencies specifically require `babel-core` vs `@babel/core` the bridge version basically just imports `@babel/core` and exports it so that everything is happy.

components/popover/component.tsx
korbinancell: Good to know. I will investigate using [babel-plugin-typescript-to-proptypes](https://github.com/milesj/babel-plugin-typescript-to-proptypes) to hopefully save a little on complexity.

.babelrc
bryanrsmith: 😱 That sounds *really* bad. Why are we doing that?

package.json
bryanrsmith: We should figure out what they are and upgrade them, or move to libraries that are more actively maintained.

components/popover/component.tsx
awesomebob: I know that it's difficult when it's a matter of preference, but I'd be interested in having some set guidelines around when it's appropriate to import `TouchEvent` versus using `React.TouchEvent` or `createRef` versus `React.createRef`. I don't want to get into a bikeshedding argument here, I'm hoping we can set a guideline once and avoid bikeshedding about it in the future.

bryanrsmith: Opened https://git/Logos/eslint-config/issues/29

PR #136, opened 2019-02-25
Fix issue at date picker, typeahead, modal

Yunfeng01:
- Customize placement of date picker input
- Modal at mobile viewport is 100% max-height minus small margins
- Add z-index to Pop-over
- Remove z-index from Typeahead dropdown arrow


components/date-picker-input/component.jsx
```diff
@@ -28,6 +28,20 @@ export class DatePickerInput extends PureComponent {
 		onChange: PropTypes.func.isRequired,
 		onFocus: PropTypes.func,
 		disabled: PropTypes.bool,
+		placement: PropTypes.oneOf([
```
korbinancell: This should probably be moved into the `popper-helpers.jsx` file and exported so we can keep them consistent.

components/modal/modal.jsx
```diff
@@ -99,8 +99,10 @@ export class Modal extends React.Component {
 			(modalWidth < 220 ||
 				(modalWidth < 320 && footerProps && Object.keys(footerProps).length === 3));
 
+		const isMobileViewport = modalWidth < 768;
```
korbinancell: [Written spec (flowdock)](https://www.flowdock.com/app/logos/faithlife-giving/threads/TvpZ7dO7niYjsJ21LAwKKNDSsCk)
[Modal spec](https://app.zeplin.io/project/58a23c2d4597afb248c13a43/screen/5ad9263c3b2a774f0136ed77?did=5c6f529f172d221ad84290f7)

components/popover/styled.jsx
```diff
@@ -29,6 +29,7 @@ export const PopoverContent = styled.div`
 	text-align: center;
 	border: ${({ styleOverrides }) => (styleOverrides.border ? styleOverrides.border : 'none')};
 	box-shadow: ${({ styleOverrides }) => (styleOverrides.hideShadow ? 'none' : colors.boxShadow)};
+	z-index: 3;
```
korbinancell: This isn't necessary because `zIndex` is a valid `styleOverride`.

components/modal/styled.jsx
```diff
@@ -14,7 +14,7 @@ export const Modal = styled.div`
 	width: fit-content;
 	height: fit-content;
 	max-width: calc(100% - 16px);
-	max-height: 80%;
+	max-height: ${props => props.theme.isMobileViewport ? `calc(100% - 16px)` : `80%`};
```
korbinancell: nit: 
```suggestion
	max-height: ${props => props.theme.isMobileViewport ? `calc(100% - ${thickness.sixteen})` : '80%'};
```

components/date-picker-input/component.jsx
```diff
@@ -141,7 +155,7 @@ export class DatePickerInput extends PureComponent {
 						</Styled.CalendarIconContainer>
 					</Styled.CalendarButton>
 					<Popover
-						placement="bottom-start"
+						placement={placement || "bottom-start"}
```
korbinancell: To be contestant and so it will show up in the docgen tables, let's add this as a default prop.

components/modal/modal.jsx
awesomebob: Consider referencing this shared constant:
https://github.com/Faithlife/styled-ui/blob/master/components/shared-styles/index.js#L96

package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@faithlife/styled-ui",
-	"version": "5.0.0-beta9",
+	"version": "5.0.0-beta10",
```
bryanrsmith: Please revert this -- version numbers should be updated independently and on the master branch immediately before a release is triggered.

components/date-picker-input/component.jsx
```diff
@@ -30,6 +31,23 @@ export class DatePickerInput extends PureComponent {
...
+			zIndex: PropTypes.number,
+		}),
+	};
+
+	static defaultProps = {
+		placement: 'bottom-start',
+		styleOverrides: {
+			width: '100%',
+			padding: '16px 20px',
+			zIndex: 3,
```
Yunfeng01: Default to 3 is preventing from some input (z-index is 3 by default) overlapping it.

awesomebob: It looks like we're passing this whole `styleOverrides` object to *both* the `<Input>` component and the `<Popover>` component, but we're relying on those components silently ignoring the overrides that we don't want to apply.

For example, the `Popover` component has a padding override, but `Input` currently does not. If someone else adds a padding override to the `Input` component down the road, this `DatePickerInput` component will suddenly have too much padding.

Let's consider being more explicit about which overrides are applied to which component.
An example of this would be how the `Modal` component specifies that it passes the `zIndex` override to the backdrop component:
https://github.com/Faithlife/styled-ui/blob/f0d5c1568b1eab0065117287c519e307799d2bc0/components/modal/modal.jsx#L28

Yunfeng01: Good catch. I also find this just right after it, so it only applies to popover, but I think you are right, I should say the comment more clear

awesomebob: To be clear: I'm suggesting an update to the comment *and* an update to the code to pass only the specific overrides that we want to pass to each component.

PR #137, opened 2019-02-26
Improve click-away-handling for popovers and add Dropdown component.

korbinancell:
These might should be in two different pr's but the Dropdown component depends on the popovers having the improved click-away.

Here is the spec I have been using for dropdowns: https://app.zeplin.io/project/58a23c2d4597afb248c13a43/screen/59ee876e4582a457c39dbefe

### DIFF
--- a/components/dropdown/component.jsx
+++ b/components/dropdown/component.jsx
@@ -1,0 +1,54 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { PopoverManager } from '../popover';
+import { DropdownContext } from './dropdown-helpers';
+
+export class Dropdown extends Component {
+	static propTypes = {
+		children: PropTypes.node,
+	};
+
+	constructor(props) {
+		super(props);
+
+		this.handleToggleDropdown = () => {
+			this.setState(({ context }) => ({
+				context: { ...context, isDropdownOpen: !context.isDropdownOpen },
+			}));
+		};
+
+		this.handleCloseDropdown = onClickHandler => event => {
+			this.setState(({ context }) => ({ context: { ...context, isDropdownOpen: false } }));
+			onClickHandler(event);
+		};
+
+		this.state = {
+			context: {
+				handleToggleDropdown: this.handleToggleDropdown,
+				handleCloseDropdown: this.handleCloseDropdown,
+				isDropdownOpen: false,
+			},
+		};
+	}
+
+	handleClickAway = () => {
+		const {
+			context: { isDropdownOpen },
+		} = this.state;
+
+		if (isDropdownOpen) {
+			this.setState(({ context }) => ({ context: { ...context, isDropdownOpen: false } }));
+		}
+	};
+
+	render() {
+		const { children } = this.props;
+		const { context } = this.state;
+
+		return (
+			<DropdownContext.Provider value={context}>
+				<PopoverManager onClickAway={this.handleClickAway}>{children}</PopoverManager>
+			</DropdownContext.Provider>
+		);
+	}
+}
--- a/components/dropdown/dropdown-helpers/dropdown-item.jsx
+++ b/components/dropdown/dropdown-helpers/dropdown-item.jsx
@@ -1,0 +1,44 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import * as Styled from '../styled';
+import { DropdownContext } from './dropdown-context';
+
+export const dropdownItemTheme = {
+	textColor: PropTypes.string,
+	hoverTextColor: PropTypes.string,
+	hoverBackgroundColor: PropTypes.string,
+};
+
+/**
+ * Must be placed inside of a `DropdownMenu` component
+ */
+export function DropdownItem({ onClick, localTheme, styleOverrides, children }) {
+	return (
+		<DropdownContext.Consumer>
+			{({ handleCloseDropdown }) => (
+				<Styled.DropdownItem
+					onClick={handleCloseDropdown(onClick)}
+					localTheme={localTheme}
+					styleOverrides={styleOverrides}
+				>
+					{children}
+				</Styled.DropdownItem>
+			)}
+		</DropdownContext.Consumer>
+	);
+}
+
+DropdownItem.propTypes = {
+	onClick: PropTypes.func.isRequired,
+	/** This will edit just the single item */
+	localTheme: PropTypes.shape(dropdownItemTheme),
+	styleOverrides: PropTypes.shape({
+		padding: PropTypes.string,
+	}),
+	children: PropTypes.node,
+};
+
+DropdownItem.defaultProps = {
+	localTheme: {},
+	styleOverrides: {},
+};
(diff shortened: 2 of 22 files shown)


components/dropdown/component.jsx
```diff
@@ -0,0 +1,54 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { PopoverManager } from '../popover';
+import { DropdownContext } from './dropdown-helpers';
+
+export class Dropdown extends Component {
```
bryanrsmith: A hooks function would simplify this implementation significantly. I don't think we should be writing any new components using classes.

components/dropdown/dropdown-helpers/dropdown-item.jsx
```diff
@@ -0,0 +1,44 @@
...
+					{children}
+				</Styled.DropdownItem>
+			)}
+		</DropdownContext.Consumer>
+	);
+}
+
+DropdownItem.propTypes = {
+	onClick: PropTypes.func.isRequired,
+	/** This will edit just the single item */
```
bryanrsmith: Can you clarify? I'm not sure what this means.

components/dropdown/component.jsx
```diff
@@ -0,0 +1,54 @@
...
+	static propTypes = {
+		children: PropTypes.node,
+	};
+
+	constructor(props) {
```
awesomebob: Is there a functional difference between these arrow function handlers defined within this constructor versus the `handleClickAway` one defined with class property syntax? Could we move these two methods and state out and remove the need for the explicit constructor?

components/dropdown/dropdown-helpers/dropdown-menu.jsx
```diff
@@ -0,0 +1,54 @@
...
+					{children}
+				</Popover>
+			)}
+		</DropdownContext.Consumer>
+	);
+}
+
+DropdownMenu.propTypes = {
+	...dropdownMenuProps,
+	/** `childrenTheme applies the styles to every `DropdownItem`. This has the same shape as `DropdownItem's` `localTheme` */
```
bryanrsmith: `DropdownItem's` -> `DropdownItem`'s

components/dropdown/component.jsx
```diff
@@ -0,0 +1,54 @@
...
+				handleCloseDropdown: this.handleCloseDropdown,
+				isDropdownOpen: false,
+			},
+		};
+	}
+
+	handleClickAway = () => {
+		const {
+			context: { isDropdownOpen },
+		} = this.state;
```
awesomebob: Consider writing this as:
`const { isDropdownOpen } = this.state.context;`

components/dropdown/dropdown-helpers/dropdown-menu.jsx
```diff
@@ -0,0 +1,54 @@
...
+				>
+					{children}
+				</Popover>
+			)}
+		</DropdownContext.Consumer>
+	);
+}
+
+DropdownMenu.propTypes = {
+	...dropdownMenuProps,
```
bryanrsmith: You're not props spreading, so this is not right. You should explicitly specify the proper you accept. Better yet, let the Popover component validate its own props instead of copying them onto this component.

components/dropdown/dropdown-helpers/dropdown-menu.jsx
```diff
@@ -0,0 +1,54 @@
...
+	styleOverrides,
+	theme,
+}) {
+	return (
+		<DropdownContext.Consumer>
+			{({ isDropdownOpen }) => (
+				<Popover
+					isOpen={isDropdownOpen}
+					placement={placement}
+					hideArrow={hideArrow !== false}
```
bryanrsmith: It seems pretty confusing for this to potentially behave differently than `Popover`'s `hideArrow` prop. Is there a reason not to pass it through directly?

components/dropdown/dropdown-helpers/dropdown-toggle.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { PopoverReference } from '../../popover';
+import { DropdownContext } from './dropdown-context';
+
+/**
+ * Must be placed inside of a `Dropdown` component
+ */
+export function DropdownToggle({ children }) {
```
bryanrsmith: What if a consumer wants to programmatically toggle a dropdown? Consider also exposing a the toggle callback in a more general-purpose way than a button.

components/utils/index.js
```diff
@@ -92,3 +92,24 @@ export const resetStyles = `
...
+	font-family: inherit;
+	cursor: pointer;
+	text-align: center;
+	transition: none;
+	-webkit-appearance: none;
+	-moz-appearance: none;
+
+	&:hover,
+	&:focus,
+	&:active,
```
bryanrsmith: Is this complete?

components/dropdown/component.jsx
```diff
@@ -0,0 +1,54 @@
...
+				context: { ...context, isDropdownOpen: !context.isDropdownOpen },
+			}));
+		};
+
+		this.handleCloseDropdown = onClickHandler => event => {
+			this.setState(({ context }) => ({ context: { ...context, isDropdownOpen: false } }));
+			onClickHandler(event);
+		};
+
+		this.state = {
```
bryanrsmith: We shouldn't abuse state as a memoization mechanism. In a class component, memoize-one can clean this up. I think we should switch to a hooks function and use `useMemo`, though.

components/dropdown/component.jsx
```diff
@@ -0,0 +1,54 @@
...
+			context: {
+				handleToggleDropdown: this.handleToggleDropdown,
+				handleCloseDropdown: this.handleCloseDropdown,
+				isDropdownOpen: false,
```
awesomebob: Could we consider supporting an `isInitiallyOpen` prop?

components/utils/click-away-handler.jsx
```diff
@@ -10,6 +15,8 @@ export class ClickAwayHandler extends Component {
 		touchEvent: PropTypes.oneOf(['onTouchStart', 'onTouchEnd']),
 		/** Callback for when the element is clicked away from */
 		onClick: PropTypes.func,
+		/** Optional ref or array of refs that are also inbounds */
+		inboundsElements: PropTypes.oneOfType([reactRef, PropTypes.arrayOf(reactRef)]),
```
bryanrsmith: How happy are we with this outer-click approach? It seems pretty broken for keyboard users. Adding a major new component that depends on the functionality seems like a good opportunity to fix it rather than spread the problem.

components/dropdown/component.jsx
bryanrsmith: > Is there a functional difference between these arrow function handlers defined within this constructor versus the `handleClickAway` one defined with class property syntax?

Technically, the class property transform uses Object.defineProperty with arguments that result in a slightly different property. Realistically, no, there's no difference.

Doesn't matter though; we shouldn't use a class!

korbinancell:
Closing in favor of https://github.com/Faithlife/styled-ui/pull/154

korbinancell:
Closing in favor of https://github.com/Faithlife/styled-ui/pull/154

PR #146, opened 2019-04-01
Change height of group selector search results to min-height with height: auto.

rmjohnson:


rmjohnson:
@bryanrsmith, do you have merge power? Looks like I'm not authorized.

rmjohnson:
@bryanrsmith, wanted to make sure this was still on your radar.

PR #151, opened 2019-04-11
Add tab component.

korbinancell:
A in depth take on tabs featuring keyboard and screen reader support.


components/tabs/use-tab-list.js
```diff
@@ -0,0 +1,22 @@
...
+	const [tabList, setTabList] = useState([]);
+
+	const registerTab = useCallback(name => {
+		setTabList(prevState => [...prevState, name]);
+	}, []);
+
+	const editTabName = useCallback(
+		(index, newName) => {
+			const newList = [...tabList];
+			newList.splice(index, 1, newName);
```
bryanrsmith: No objection to this, but just to give you an alternative: I like to use `.map` and `.filter` for immutable collection changes so I don't to worry about mutation.

```js
const newList = tabList.map((tab, i) => i === index ? newName : tab);
```

components/tabs/tab-helpers.jsx
```diff
@@ -0,0 +1,59 @@
+import React, { useContext } from 'react';
+import PropTypes from 'prop-types';
+import { TabContext, useKeyboardNav } from './tab-utils';
+import * as Styled from './styled.jsx';
+
+export function TabList({ children }) {
+	const { onSelectTab, selectedTabIndex, panelsContainerRef, theme, styleOverrides } = useContext(
```
bryanrsmith: Instead of exporting the full context from tab-utils, export a custom hook (`useTabContext`, or something) that hides the use of context. This lets you retain some flexibility in implementation, and simplifies components that need the context. https://github.com/Faithlife/CodingGuidelines/blob/master/sections/react.md#use-custom-hooks-to-expose-context

components/tabs/tab.jsx
```diff
@@ -0,0 +1,44 @@
+import React, { useCallback, useRef, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import * as Styled from './styled.jsx';
+
+export function Tab({ children, disabled, ...otherProps }) {
+	const { index, selected, onSelectTab, theme, styleOverrides } = otherProps;
```
bryanrsmith: I find this split destructuring harder to read. I suggest accepting `props` as the only argument, and using a single destructuring expression.

components/tabs/styled.jsx
```diff
@@ -0,0 +1,123 @@
...
+	&:focus {
+		box-shadow: inset 0 0 0 0.2rem rgba(30, 145, 214, 0.5);
+		outline: none;
+	}
+`;
+
+export const TabPanel = styled.div.attrs({
+	role: 'tabpanel',
+	id: ({ index }) => `panel:${index}`,
+	'aria-labelledby': ({ index }) => `tab:${index}`,
```
bryanrsmith: This breaks as soon as there are two instances on the page.

components/tabs/tab-helpers.jsx
```diff
@@ -0,0 +1,59 @@
+import React, { useContext } from 'react';
+import PropTypes from 'prop-types';
+import { TabContext, useKeyboardNav } from './tab-utils';
+import * as Styled from './styled.jsx';
```
bryanrsmith: Unnecessary file extension.

catalog/tabs/variations.md
```diff
@@ -0,0 +1,67 @@
...
+</TabDemo>
+```
+
+### Style Variations
+
+```react
+showSource: true
+---
+<TabDemo>
+	<TabManager theme={{ tabHighlightColor: 'plum' }} styleOverrides={{ width: '200px' }}>
```
bryanrsmith: It's unclear what the styleOverrides apply to. Width of what? Consider accepting styleOverrides on the elements they apply to instead. (`theme` totally makes sense where it is)

components/tabs/styled.jsx
```diff
@@ -0,0 +1,123 @@
...
+	flex-direction: row;
+
+	& > *:not(:last-child) {
+		margin-right: ${thickness.eight};
+	}
+`;
+
+export const TabPanels = styled.div.attrs({ tabIndex: -1 })`
+	width: 100%;
+	height: 100%;
```
bryanrsmith: I wonder if this container styling will be appropriate for most consumers. Do we have places where panels need to be sized to their contents, for example? It seems low-risk, since a theme or styleOverride could easily be added to control it.

components/tabs/styled.jsx
```diff
@@ -0,0 +1,123 @@
...
+	transition: all 0.25s ease 0s;
```
bryanrsmith: What are we transitioning here? The focus style? Consider reducing it from `all` to `box-shadow`, so that we don't accidentally trigger expensive layout-thrashing animations if a consumer changes padding or something.

components/tabs/styled.jsx
```diff
@@ -0,0 +1,123 @@
...
+
+	&:focus {
+		box-shadow: inset 0 0 0 0.2rem rgba(30, 145, 214, 0.5);
```
bryanrsmith: I'm a little surprised to see a focus styling on a container. Why is the container focusable at all? I'd expect focus to move inside the active panel instead.

components/tabs/tab.jsx
```diff
@@ -0,0 +1,44 @@
...
+				selected={selected}
+			>
+				{children}
+			</Styled.TabContent>
+		</Styled.Tab>
+	);
+}
+
+Tab.propTypes = {
+	/** What will be displayed on the tab, usually text */
```
bryanrsmith: This is a little ambiguous, and I don't think it should make assumptions. I've implemented far more tabs labeled by icons than text. Maybe something like "The tab's label"?

components/tabs/tab.jsx
```diff
@@ -0,0 +1,44 @@
...
+	return (
+		<Styled.Tab disabled={disabled} index={index} selected={selected} onClick={handleSelectTab}>
+			<Styled.TabContent
+				ref={tabRef}
+				disabled={disabled}
+				theme={theme}
+				styleOverrides={styleOverrides}
+				selected={selected}
+			>
+				{children}
```
bryanrsmith: I think it would be worth accepting either a node or a function here, to support deferring load until the tab content is visible. Something like `{typeof children === function ? children({ selected, disabled }) : children}`, plus adjustments to `propTypes` should do the trick.

components/tabs/tab.jsx
korbinancell: I tried and the linter catches on that that they are not in the propTypes declaration. These are props that we want hidden from consumers as they will be overridden.

components/tabs/styled.jsx
korbinancell: 🤦‍♂️  Do you recommend we pick up something like Reach-Ui does [here](https://github.com/reach/reach-ui/blob/master/packages/auto-id/src/index.js) or use an id generator like shortid?

components/tabs/tab.jsx
bryanrsmith: Use an eslint-disable comment, with an additional comment explaining why it's necessary.

components/tabs/styled.jsx
korbinancell: I thought it would be a nice way to offer the focus styling for free to a consumer. I can change it to focus the panel if you think that is best.

components/tabs/styled.jsx
bryanrsmith: The simple solution seems fine. We just need to make sure IDs are generated on mount and not on render, so server rendering doesn't break. Note Reach's effect to accomplish this.

components/tabs/styled.jsx
bryanrsmith: Sorry, I don't understand. These are divs that contain the tab contents, right? I wouldn't expect those to be focusable at all.

components/shared-hooks/use-basic-map.js
```diff
@@ -0,0 +1,17 @@
+import { useState, useCallback } from 'react';
+
+export function useBasicMap(initialState) {
+	const [map, setMap] = useState(initialState || {});
+
+	return {
+		map,
+		clear: useCallback(() => setMap({}), []),
+		add: useCallback((key, value) => setMap(prevState => ({ ...prevState, [key]: value })), []),
+		remove: useCallback(key =>
```
bryanrsmith: Specify deps.

components/tabs/tab-helpers.jsx
```diff
@@ -0,0 +1,74 @@
...
+
+export function TabPanel(props) {
+	// PropType linting is diabled so out hidden props can be destuctured along with own consumer props
+	// eslint-disable-next-line react/prop-types
+	const { children, selected, registerPanelId, unRegisterPanelId, index } = props;
+	const id = useId();
+
+	useEffect(
+		() => {
+			registerPanelId(index, id);
```
bryanrsmith: Is it safe (or useful) to register a bunch of null panel IDs? Perhaps either this code, or the registerPanelId implementation should bail out on nulls.

components/tabs/tab-helpers.jsx
```diff
@@ -0,0 +1,74 @@
...
+			return () => unRegisterPanelId(index);
```
bryanrsmith: Be deliberate with your return types. Add a block body to this cleanup function to clarify that it doesn't return a useful result. (There are a few other places in the PR that could benefit from similar changes)

components/shared-hooks/use-id.js
```diff
@@ -0,0 +1,17 @@
...
+let id = 1;
+function genId() {
+	return id++;
+}
+
+export function useId() {
+	const [currentId, setCurrentId] = useState(null);
+
+	useEffect(() => {
+		setCurrentId(genId());
```
bryanrsmith: I usually prefer to use strings for identifiers to clarify that they should be considered opaque. Maybe toString the ID here?

components/tabs/tab-utils.js
```diff
@@ -0,0 +1,77 @@
+import React, { useCallback, useContext } from 'react';
+import { useBasicMap } from '../shared-hooks';
+
+const handledKeys = Object.freeze({
```
bryanrsmith: Freezing seems unnecessary, since it's a private object.

components/tabs/tab-utils.js
```diff
@@ -0,0 +1,77 @@
...
+	arrowRight: 'ArrowRight',
+	arrowLeft: 'ArrowLeft',
+	arrowDown: 'ArrowDown',
+	home: 'Home',
+	end: 'End',
+});
+
+export const TabContext = React.createContext();
```
bryanrsmith: Avoid exporting the whole context. Export the Provider, and a custom hook that hides the implementation detail from consumers.

components/tabs/tab-utils.js
```diff
@@ -0,0 +1,77 @@
...
+				}
+				case handledKeys.end: {
+					onSelectTab(enabledTabIndexes[enabledTabIndexes.length - 1]);
+					break;
+				}
+				default:
+					return;
+			}
+		},
+		[selectedIndex, onSelectTab, children],
```
bryanrsmith: `children` will change on every render, making the `useCallback` useless. Consider storing them in a ref and retrieving the current children inside the handler, letting you remove `children` from deps.

PR #153, opened 2019-04-24
Amber lightbox

VirginiaLP:
This PR adds the Amber Lightbox component. I'm using the PR builder to have design look at it then I'll formally ask for code review.


components/amber-lightbox/tab-section.jsx
```diff
@@ -0,0 +1,60 @@
...
+			if (amberRef.current && window.amberfile) {
+				window.amberfile.embedded.load({
+					url: '/embed/',
+					container: amberRef.current,
+					groupId: vaultId,
+					multiSelect: false,
+					viewStyle,
+					filter,
+				});
+				window.addEventListener('message', onMessageRef.current);
```
bryanrsmith: The dependency chain is a bit broken here. If the `onMessage` prop changes, onMessageRef.current will be rebound, but since the event is bound to the reference set at the time this effect runs, events will be routed to the old handler instead of the new one. It'd be more correct to rerun an effect that rebinds the handler to the most recent handler, which would also let you remove the ref.

components/amber-lightbox/tab-section.jsx
```diff
@@ -0,0 +1,60 @@
...
+import React, { useRef, useCallback, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import * as Styled from './styled';
+
+export function TabSection({ vaultId, filter, viewStyle, onFileSelected }) {
+	const onMessageRef = useRef();
+
+	const onMessage = useCallback(
+		event => {
+			if (event.data) {
```
bryanrsmith: This isn't safe; you need to verify that the message came from the frame you're trying to talk to, and not some other frame on the page.

components/amber-lightbox/styled.jsx
```diff
@@ -0,0 +1,64 @@
+import styled from 'styled-components';
+import { colors } from '../shared-styles';
+
+export const Container = styled.div`
+	background: ${colors.white};
+	width: 800px;
+	height: 600px;
+`;
+
+export const Title = styled.h1`
```
bryanrsmith: h1 is a semantic element that only makes sense in the context of an entire document that has no other h1 (same with h2, h3; those are document-level headings). In a React component, that's an unsafe assumption to make about the consuming code. It'd be better to use `section` and `header` elements, which convey semantic meaning without making assumptions about complete document structure.

components/amber-lightbox/component.jsx
```diff
@@ -0,0 +1,142 @@
...
+import { DropZone } from '../drop-zone';
+import { TabManager, TabList, TabPanels, Tab, TabPanel } from '../tabs';
+import { TabSection } from './tab-section';
+import { CameraSvg } from './svgs';
+import * as Styled from './styled';
+
+// TODO standardize localization
+// TODO implement multiple selection feature
+
+export function AmberLightbox({
```
bryanrsmith: Is this a reasonable name, or should we use something more generic that doesn't assume use of Amber? `FilePicker`, or something?

Does any part of this component make sense outside of a lightbox? I could imagine someone wanting a file upload in a sidebar, or tab of an existing UI. Should the file picker be separated from the lightbox? It's easy for a consumer to put it in a lightbox if that's what they want.

components/amber-lightbox/component.jsx
```diff
@@ -0,0 +1,142 @@
...
+							</TabPanel>
+						))}
+				</TabPanels>
+			</TabManager>
+		</Styled.Container>
+	);
+}
+
+AmberLightbox.propTypes = {
+	tabs: PropTypes.arrayOf(
```
bryanrsmith: This looks very Amber-specific, and I'm concerned that the API isn't flexible enough to accommodate potential future file-selector tab implementations. We had talked about providing child components that could let a consumer more directly control the tabs included; did you decide you didn't like that approach?

From what I recall, we talked about something like…

```jsx
<FilePicker onFileSelected={handleFileSelected} fileTypes={fileTypes} {...etc}>
	<FilePicker.FileUpload title="File Upload" />
	<FilePicker.AmberVault title={group.title} vaultId={123} />
	<FilePicker.AmberVault title="Stock Photos" vaultId={456} />
</FilePicker>
```

veepower:
Right now I have the catalog using my user ID and the Example Community Church Amber vault. Should we make a new user just for this purpose?

KyleKennaw:
We can use the Stock Media bucket for the demo: 6817140 — not sure if we have an existing pattern for users.

catalog/index.html
```diff
@@ -1,28 +1,35 @@
...
+		<meta http-equiv="X-UA-Compatible" content="IE=edge" />
+		<meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no" />
+		<link rel="shortcut icon" href="%PUBLIC_URL%/favicon.png" />
+		<link href="https://fonts.googleapis.com/css?family=Source+Sans+Pro" rel="stylesheet" />
+		<link href="https://cdnjs.cloudflare.com/ajax/libs/normalize/8.0.1/normalize.min.css" />
+		<script
+			src="https://browser.sentry-cdn.com/4.4.1/bundle.min.js"
+			crossorigin="anonymous"
+		></script>
+		<script src="https://amber.faithlife.com/scripts/api/embeddedBucket.js"></script>
```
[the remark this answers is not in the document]
veepower: Should this just be included in the picker's jsx or injected with a portal? Or something different?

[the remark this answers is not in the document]
bryanrsmith: A custom hook that injects the script, waits for it to load, and then returns the exposed global would probably work well.

components/amber-lightbox/tab-section.jsx
veepower: It it admissible to check the origin property of the event to make sure it contains 'amber.faithlife.com'?

AuresaNyctea:
@bryanrsmith we need this functionality. Can you let me know why you closed it? I'm re-opening it.

bryanrsmith:
@AuresaNyctea I closed a handful of PRs that seemed dead. This one has quite a few requested changes, and it didn't seem like it was being actively worked on. I'm happy for it to stay open if that's not the case; just trying to keep things tidy.

components/amber-lightbox/component.jsx
veepower: `FilePicker` +1

components/amber-lightbox/amber-content.jsx
```diff
@@ -0,0 +1,62 @@
+import React, { useRef, useCallback, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import * as Styled from './styled';
+
+export function AmberContent({
+	vaultId,
```
ddunkin: I recommend `accountId`. Amber doesn't use the term "vault" in its API.

components/amber-lightbox/component.jsx
veepower: This component is made with the intention of bringing Amber into more places. That is why I've written it to be Amber-specific.

components/amber-lightbox/component.jsx
```diff
@@ -0,0 +1,142 @@
...
+import { TabSection } from './tab-section';
+import { CameraSvg } from './svgs';
+import * as Styled from './styled';
+
+// TODO standardize localization
+// TODO implement multiple selection feature
+
+export function AmberLightbox({
+	tabs,
+	onFileSelected,
```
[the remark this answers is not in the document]
ddunkin: Amber will return an array of assets. `onAssetsSelected` would be more accurate.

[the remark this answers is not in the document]
ddunkin: Actually, it depends on the `pickerMode`. Most use cases will set the `pickerMode` to `asset`, which will return an array of assets. Other options are `filter` and `file`.

components/drop-zone/component.jsx
```diff
@@ -43,6 +43,7 @@ export class DropZone extends PureComponent {
 	handleDrop = e => {
 		this.dropTargets = 0;
 		this.setState({ showHighlight: false });
+		console.log(e);
```
bryanrsmith: Remove debug log.

components/amber-lightbox/component.jsx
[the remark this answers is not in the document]
veepower: I'd like to keep `onFilesSelected` to keep it consistent with the language of uploading files from your computer which are not `assets`.

components/file-picker/amber-content.jsx
```diff
@@ -0,0 +1,65 @@
...
+	accountId,
+	filter,
+	viewStyle,
+	onFilesSelected,
+	allowMultiSelect,
+	onCancel,
+}) {
+	const handleMessage = useCallback(
+		event => {
+			if (event.origin.includes('amber.faithlife.com') && event.isTrusted) {
```
bryanrsmith: Why `isTrusted`?

It's probably not a big deal in this case, but when verifying origins it's safer to place a constraint specifically on the end of the domain. The current check would pass for the origin `amber.faithlife.com.malicious.ru`.

components/file-picker/amber-content.jsx
```diff
@@ -0,0 +1,65 @@
...
+	useEffect(() => {
+		window.addEventListener('message', handleMessage);
+
+		return () => {
+			window.removeEventListener('message', handleMessage);
+		};
+	}, [handleMessage]);
+
+	useEffect(() => {
+		if (amberRef.current && window.amberfile) {
```
bryanrsmith: It'd be cleaner to consolidate all of the `window` access in a hook, so component code can use `const amber = useAmber()` and not rely on where it comes from.

components/file-picker/component.jsx
```diff
@@ -0,0 +1,80 @@
...
+import { TabManager, TabList, TabPanels, Tab, TabPanel } from '../tabs';
+import { AmberContent } from './amber-content';
+import * as Styled from './styled';
+import { UploadPage } from './upload-page';
+
+export function FilePicker({
+	tabs,
+	onFilesSelected,
+	allowMultiSelect,
+	localizationProps,
```
bryanrsmith: Typically we name this prop `localizedResources`.

components/file-picker/component.jsx
```diff
@@ -0,0 +1,80 @@
...
+	/** Is selecting multiple files allowed */
+	allowMultiSelect: PropTypes.bool,
+	/** Customized text */
+	localizationProps: PropTypes.shape({
+		addText: PropTypes.string,
+		cancelText: PropTypes.string,
+	}),
+	/** This function handles exiting the file picker */
+	onCancel: PropTypes.func.isRequired,
+	/** The minimum allowed file size; optional */
```
bryanrsmith: Can this be more strictly specified? The file size in bytes? Why is it a string? I don't understand the value used in the example either.

components/file-picker/component.jsx
```diff
@@ -0,0 +1,80 @@
...
+	allowMultiSelect,
+	localizationProps,
+	onCancel,
+	minFileSize,
+}) {
+	return (
+		<Styled.Container>
+			<TabManager>
+				<TabList>
+					<Tab>{allowMultiSelect ? 'Upload Files' : 'Upload File'}</Tab>
```
bryanrsmith: Any reason not to add strings like this to the existing object of localized resources?

components/file-picker/image-preview.jsx
```diff
@@ -0,0 +1,28 @@
...
+
+	const handleLoad = useCallback(() => {
+		setIsLoaded(true);
+	}, []);
+
+	return (
+		<Styled.BorderBox>
+			<Styled.Image src={imageSource} onLoad={handleLoad} />
+			<Styled.CloseButton onClick={onRemoveFile}>
+				<img alt={'close button'} src={CloseButtonSvg} />
```
bryanrsmith: Alt text should be a suitable replacement for the image, not a description of the image. "close" would probably be more appropriate.

components/file-picker/image-preview.jsx
```diff
@@ -0,0 +1,28 @@
...
+export function ImagePreview({ imageSource, onRemoveFile }) {
+	const [isLoaded, setIsLoaded] = useState(false);
+
+	const handleLoad = useCallback(() => {
+		setIsLoaded(true);
+	}, []);
+
+	return (
+		<Styled.BorderBox>
+			<Styled.Image src={imageSource} onLoad={handleLoad} />
```
bryanrsmith: onLoad will fire multiple times if the imageSource changes. Should this handle that? You need to handle load errors as well. Seems like we'll show a never-ending spinner if the images fails to load here.

components/file-picker/lightbox-drop-zone.jsx
```diff
@@ -0,0 +1,64 @@
...
+				{showDetails && (
+					<Styled.SvgContainer>
+						<img alt="drag and drop icon" src={CameraSvg} />
+					</Styled.SvgContainer>
+				)}
+				<Styled.DragDropText>{dragDropText}</Styled.DragDropText>
+				<Button primaryTransparent onClick={handleBrowseFilesClick}>
+					{browseText}
+				</Button>
+				<br />
```
bryanrsmith: Please use CSS instead of non-semantic elements.

components/file-picker/upload-page.jsx
```diff
@@ -0,0 +1,134 @@
...
+export function UploadPage({
+	allowMultiSelect,
+	onFilesSelected,
+	localizationProps,
+	onCancel,
+	minFileSize,
+}) {
+	const [selectedFiles, setSelectedFiles] = useState([]);
+
+	const handleRemoveFile = useCallback(
```
bryanrsmith: Memoizing this with useCallback isn't helping you because every invocation is producing a new function, and that new function is the thing being passed as a prop.

components/file-picker/upload-page.jsx
```diff
@@ -0,0 +1,134 @@
...
+	const addFile = useCallback(files => {
+		setSelectedFiles(currentSelectedFiles => [
+			...currentSelectedFiles,
+			...files.map(file => ({ file, dataUrl: URL.createObjectURL(file) })),
+		]);
+	}, []);
+
+	if (selectedFiles.length) {
+		return (
+			<React.Fragment>
```
bryanrsmith: Prefer `<>` if you don't need a `key`.

components/file-picker/lightbox-drop-zone.jsx
```diff
@@ -0,0 +1,64 @@
...
+				<Styled.FileInputLabel>
+					<input
+						ref={fileInputRef}
+						onChange={handleUploadWithFileInput}
+						multiple={allowMultiSelect ? 'multiple' : ''}
+						type="file"
+						name="file"
+					/>
+				</Styled.FileInputLabel>
+				{showDetails && minFileSize && <Styled.MinSizeLabel>{minFileSize}</Styled.MinSizeLabel>}
```
bryanrsmith: Ah, now I see. Why not use `localizedResources`? What's enforcing this minimum?

components/file-picker/styled.jsx
```diff
@@ -0,0 +1,148 @@
...
+`;
+
+export const MinSizeLabel = styled.section`
+	font-size: 16px;
+	color: ${colors.gray34};
+	margin-top: 24px;
+`;
+
+export const SvgContainer = styled.div`
+	margin-top: 109px;
```
bryanrsmith: That's very specific. Why?

components/file-picker/upload-page.jsx
```diff
@@ -0,0 +1,134 @@
...
+	onCancel: PropTypes.func.isRequired,
+	minFileSize: PropTypes.string,
+};
+
+function ButtonSection({ selectedFiles, onFilesSelected, localizationProps, onCancel }) {
+	const onSelect = useCallback(() => {
+		onFilesSelected(selectedFiles);
+	}, [onFilesSelected, selectedFiles]);
+
+	const addText = (localizationProps && localizationProps.addText) || 'Insert';
```
bryanrsmith: It'd be better to inject default localizations in the root component instead of adding defaults at all the sites that use them.

components/file-picker/lightbox-drop-zone.jsx
vpettit: I will add this to localizedResources. I also heard from design that we should just have a recommended, not enforced, size minimum. I will updated to reflect that.

components/file-picker/component.jsx
vpettit: Most of these questions answered below; As for "800 x 400", I chose that because that is what it says on the zeppelin spec.

components/amber-lightbox/component.jsx
AuresaNyctea: @bryanrsmith Virginia and I are confused as to what you're trying to accomplish here. Can you help me understand?

My understanding of how it works currently (thanks to Virginia's explanation) is that you can control which `amber` tabs you'd like to show by passing in the objects to the `tabs` array.

In the interest of shipping and maintaining the ability to be future facing would it work for you if we renamed `tabs` to `amberTabs` and in the future if we have some other type of media solution that we'd like to use here we can add a property then?

bryanrsmith: The goal is to avoid building an Amber-centric API for a component that could easily support broader use cases. There's no reason this couldn't be used in scenarios that called for a file picker without an Amber picker, or for apps that want a tab for choosing an image from a Smart Media Editor instance, or whatever.

> In the interest of shipping and maintaining the ability to be future facing would it work for you if we renamed tabs to amberTabs and in the future if we have some other type of media solution that we'd like to use here we can add a property then?

I don't think that would alleviate my concerns-- The design of the current API doesn't allow for any possibility of extensibility by consumers, or code splitting to support other uses cases, or tab-specific configuration.

What is the objection to the API posted above? It seems like a fairly straight-forward change that would significantly improve flexibility. It's a cleaner API, even if we don't plan on extending it. Virginia seemed on board with it when we spoke in April.

components/amber-lightbox/component.jsx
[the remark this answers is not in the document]
bryanrsmith: I'm confused about the data passed to this callback. Can this not receive either references to local files, or Amber assets?

components/amber-lightbox/component.jsx
vpettit: >The goal is to avoid building an Amber-centric API

When the case was assigned to me it was titled "Build shared Amber lightbox component." I guess I misunderstood exactly what you suggested when we spoke in April, but I've been under the impression the whole time that this should be specific to Amber and adding Amber integrations more places is the whole driving force behind the case. I can make the change to the API but it seems against the point of the component to me.

components/file-picker/amber-content.jsx
```diff
@@ -0,0 +1,65 @@
...
+			window.amberfile.embedded.load({
```
ddunkin: This should set `pickerMode` to `asset` by default.

components/amber-lightbox/component.jsx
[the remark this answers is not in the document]
ddunkin: Files and assets are separate entities in Amber. It's confusing if assets are called files.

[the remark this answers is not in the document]
ddunkin: If there were other non-Amber tab implementations, what would be passed to the callback? Maybe it should be a discriminated union?
```ts
{
  assets?: Asset[];
  files?: File[];
}
```

components/file-picker/component.jsx
```diff
@@ -0,0 +1,58 @@
...
+FilePicker.propTypes = {
+	/** This function handles your files once selected */
+	onFilesSelected: PropTypes.func.isRequired,
+	/** Is selecting multiple files allowed */
+	allowMultiSelect: PropTypes.bool,
+	/** Customized text */
+	localizedResources: PropTypes.shape({
+		addText: PropTypes.string,
+		cancelText: PropTypes.string,
+		reccomendedMinSize: PropTypes.string,
```
JeremyEinfeld: Typo (here and in the docs).

components/file-picker/amber-content.jsx
```diff
@@ -0,0 +1,93 @@
...
+
+	const handleMessage = useCallback(
+		event => {
+			if (event.origin.includes('amber.faithlife.com') && event.isTrusted) {
+				if (event.data && typeof event.data === 'string') {
+					const parsedEvent = JSON.parse(event.data);
+					if (parsedEvent.canceled) {
+						onCancel();
+					} else if (parsedEvent.assets) {
+						onFilesSelected(parsedEvent.assets);
```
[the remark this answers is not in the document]
VirginiaLP: Is this still necessary if it will only ever return files OR assets? Not a mix of both. In that case would it be acceptable to just have it return a `files[]` or a `assets[]`?

components/file-picker/image-preview.jsx
VirginiaLP: I'll add the error handling where it stops trying to load the image and lets the consumer decide what to do. Faithlife.com would probably want to add an error banner but I know other teams/sites do other things. @bryanrsmith what's the risk of not handling if the image source changes?

bryanrsmith: I can only guess at the behavior by reading the code, but it seems like if the user changes the selected image this component would continue showing the previously-selected image while the new one loads (instead of the loading indicator). Some of the uses are keyed, so you shouldn't see any issue there, but the unkeyed single-image case could exhibit undesirable behavior.

components/file-picker/amber-content.jsx
```diff
@@ -0,0 +1,98 @@
...
+		window.addEventListener('message', handleMessage);
+
+		return () => {
+			window.removeEventListener('message', handleMessage);
+		};
+	}, [handleMessage]);
+
+	useEffect(() => {
+		if (amberRef.current && amber) {
+			amber.embedded.load({
```
ddunkin: Set `pickerMode`. It should be a prop that defaults to `asset`. https://git.faithlife.dev/Logos/AssetDesk/blob/master/docs/EmbeddingAnAmberPicker.md#pickermode

components/file-picker/amber-content.jsx
```diff
@@ -0,0 +1,98 @@
...
+				url: '/embed/',
+				container: amberRef.current,
+				groupId: accountId,
```
ddunkin: Use `accountId`. https://git.faithlife.dev/Logos/AssetDesk/blob/master/docs/EmbeddingAnAmberPicker.md#accountid

catalog/index.html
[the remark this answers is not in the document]
ronnywoods: Noted that `embeddedBucket.js` determines whether to use amber.faithlife.com or internal.amber.faithlife.com iframe based on parent url (e.g. there is a regex that checks if parent url contains localhost, internal, test, pr, etc and if so it uses the internal.amber...).  FWIW, I ran into an issue w/ a PR builder url that doesn't prepend 'pr' on the PR number...thus, iFrame didn't use internal even though I was logged into my internal account on the parent page...caused SSO to fail, access denied, etc.  Perhaps an edge case but wanted to mention in case there are other ideas for how to determine use of internal.amber... vs just amber...

components/file-picker/styled.jsx
```diff
@@ -0,0 +1,148 @@
+import styled from 'styled-components';
+import { colors } from '../shared-styles';
+import { resetStyles } from '../utils';
+
+export const Container = styled.div`
+	background: ${colors.white};
+	width: 800px;
```
ronnywoods: From what I can tell, this and other explicit height / width properties here:
- Cause the amber page to always show in small-screen view, even if viewing on a large screen, and
- Don't allow the component to scale down on a mobile device
- Result in empty vertical space if tabs aren't used (e.g. if just consuming the Amber iframe in the FilePicker)

components/file-picker/amber-content.jsx
```diff
@@ -0,0 +1,98 @@
+import React, { useRef, useCallback, useEffect, useState } from 'react';
+import PropTypes from 'prop-types';
+import { useFilePickerContext } from './file-picker-helpers';
+import * as Styled from './styled';
+
+export function AmberContent({ accountId, filter, viewStyle }) {
```
ronnywoods: One of the strange UI things I noticed (while may not be addressable within this component, but mentioning in case) is that in the Amber iframe a user is not limited to uploading only items that match the filter that's passed in.  So if I try to upload a non-image, it lets me, but then will not display to me the uploaded item so that it can be inserted.

catalog/file-picker/variations.md
```diff
@@ -0,0 +1,63 @@
...
+					</Tab>
+					<Tab>{'Stock photos'}</Tab>
+				</TabList>
+				<TabPanels>
+					<TabPanel>
+						<FileUpload/>
+					</TabPanel>
+					<TabPanel>
+						<AmberContent
+							accountId={6817140}
```
ronnywoods: Probably just a placeholder, but when I ran catalog locally I (probably stating the obvious) didn't have access to this account so just got 'access denied' in the catalog

catalog/file-picker/variations.md
```diff
@@ -0,0 +1,63 @@
...
+		uploadFiles: 'Upload Files',
+		addText: 'Insert',
+		cancelText: 'Cancel',
+		dragDropText: 'Drag and drop to upload image',
+		browseText: 'or browse files',
+	},
+}
+---
+<div>
+	<FilePickerDemo>
```
ronnywoods: Would this generally live inside a modal?  If so, should modal be built into the component itself by default, or at least used in the example so that it's implemented consistently?

catalog/file-picker/variations.md
```diff
@@ -0,0 +1,63 @@
...
+				</TabList>
+				<TabPanels>
+					<TabPanel>
+						<FileUpload/>
+					</TabPanel>
+					<TabPanel>
+						<AmberContent
+							accountId={6817140}
+							filter={'kind:"image"'}
+							viewStyle={'tinygrid'}
```
ronnywoods: Allow for viewstyle changes in small viewport vs large viewport (e.g. 'tinygrid' on large and 'table' on small)?  Or maybe that's fine left to the consumer...

components/file-picker/amber-content.jsx
ddunkin: Interesting. I'll open an issue for Amber.

catalog/file-picker/variations.md
AuresaNyctea: You're correct, this is for testing. It is the faithlife stock media group. You should probably follow this group =)

catalog/file-picker/variations.md
AuresaNyctea: @ronnywoods I'm not sure what you're suggesting here? Could you give a bit more information?

ronnywoods: Just looks like “table” view is probably always the appropriate view on mobile, so wondering if the component should always use that view in a small viewport rather than relying on the passed in viewStyle.

components/file-picker/amber-content.jsx
AuresaNyctea: @ddunkin I believe I have done this, can you confirm?

ddunkin: Can you add a default?
```js
AmberContent.defaultProps = {
  pickerMode: 'asset',
};
```

catalog/file-picker/variations.md
AuresaNyctea: I think I'm okay with leaving it outside the modal. I don't believe we will be using the styled-ui modal for this (as the tabs are baked in) so I'm okay with leaving the consumer to decide how to place it.

catalog/file-picker/variations.md
AuresaNyctea: The amber modal itself is aware of its size and changes view based on how much space it has.

components/file-picker/amber-content.jsx
ddunkin: We're also adding an optional `sort` string parameter. Shipping today.

AuresaNyctea: mkay I'll update tomorrow then?

components/file-picker/amber-content.jsx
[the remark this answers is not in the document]
AuresaNyctea: @bryanrsmith can you speak to this?

components/file-picker/amber-content.jsx
AuresaNyctea: so you're suggesting we put `if (event.origin.endsWith` instead of `includes`?

components/file-picker/amber-content.jsx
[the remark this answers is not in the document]
bryanrsmith: Yes, it's still necessary.

components/amber-lightbox/component.jsx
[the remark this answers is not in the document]
AuresaNyctea: `discriminated union` is this a fancy word for object? I've googled it and only come up with typescript references, now I know it's been a while since I've written javascript but it doesn't look to me like this project is using typescript. https://basarat.gitbooks.io/typescript/docs/types/discriminated-unions.html @ddunkin  or @bryanrsmith could you please give me some more information here?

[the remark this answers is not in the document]
AuresaNyctea: with a little more googling I found this article https://medium.com/@justintormey/write-beautiful-js-with-union-types-ddd11e5e9241 I think it does what you're suggesting so I'll endevour to attempt that solution.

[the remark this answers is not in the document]
ddunkin: In this case, just an object that will have one of the properties `assets` or `files` but never both. It may or may not also have a tag that indicates which type it is, e.g. `{ kind: 'assets', assets: [] }`.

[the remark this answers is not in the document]
AuresaNyctea: the but never both is just code convention right? Okay so it is just a fancy word for a specific type of object :) thanks

components/file-picker/amber-content.jsx
```diff
@@ -0,0 +1,93 @@
...
+
+			const amberScript = document.createElement('script');
+			amberScript.src = 'https://amber.faithlife.com/scripts/api/embeddedBucket.js';
+			amberScript.async = true;
+
+			amberScript.addEventListener('load', onScriptLoad);
+			amberScript.addEventListener('error', onScriptError);
+
+			document.body.appendChild(amberScript);
+		}
```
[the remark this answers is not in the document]
AuresaNyctea: @bryanrsmith it looks like this has been solved above. Is that the case?

ddunkin:
I'm going to push up some commits sortly.

PR #157, opened 2019-05-01
Accordion component

RobertBolender:
Ready for review.

Notes to myself / TODO list:

- [x] Basic keyboard navigation
- [x] Basic collapsing behavior
- [x] Reduced padding on mobile, according to the spec.
- [x] Arrow key navigation should only be available when a header is selected. All form elements within the body should be in the normal tab order, and while focus is within the body there is no special focus handling.
[§ Keyboard Interaction](https://www.w3.org/TR/wai-aria-practices-1.1/#accordion)

- [x] Headers should receive focus when clicked.

- [x] Visually persistent elements that appear *adjacent* to the header should not be descendants within the DOM. My current approach is to use CSS Grid to position the element visually, but if the element is interactive it will still be within the document's tab order.
I think that ARIA might be happier if the header element was focusable and the adjacent indicator element was not focusable. We could still have visual styling that makes it *look* like the indicator was the focused element.
[§ WAI-ARIA Roles, States, and Properties](https://www.w3.org/TR/wai-aria-practices-1.1/#accordion)

- [x] I've decided against using a `PropTypes.node` prop for `customIndicator`. For now I'm also not going to add a "left" or "right" `indicatorPosition` prop.

- [x] I'm going to use a render prop for the `customIndicator`, passing it an `isExpanded` prop.

- [x] ~Add the appropriate `aria-level` attribute.~ Handled by HTML heading elements.
- [x] Added a `PropTypes.node` prop for `subtitle`.
- [x] Added the appropriate `aria-controls` attribute.


RobertBolender:
I want to run this by design real quick to make sure I've covered everything in [the spec](https://app.zeplin.io/project/58a23c2d4597afb248c13a43/screen/5b9bfe0f73b5c068de1b3eff), but the code is ready for review.

components/accordion/accordion-header.jsx
```diff
@@ -0,0 +1,111 @@
...
+	const buttonRef = useRef();
+	const isSelected = focusedMenuItem && focusedMenuItem === headerId;
+
+	useEffect(
+		() => {
+			if (isSelected && buttonRef.current) {
+				buttonRef.current.focus();
+			}
+		},
+		[isSelected, buttonRef],
```
bryanrsmith: You probably want `buttonRef.current`. `useRef` always returns the same object, it just has a mutable `current` property.

components/accordion/accordion-header.jsx
```diff
@@ -0,0 +1,111 @@
...
+		() => {
+			if (headerId) {
+				focusableChildList.current.push(headerId);
+			}
+		},
+		[headerId, focusableChildList],
+	);
+	return (
+		<Styled.HeadingWrapper customIndicator={customIndicator}>
+			<Styled.Heading as={`h${headingLevel}`}>
```
bryanrsmith: A `<section>` with `<header>` would be much easier to get semantically correct. h{1-6} assume too much context that I wouldn't expect a consumer of a component like this to have.

components/accordion/accordion-header.jsx
```diff
@@ -0,0 +1,111 @@
...
+					isExpanded={isExpanded}
+					onBlur={handleBlur}
+					onClick={handleExpansion}
+					onFocus={handleFocus}
+					ref={buttonRef}
+					panelId={panelId}
+					headerId={headerId}
+				>
+					<Styled.ButtonContentWrapper hideArrows={hideArrows} subtitle={subtitle}>
+						<React.Fragment>
```
bryanrsmith: `<>`?

components/accordion/accordion-header.jsx
```diff
@@ -0,0 +1,111 @@
...
+				</AccordionIndicator>
+			) : null}
+		</Styled.HeadingWrapper>
+	);
+}
+
+AccordionHeader.propTypes = {
+	children: PropTypes.node,
+	/** A render prop which receives an isExpanded boolean value. */
+	customIndicator: PropTypes.func,
```
bryanrsmith: Consider using `render` in the name. I'd expect this to accept a React node with its current name (which we may also want to support).

components/accordion/accordion-header.jsx
```diff
@@ -0,0 +1,111 @@
...
+	/** Which HTML heading element to use. */
+	headingLevel: PropTypes.number,
+	/** In most cases the subtitle should be hidden on mobile viewports, but that is a responsibility of the consumer. */
```
bryanrsmith: Why mention it at all? This isn't documenting the prop.

components/accordion/accordion-item.jsx
```diff
@@ -0,0 +1,44 @@
...
+
+	const context = useMemo(
+		() => ({
+			headerId,
+			panelId,
+			index,
+			isExpanded,
+			onExpansion: handleExpansion,
+		}),
+		[headerId, panelId, index, expandedSections, handleExpansion],
```
bryanrsmith: `isExpanded`?

components/accordion/accordion-item.jsx
```diff
@@ -0,0 +1,44 @@
...
+
+	return (
+		<Styled.AccordionItem>
+			<AccordionItemContextProvider value={context}>{children}</AccordionItemContextProvider>
+		</Styled.AccordionItem>
+	);
+}
+
+AccordionItem.propTypes = {
+	children: PropTypes.node.isRequired,
```
bryanrsmith: It'd be nice to document how `children` is used in each of these components. e.g., `/** The accordion section contents **/`

components/accordion/accordion-util.js
```diff
@@ -0,0 +1,74 @@
...
+
+export function useKeyboardNav(selectedIndex, setSelectedIndex, focusableItemIndexes) {
+	const handleKeyboardNav = useCallback(
+		event => {
+			const currentEnabledIndex = focusableItemIndexes.indexOf(selectedIndex);
+
+			// According to the ARIA Accordion spec, keyboard navigation should only be available
+			// when a header is focused, not when the focus is within a panel as panels may contain
+			// other form fields.
+			if (currentEnabledIndex === -1) {
```
bryanrsmith: 👍 Nice detail & comment.

components/main.js
```diff
@@ -3,6 +3,13 @@
 // We could use tree-shaking in the future to prune out unused exports,
 // but not all projects are able to use that right now.
 
+export {
+	Accordion,
+	AccordionHeader,
+	AccordionIndicator,
+	AccordionItem,
+	AccordionPanel,
+} from './accordion';
```
bryanrsmith: In cases where tightly coupled components are always used together, I often attach the helpers to the parent component as static properties, to simplify imports. What do you think? (No objection to the current approach; just offering an alternative)

```js
Accordion.Header = AccordionHeader;
Accordion.Indicator = AccordionIndicator;
Accordion.Item = AccordionItem;
Accordion.Panel = AccordionPanel;

export { Accordion };
```

```jsx
<Accordion>
	<Accordion.Item>
			<Accordion.Header subtitle="Subtitle for Section One.">
				Section One Title
			</Accordion.Header>
	</Accordion.Item>
	...
</Accordion>
```

catalog/accordion/variations.md
```diff
@@ -0,0 +1,90 @@
+These `<Accordion>` components follow the [WAI-AIRA spec for accordions](https://www.w3.org/TR/wai-aria-practices-1.1/#accordion).
+
+We do not currently support any options for requiring that one section always be open, or else limiting that no more than one section can be open. If you need this functionality, consider opening a PR.
```
bryanrsmith: Seems like a consumer could implement this themselves pretty easily, since they own the expansion state. We could make it even easier by specifying which section(s) expanded and which collapsed, in addition to the new set. That's easy to add later if we need it.

components/accordion/accordion-header.jsx
RobertBolender: We may need to update Stylelint, Prettier, or something to make it happy with that syntax.
I know that we've started using `<>` in FLCOM, but Styled-UI wasn't happy with it. I didn't spend any time looking into how to get that configured.

components/accordion/accordion-header.jsx
RobertBolender: I believe that this should be mentioned *somewhere*, but maybe it should be in the examples documentation rather than in prop types.

components/main.js
RobertBolender: I do like the look of that 👍

components/accordion/accordion-header.jsx
RobertBolender: I'll rename it to `renderCustomIndicator`.
I know that in some situations we prefer to avoid using render props, but I haven't seen any written guidelines of when render props are appropriate or any guidance around keeping our use of render props consistent.

components/accordion/accordion-header.jsx
RobertBolender: The ARIA spec says that the accordion `role=heading` component must either have `aria-level` specified or be a native element that has an implicit `aria-level` attribute, such as HTML heading elements.
[§ WAI-ARIA Roles, States, and Properties](https://www.w3.org/TR/wai-aria-practices-1.1/#accordion)

Would you prefer to make this a div with an `aria-level` attribute, rather than an HTML heading element?

bryanrsmith: Hmm, I don't know. I just have no idea how I would choose a reasonable value for it as a consumer of this component. I'd expect this to live in a sidebar or modal wizard of some sort, where the "document structure" of the greater page isn't really known or relevant to the component. I guess it seems to me like a wrong value is worse than no value.

components/accordion/accordion-header.jsx
bryanrsmith: Do you mean that we encourage custom hooks over render props now? Until recently, render props were a popular mechanism for reusable bits of behavior that didn't necessarily have UI attached to it. Hooks solve that problem much better, and we prefer to use them now. Render props are still a valid way to delegate rendering of a subsection of UI to the consumer, and (IMO) are often the cleanest way to do it. I think this is a good use. I prefer to name them with the "render" prefix, so it's clear to a consumer that they need to pass a function and not a component.

components/accordion/accordion-header.jsx
RobertBolender: > I guess it seems to me like a wrong value is worse than no value.

This is consistent with ARIA guidelines. ["No ARIA is better than bad ARIA"](https://www.w3.org/TR/wai-aria-practices/#no_aria_better_bad_aria).

The section on `aria-level` in particular suggests that assistive technology can *generally* calculate the level of an item from the DOM structure, and this attribute is primarily needed if the DOM does *not* match the information hierarchy.

I suggest that we use `<section>` and `<header>` instead of HTML heading elements, and support the `aria-level` prop but *not* set a default value. Consumers that feel that they need the `aria-level` can set it, and most of the rest of us will not use the prop.

PR #160, opened 2019-05-07
Apply classname to button, not content.

korbinancell:


components/button/anchor-button.jsx
```diff
@@ -19,4 +19,6 @@ AnchorButton.propTypes = {
 	...BaseButton.propTypes,
 	/** The destination the AnchorButton should link to. */
 	href: PropTypes.string,
+	/** See the docs for how to override styles properly */
```
bryanrsmith: Won't this comment appear *in* the docs? I think accurately describing the prop would be better than telling the reader to go away.

korbinancell: I think that's a direct copy-paste from react-strap and we don't even have any docs on it

korbinancell:
Looks like the button documentation is having issues as well. I'm looking into it.

korbinancell:
The current version of babel-react-docgen cannot handle React.forwardRef and pretty much anything that is not a class component. Unfortunately, the next version requires Babel v7 which is a task in itself. https://github.com/Faithlife/styled-ui/pull/135 includes an upgrade to v7 so I will look into it when I get the time.

korbinancell:
@bryanrsmith this is ready for review/merge

PR #161, opened 2019-05-13
Make popover focusaway handler optional and inline-block.

korbinancell:


bryanrsmith:
block -> inline-block is a breaking change, right?

korbinancell:
It will actually revert a breaking change. Popovers originally had inline-block but I messed that up when I added this extra div.

PR #163, opened 2019-05-16
Add visual regression testing

korbinancell:
This pr uses jest-image-snapshot to test for visual regressions in most of our components.

Not all components are covered as some it would be impossible currently (calendar input etc.) and some would require some additional navigation work.

I'm not sure if netlify will like puppeteer (runs headless chromium) or not, so we'll see.


korbinancell:
For some reason puppeteer is running a different size chromium whether it's in local or the build server. I am looking into it, but it's going slow.

PR #164, opened 2019-05-18
Add click handler to hide tooltip.

sportnak:
The tooltip doesn't *always* vanish when the mouse leaves the element. I thought it might be because of the debounce func, but rather than that I figured it would be more effective to just hide the tooltip on document click.

### DIFF
--- a/components/popover/tooltip.jsx
+++ b/components/popover/tooltip.jsx
@@ -17,7 +17,12 @@ export class Tooltip extends Component {
 		tooltipIsOpen: false,
 	};
 
+	componentDidMount() {
+		window.addEventListener('click', this.handleClick);
+	}
+
 	componentWillUnmount() {
+		window.removeEventListener('click', this.handleClick);
 		this.handleMouseLeave.cancel();
 	}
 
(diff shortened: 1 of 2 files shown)


bryanrsmith:
Thanks, but if there's a timer being mismanaged I'd prefer to solve the underlying problem. Are you able to consistently reproduce the problem? If so, can you post the steps to reproduce?

sportnak:
I do think a document click to hide would be helpful here. I don't have an easy way to repro it other than hovering over the tooltip quickly several times until it doesn't go away. You can see it if you do it a few times on the chms tool

bryanrsmith:
I'm not seeing it. Which part of chms tool?

sportnak:
The inferred base. Open beta. Create a new record with a gendered name. There's a lightbulb tooltip where it happens relatively often.

korbinancell:
It is possible that it is the `onMouseLeave` event. It is known to be flaky, especially on chrome, under the right (wrong?) circumstances.

bryanrsmith:
Closing in favor of #165

PR #173, opened 2019-06-10
Popover refactor

smaxa20:
* Add an unstyled option for popovers.
* Rebuild popover to use unstyled popovers.


components/popover/styled.jsx
```diff
@@ -18,23 +19,57 @@ export const getPlacement = placement => {
...
+		styleOverrides.fontSize ? styleOverrides.fontSize : 'medium'};
+	font-weight: ${({ styleOverrides }) =>
+		styleOverrides.fontWeight ? styleOverrides.fontWeight : 'normal'};
+	height: ${({ styleOverrides }) => (styleOverrides.height ? styleOverrides.height : 'auto')};
+	line-height: ${({ styleOverrides }) =>
+		styleOverrides.lineHeight ? styleOverrides.lineHeight : 'normal'};
+	margin: ${({ styleOverrides }) => (styleOverrides.margin ? styleOverrides.margin : '0px')};
+	max-height: ${({ styleOverrides }) =>
+		styleOverrides.maxHeight
+			? isNaN(styleOverrides.maxHeight)
```
bryanrsmith: `NaN` is falsy, so we would never enter this branch of the ternary. What's the reason for checking `NaN`? I wouldn't expect to need to handle that.

components/popover/unstyled-popover.jsx
```diff
@@ -0,0 +1,195 @@
...
+import { ThemeProvider } from 'styled-components';
+import { Popper } from 'react-popper';
+import { colors } from '../shared-styles';
+import { PlacementOptionsProps, FocusHandlerInboundsElement } from './popper-helpers';
+import * as Styled from './styled';
+
+/** Positioning helper used to display content above another element.
+ *  Refs are not supported, please use PopoverManager and PopoverReference to handle positioning.
+ */
+export class UnstyledPopover extends Component {
```
bryanrsmith: The name of this is a little awkward, and a little misleading since it's not entirely unstyled. A suffix instead of a prefix would also make it more recognizable as being closely related to `Popover`. What about something like... `PopoverBase`?

components/popover/component.jsx
```diff
@@ -1,184 +1,27 @@
...
+import { UnstyledPopover } from './unstyled-popover';
+
+/** Pre-styled popover that uses unstyled popovers internally. */
+export function Popover(props) {
+	const { children, placement, theme, styleOverrides, ...otherProps } = props;
+	const newStyle = { ...Popover.defaultProps.styleOverrides, ...styleOverrides };
+	return (
+		<UnstyledPopover
+			{...otherProps}
+			placement={placement || 'top'}
```
bryanrsmith: It looks like `UnstyledPopover` already provides a default for this, so you can probably remove it and let that component fill in the default.

components/popover/styled.jsx
smaxa20: What I was trying to do here was accept both ints and strings (ex. `12` and `'12px'`) for `max-height` and other similar props. We saw that some props in the old `Popover` required ints and others required strings so I was trying to move toward strings being uniform but also allow for previous consumers to keep what they already have. With `NaN` I was trying to see if what was input was a string in which case I wouldn't have to add the px label onto the end.

components/popover/unstyled-popover.jsx
smaxa20: Ok I can change this.

components/popover/component.jsx
smaxa20: Perfect.

components/popover/styled.jsx
smaxa20: It seems to work for both cases.

bryanrsmith: I recommend `typeof` for checking the type of something. Check out https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN (note the comment at the top indicating that `isNaN` has some pretty confusing behavior). Generally I'd only recommend using `isNaN`/`Number.isNaN` for checking if a value is the global [`NaN` value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN), which is rarely necessary.

smaxa20: Ok I see what you're saying. `typeof` is easier to read too (for me anyway).

PR #175, opened 2019-06-11
Display deprecation warnings to the consumer.

smaxa20:
The only deprecated element I could find was text-input v1 which was built using Typeahead and AsyncTypeahead, so I included those two elements. If I'm mistaken on how text-input v1 works or if I missed some deprecated elements let me know.

### DIFF
--- a/components/text-input/typeahead.jsx
+++ b/components/text-input/typeahead.jsx
@@ -73,65 +73,39 @@ const RelativeContainer = styled.div`
 	position: relative;
 `;
 
-export class AsyncTypeahead extends React.Component {
-	static propTypes = { inferred: PropTypes.bool };
+export const Typeahead = ({ inferred, ...props }) => (
+	<BootstrapContainer>
+		<RelativeContainer>
+			<StyledTypeahead {...props} />
+			{!inferred && (
+				<IndicatorContainer>
+					<SolidTriangleIcon />
+				</IndicatorContainer>
+			)}
+		</RelativeContainer>
+	</BootstrapContainer>
+);
 
-	componentDidMount() {
-		if (process.env.NODE_ENV !== 'production') {
-			console.warn(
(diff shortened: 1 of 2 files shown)


components/text-input/typeahead.jsx
```diff
@@ -73,39 +73,65 @@ const RelativeContainer = styled.div`
...
-);
-
-Typeahead.propTypes = {
-	inferred: PropTypes.bool,
-};
-
-AsyncTypeahead.propTypes = {
-	inferred: PropTypes.bool,
-};
+export class AsyncTypeahead extends React.Component {
```
bryanrsmith: Instead of converting to the older class API to gain access to the lifecycle method, use a `useEffect` callback with an empty dependency array.

https://reactjs.org/docs/hooks-effect.html

Something like:

```js
if (process.env.NODE_ENV !== 'production') {
	useEffect(() => {
		console.warn()
	}, []);
}
```

smaxa20: I tried this and it kept saying that useEffect was undefined. I now realize that I never included it from react, \*facepalm\*

smaxa20:
While reviewing what Bryan suggested, I found that the `Typeahead` and `AsyncTypeahead` elements were built off of Bootstrap. This means that the warnings I threw from the bootstrap elements are also being thrown for these elements. In my last commit [8fe6111](https://github.com/Faithlife/styled-ui/pull/175/commits/8fe61116792f76a42fc97b03067b9ac2b350c9b9) I undid my changes from the first commit in this PR [2d7b6e5](https://github.com/Faithlife/styled-ui/pull/175/commits/2d7b6e51c4931f0ebd1812bcb3f6e68cf9058a50). This removes the duplicate warnings that were being thrown. Now the only warnings that are being thrown are from the Bootstrap elements.

PR #176, opened 2019-06-18
Mobile compatibility fixes

smaxa20:


components/simple-toast/component.jsx
```diff
@@ -2,17 +2,21 @@ import React, { PureComponent } from 'react';
...
 /**
  * A mobile first Toast. A toast indicates that an action is being taken that does not require the user's
  * specific attention.
  */
 export class SimpleToast extends PureComponent {
 	static propTypes = {
+		/** Show time on desktop in milliseconds */
+		desktopTime: PropTypes.number,
+		/** Show time on mobile in milliseconds */
+		mobileTime: PropTypes.number,
```
bryanrsmith: Did we talk about unifying these and having a single duration for both mobile and desktop? I don't really get why we need two (especially with the addition of the dismiss button).

components/simple-toast/styled.jsx
```diff
@@ -115,3 +114,8 @@ export const ToastContainer = styled.div`
 export const ToastContent = styled.div`
 	white-space: nowrap;
 `;
+
+export const ToastClose = styled.div`
```
bryanrsmith: We should use a button for this. Generally, interaction handlers should only be attached to interactive/form elements like buttons.

components/share-dialog/styled.jsx
```diff
@@ -17,6 +22,10 @@ export const ShareContainer = styled.div`
 
 export const CopyContainer = styled.div`
 	margin: 0 ${thickness.eight};
+
+	input {
+		text-overflow: ellipsis;
```
bryanrsmith: Interesting. Does this work? Usually inputs scroll their contents with caret position.

components/simple-toast/component.jsx
smaxa20: We hadn't talked about it, but I could definitely unify them if that's what we want.

components/simple-toast/styled.jsx
smaxa20: Ok. I copied this from how it is in modals, should I change it there too?

catalog/drop-zone/variations.md
```diff
@@ -1,5 +1,7 @@
 ## Component Variations
 
+In general, it is recommended to provide some other alternate UI on mobile .
```
bryanrsmith: 👍

PR #179, opened 2019-06-19
Disable Popover GPU acceleration on Windows.

RobertBolender:
See also:

1. Earlier PR #28, 
2. Associated Bootstrap Bug twbs/bootstrap#22610.

### DIFF
--- a/components/popover/popover-base.jsx
+++ b/components/popover/popover-base.jsx
@@ -149,7 +149,6 @@ export class PopoverBase extends Component {
 		const {
 			children,
 			placement: popoverPlacement,
-			modifiers,
 			hideArrow,
 			delay,
 			theme,
--- a/components/popover/popover-base.jsx
+++ b/components/popover/popover-base.jsx
@@ -80,6 +80,7 @@ export class PopoverBase extends Component {
 
 	state = {
 		showPopper: this.props.isOpen || false,
+		popperModifiers: this.props.modifiers,
 	};
 
 	_timeout = null;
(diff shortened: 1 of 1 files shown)


components/popover/popover-base.jsx
```diff
@@ -143,10 +144,22 @@ export class PopoverBase extends Component {
 		} = this.props;
 		const { showPopper } = this.state;
 
+		const popperModifiers = {
+			...modifiers,
+			computeStyle: {
+				...modifiers.computeStyle,
+				gpuAcceleration: !(
+					typeof window !== `undefined` &&
```
bryanrsmith: Please move move `window` access to a lifecycle method.

RobertBolender: Is that still necessary, even though I'm checking whether the window is undefined?

components/popover/popover-base.jsx
```diff
@@ -74,11 +74,13 @@ export class PopoverBase extends Component {
 		theme: {
 			backgroundColor: colors.white,
 		},
+		modifiers: {},
 		styleOverrides: {},
 	};
 
 	state = {
 		showPopper: this.props.isOpen || false,
+		popperModifiers: this.props.modifiers,
```
RobertBolender: I realize that copying props into state is almost never a good idea.
I don't expect consumers to change this `modifiers` prop, are we comfortable with this tradeoff, and does it need to be explicitly called out in a comment or something?

smaxa20: Why is it not expected that consumers will change the `modifiers` prop? And if it's not expected to be used why is it included as a prop?

RobertBolender: I should clarify: I don't expect that a consumer would change the prop between renders, so we would be OK with ignoring updates to the prop and only use the initial value that is passed in.

smaxa20: Oh ok that makes more sense. Thank you.

bryanrsmith: We could avoid the problem by only putting a `shouldDisableGPUAcceleration` flag in state, and merging the necessary modifiers with the modifiers specified by the consumer at render time.

...Is it worth any of this complication? Should we disable it unconditionally?

RobertBolender: I'm not opposed to disabling GPU acceleration unconditionally.
Without objection, doing that.

PR #181, opened 2019-06-21
Use "body" as the default container for popovers/modals

kbsletten:
Previously, the default was inline which caused clipping issues when used in containers with overflow:hidden. Changing the default to 'body' allows most modals/popovers to work as expected with the option of containing the popup with the container prop.

### DIFF
--- a/catalog/popover/variations.md
+++ b/catalog/popover/variations.md
@@ -110,13 +110,13 @@ state: { isOpen: false }
 		<PopoverReference>
 			<Button primary medium onClick={() => setState({ isOpen: !state.isOpen })}>Show a Popover!</Button>
 		</PopoverReference>
-		<Popover isOpen={state.isOpen} placement="top">I'm inline</Popover>
+		<Popover isOpen={state.isOpen} placement="top" container="inline">I'm inline</Popover>
 	</PopoverManager>
 	<PopoverManager>
 		<PopoverReference>
 			<Button primary medium onClick={() => setState({ isOpen: !state.isOpen })}>Show a Popover!</Button>
 		</PopoverReference>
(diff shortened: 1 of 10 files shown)


components/modal/modal.jsx
```diff
@@ -73,15 +73,14 @@ export class Modal extends React.Component {
...
-		if (container) {
-			if (typeof container === 'string') {
-				// must be an id or body
-				this.targetContainer =
-					container === 'body' ? document.body : document.getElementById(container);
-			} else {
-				// must be a ref
-				this.targetContainer = typeof container === 'object' ? container.current : container();
-			}
+		if (container === undefined || container === 'body') {
```
bryanrsmith: This probably needs to accept `null` as well to avoid a crash on line 83. So probably `!container || container === 'body'`

components/modal/modal.jsx
```diff
@@ -73,15 +73,14 @@ export class Modal extends React.Component {
...
+			this.targetContainer = document.body;
+		} else if (typeof container === 'string') {
+			// must be an id or body
```
bryanrsmith: "or body" doesn't apply to this branch.

catalog/popover/variations.md
```diff
@@ -110,13 +110,13 @@ state: { isOpen: false }
 		<PopoverReference>
 			<Button primary medium onClick={() => setState({ isOpen: !state.isOpen })}>Show a Popover!</Button>
 		</PopoverReference>
-		<Popover isOpen={state.isOpen} placement="top">I'm inline</Popover>
+		<Popover isOpen={state.isOpen} placement="top" container="inline">I'm inline</Popover>
```
bryanrsmith: Does the `inline` container actually exist?

PR #182, opened 2019-06-23
Helpbox refactor

smaxa20:
Here is the [HelpBox spec](https://app.zeplin.io/project/58a23c2d4597afb248c13a43/screen/5d0d6b8991c0d270a9b83dad)


smaxa20:
I have placeholder icons right now, I still need to find the correct ones.

RobertBolender:
The CTA buttons can use the `small` 32px button size, per Amber's comment on the spec:
https://app.zeplin.io/project/58a23c2d4597afb248c13a43/screen/5d0d6b8991c0d270a9b83dad

bryanrsmith:
👍 Looks good! Have we verified that this is a non-breaking change for consumers still using the previous API?

smaxa20:
@bryanrsmith as far as I know it shouldn't be breaking. I didn't take out any props or change how any of them work, I just added. Some things will look differently than they did with the previous API such as light bulbs: previously the default light bulb icon was 40x40px IIRC and now it's 24x24px with the `large` prop making it 42x42px. I think that might be it though and those design changes are desired.

Using the old demo's first example as a reference, if a consumer wrote: `<HelpBox showLightBulb handleClose={() => true}>This is an alert with a light bulb.</HelpBox>` it displays what you would expect, just with a smaller light bulb.

Issue #142, opened 2019-03-05
All `classname` props on non-bootstrap components should be depricated in favor of the `theme` and `styleOverride` pattern.


korbinancell:
There are a couple older components that still accept classnames as a props. We should move them to use the new theme/styleOverride pattern and warn when they are used. Looks like Button, Checkbox, HelpBox, LoadingSpinner, and Radio are all offenders.

Moving away from this would get us one step closer to our global theming api goal.

korbinancell:
Closed in favor of v6's solution.

PR #185, opened 2019-06-25
Button refactor

smaxa20:


smaxa20:
Using the `React.forwardRef` function to forward the refs to `AnchorButton` and `Button` breaks the documentation for those components because our babel isn't up to date enough to support that function. Updating babel will come in a different PR by Bryan, but until then **don't merge this.**

components/button/anchor-button.jsx
```diff
@@ -1,31 +1,36 @@
...
-			);
-		}
-	},
-);
+AnchorButton.propTypes = {
+	...Button.propTypes,
+	/** The destination the AnchorButton should link to. */
+	href: PropTypes.string,
+	/** This should only be used as a last resort if Theme and StyleOverrides will not do what you need */
+	className: PropTypes.string,
```
bryanrsmith: Button should accept a className as well.

components/button/component.jsx
```diff
@@ -1,35 +1,77 @@
...
-);
+	return (
+		<MappedStyledComponent
+			ref={ref}
+			as={as}
+			theme={theme}
+			{...filteredProps || {}}
+			styleOverrides={styleOverrides}
+		>
+			<Styled.ButtonContentWrapper {...filteredProps || {}} styleOverrides={styleOverrides}>
```
bryanrsmith: What would `filteredProps` include here? I'm surprised to see any props being passed from the consumer to the wrapper.

components/button/styled.jsx
```diff
@@ -69,94 +63,120 @@ export const Button = styled.button`
 
 export const variationMap = {
 	primary: component => styled(component)`
-		border: 1px solid ${props => props.theme.defaultColor || buttonColors.default};
+		${ButtonContentWrapper} {
+			border: 1px solid ${props => props.theme.defaultColor || buttonColors.default};
```
bryanrsmith: The border should be on the button, not the wrapper.

components/button/anchor-button.jsx
smaxa20: [Are we not deprecating `className`?](https://github.com/Faithlife/styled-ui/issues/142)

components/button/anchor-button.jsx
```diff
@@ -1,31 +1,36 @@
...
 		};
+	}
+	return (
+		<Button
+			ref={ref}
+			disabled={disabled}
+			onClick={makeDisabled}
+			as="a"
+			{...baseProps}
+			{...anchorProps}
```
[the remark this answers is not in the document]
smaxa20: Yes we can.

components/button/component.jsx
smaxa20: `filteredProps` wasn't needed, I don't actually know why I included that. I must've missed it when I was taking things out to see if they were necessary.

As for the consumer passing any props to the wrapper, `styleOverrides` needs to be passed because some of the styles go on the button and some go on the wrapper.

components/button/styled.jsx
smaxa20: If the border is on the button there is a small area where the user can click and the focus will go onto the button and the focus styles with it. Nowhere else does that happen, and it doesn't happen like that in the live version. Everywhere the user clicks is actually on the wrapper.

bryanrsmith: I don't understand. The button _should_ have focus following a click.

components/button/component.jsx
bryanrsmith: > some of the styles go on the button and some go on the wrapper.

This severely complicates the task of customizing the appearance of a button from the consumer's standpoint. We should find a path forward that lets us stop doing this.

bryanrsmith: If we need a breaking change to get there, we can hold off on that part.

PR #186, opened 2019-06-26
Hide Tooltips on mobile

smaxa20:


components/popover/tooltip.jsx
```diff
@@ -1,57 +1,55 @@
...
+	const [isOnMobile, setIsOnMobile] = useState(false);
+
+	const handleMouseEnter = () => {
+		handleMouseLeave.cancel();
+		setToolTipIsOpen(true);
 	};
 
-	handleMouseLeave = debounce(() => {
-		this.setState({ tooltipIsOpen: false });
+	const handleMouseLeave = debounce(() => {
```
bryanrsmith: This will recreate the debounced function on every render, breaking the effect's ability to cancel it.

components/popover/tooltip.jsx
```diff
@@ -1,57 +1,55 @@
...
+		return () => {
+			handleMouseLeave.cancel();
+		};
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, []);
+
+	return (
+		<PopoverManager>
+			<PopoverReference
+				style={{ display: isOnMobile && 'none' }}
```
bryanrsmith: A cleaner strategy would be to avoid rendering any of the tooltip UI, and return `children` directly.

components/popover/tooltip.jsx
```diff
@@ -1,57 +1,55 @@
...
-					{children}
-				</PopoverReference>
-				<Popover {...otherProps} isOpen={tooltipIsOpen || isOpen}>
-					{content || text}
-				</Popover>
-			</PopoverManager>
-		);
-	}
+	useEffect(() => {
+		setIsOnMobile(window.matchMedia(`(max-width: ${mediaSizes.tablet})`).matches);
```
bryanrsmith: I think this is fine if we observe window resizes. As is, I'd expect this to fairly frequently catch small desktop browser windows, and fail to adapt when they're resized. An alternate approach might be to set this value in `onTouchStart`, so we only disable if they actually exercise a touch screen.

components/popover/tooltip.jsx
bryanrsmith: Also, it looks like this might display: none the children on mobile. This shouldn't affect the visible page at all; only disable the tooltip popup.

components/popover/tooltip.jsx
smaxa20: Actually, because the problem with `Tooltips` come when the user goes to tap it, should I use a `hover: hover` media query instead? Do we want them hidden on small screens or just on mobile/touch screens? 

I see what you're saying about it not being able to catch screen resizes, but if we always hide it on touch screens and always show it on non-touch screens then the resizing shouldn't matter.

components/popover/tooltip.jsx
smaxa20: I agree with your first point about not displaying any of the tooltip UI if we don't want to.

But if I display the children but disabled the tooltip popup the user will just see a button but when they try to interact with it it won't do anything and they're given no reason as to why. I think this should affect the visible page.

bryanrsmith: image

In this github UI, that would mean removing toolbar buttons on mobile. Anything can have a tooltip. If a consumer wants to put a tooltip on a non-interactive element, that's their business. We can't remove children from the tree.

components/popover/tooltip.jsx
bryanrsmith: Yes, a hover query sounds good.

components/popover/tooltip.jsx
smaxa20: Ok that makes sense.

smaxa20: I didn't think about having tooltips on elements that aren't just tooltips like the Github toolbar buttons. It was just weird on the demo.

PR #192, opened 2019-07-08
Add disabled prop to Checkbox and Radio

mdombrowski1997:


components/check-box/checkbox-content.jsx
```diff
@@ -2,10 +2,10 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import * as Styled from './styled';
 
-export function CheckboxContent({ isChecked, title, children }) {
+export function CheckboxContent({ isChecked, title, children, disabled }) {
```
korbinancell: Make sure and add this to PropTypes

components/check-box/component.jsx
```diff
@@ -46,6 +52,7 @@ export class Checkbox extends Component {
 					className={className}
 					role={'checkbox'}
 					aria-checked={isChecked}
+					disabled={disabled}
 				>
 					<CheckboxContent isChecked={isChecked} title={title}>
```
korbinancell: You expect the disabled prop in CheckboxContent but do not pass it along.

components/check-box/styled.jsx
```diff
@@ -8,6 +8,14 @@ export const CheckboxDiv = styled.div`
 	width: 14px;
 	height: 14px;
 	background: transparent;
+
+	${props =>
+		props.disabled
+			? `
+		border: solid 1px ${props.theme.disabledBorder};
```
korbinancell: This might be a prettier problem but if it will let you lets increase the indent here.

mdombrowski1997: Prettier won't allow different formatting, but I edited on GitHub to have it how I thought was better

components/check-box/component.jsx
```diff
@@ -60,6 +67,8 @@ Checkbox.defaultProps = {
 	theme: {
 		primary: '#1E91D6',
 		border: '#95908f',
+		disabledBackground: '#ebebeb',
+		disabledBorder: '#c7c7c7',
```
smaxa20: `primary`, `disabledBackground`, and `disabledBorder` should use their [shared-styles](https://github.com/smaxa20/styled-ui/blob/master/components/shared-styles/index.js#L50) names. They should be `blueBase`, `gray8`, and `gray22` respectively. I can't find one for `border`

components/radio/component.jsx
```diff
@@ -10,19 +10,27 @@ export class Radio extends Component {
...
 		disableAutoBlur: PropTypes.bool,
+		disabled: PropTypes.bool,
 	};
 
 	static defaultProps = {
 		theme: {
 			primary: '#1E91D6',
 			border: '#95908f',
+			disabledBackground: '#ebebeb',
+			disabledBorder: '#c7c7c7',
```
smaxa20: These should also use shared-styles.

bryanrsmith:
Thanks!

Issue #189, opened 2019-06-29
[RFD] Sequenced Tab Navigation


smaxa20:
#  Sequenced Tab Navigation

* Specifically describe the problem you'd like the new component to solve:

This will probably mostly be used in onboarding/account creation pages or around multi-page forms where there is a specific sequence that the user must follow.
* Are there existing solutions or examples where this component would help?

Giving has one implemented, and it's very difficult to get to. So here are the steps:
1. go to https://test.faithlife.com/examplechurch-finance-team/stewardship/#/overview if you're not a member of this group, ask Korbin to invite you.
2. Inspect the page and open up React tools.
3. Using the selector tool select the main panel's body
4. (Here's where it gets tricky) You need to find the <t> element with no props attached to it, and click on the <t> element right beneath it.
5. Change this component's state.merchant.status to "Preview" (note capital p)
    1. If you can't find this state you're looking in the wrong <t> element.
6. The page should have updated itself. Click "Activate your account" in the HelpBox at the top.
7. Click "Get started"
8. If you need help, ask me or Korbin and we hopefully will be able to help you.

Material-UI has a similar component implemented [here](https://material-ui.com/components/steppers/#steppers)

The way that Bryan and I think would be best going forward would be to build on top of Styled-UI's [Tabs](https://faithlife.github.io/styled-ui/#/tabs/variations).
* Describe the proposed API:

Bryan and I weren't sure which was desired, having the component just be the tabs or have the tabs and the content associated with each tab. We think that it would be most flexible for the consumer if the component was just the tabs that were controlled by state and have the content be the consumer's responsibility which could also be based on state.

Here is a sample of consuming code for just the tabs, it is similar to consuming code for `Tabs` but without the `TabPanels` component:
```
<TabManager selectedTab={state.currentTab} onSelectedTabChange={tabIndex => setState({ currentTab: tabIndex })}>
    <SequenceTabList>
        <SequenceTab>First Tab</SequenceTab>
        <SequenceTab>Second Tab</SeqeneceTab>
        <SequenceTab>Third Tab</SequenceTab>
    </SequenceTabList>
</TabManager>
```

Here is a sample of consuming code for the tabs and the content, it is very similar to consuming code for `Tabs`:
```
<TabManager selectedTab={state.currentTab} onSelectedTabChange={tabIndex => setState({ currentTab: tabIndex })}>
	<SequenceTabList>
		<SequenceTab>First Tab</SequenceTab>
[...]

PR #191, opened 2019-07-06
Fix misc errors and warnings.

timothy-b:
I got some nice warnings and errors while browsing the catalog. This fixes most of them.

Specifically, styled-components, i.e.:
```javascript
breadcrumbs.ts:155 Functions as object-form attrs({}) keys are now deprecated and will be removed in a future version of styled-components. Switch to the new attrs(props => ({})) syntax instead for easier and more powerful composition. The attrs key in question is "aria-level" on component "styled-header__Heading". 
 Error
    at http://localhost:4000/static/catalog.js:10556:331
    at StyledComponent.once [as warnAttrsFnObjectKeyDeprecated] (http://localhost:4000/static/catalog.js:10270:10)
    at http://localhost:4000/static/catalog.js:10669:22
    at Array.forEach (<anonymous>)
    at StyledComponent.buildExecutionContext (http://localhost:4000/static/catalog.js:10649:11)
    at StyledComponent.generateAndInjectStyles (http://localhost:4000/static/catalog.js:10702:65)
    at StyledComponent.renderInner (http://localhost:4000/static/catalog.js:10607:33)
    at updateContextConsumer (http://localhost:4000/static/catalog.js:148023:19)
    at beginWork (http://localhost:4000/static/catalog.js:148211:14)
    at performUnitOfWork (http://localhost:4000/static/catalog.js:151851:12)
    at workLoop (http://localhost:4000/static/catalog.js:151891:24)
    at renderRoot (http://localhost:4000/static/catalog.js:151974:7)
    at performWorkOnRoot (http://localhost:4000/static/catalog.js:152881:7)
    at performWork (http://localhost:4000/static/catalog.js:152793:7)
    at performSyncWork (http://localhost:4000/static/catalog.js:152767:3)
    at requestWork (http://localhost:4000/static/catalog.js:152636:5)
    at scheduleWork (http://localhost:4000/static/catalog.js:152450:5)
    at Object.enqueueSetState (http://localhost:4000/static/catalog.js:143884:5)
    at PageLoader.Component.setState (http://localhost:4000/static/catalog.js:131165:16)
    at http://localhost:4000/static/catalog.js:68991:14
```

And the listbox page:
```javascript
Warning: Each child in a list should have a unique "key" prop.

Check the render method of `ReactSpecimen`. See https://fb.me/react-warning-keys for more information.
    in ListItem (created by ReactSpecimen)
    in ReactSpecimen (created by SpecimenContainer)
    in div (created by Span)
[...]

PR #193, opened 2019-07-09
Sequenced Tab Navigation

smaxa20:
* [spec](https://app.zeplin.io/project/58a23c2d4597afb248c13a43/screen/5d12bc1140481c643ddff1b7)

* [RFD issue](https://github.com/Faithlife/styled-ui/issues/189)


components/tabs/tab-helpers.jsx
```diff
@@ -4,32 +4,56 @@ import { useId } from '../shared-hooks';
 import { useTabContext, useKeyboardNav } from './tab-utils';
 import * as Styled from './styled';
 
-export function TabList({ children }) {
+export function TabList({ children }, { isSequenced }) {
 	const { onSelectTab, selectedTabIndex, theme, panelIdsMap } = useTabContext();
 
 	const handleKeyboardNav = useKeyboardNav(selectedTabIndex, onSelectTab, children);
```
bryanrsmith: Would you touch base with @korbinancell on accessibility? I know he put some care into the keyboard navigation for tabs, and he might have opinions on how it can or should work with the sequenced variant.

components/tabs/tab-helpers.jsx
```diff
@@ -4,32 +4,56 @@ import { useId } from '../shared-hooks';
...
+		);
+	}
 }
 
 TabList.propTypes = {
 	children: PropTypes.node.isRequired,
 };
 
+export function SequencedTabList({ children }) {
+	return TabList({ children }, { isSequenced: true });
```
bryanrsmith: We should never manually invoke a component function. If you need to do this you can inject a prop.

```jsx
function SequencedTabList({ children }) {
	return <TabList isSequenced>{ children }</TabList>;
}
```

In this case, though, I think you should split the components apart.

components/tabs/styled.jsx
```diff
@@ -110,3 +110,78 @@ export const TabPanel = styled.div.attrs({
...
+	flex-direction: row;
+`;
+
+export const SequencedTab = styled.button.attrs(props => ({
+	role: 'tab',
+	'aria-selected': props.selected,
+	'aria-controls': `panel:${props.panelId}`,
+	'aria-disabled': props.disabled,
+	tabIndex: props.selected ? '0' : '-1',
+}))`
```
bryanrsmith: Consider extending the existing `Tab` component instead of duplicating it.

```js
const SequencedTab = styled(Tab)`
    background: ${({ selected }) => (selected ? colors.blueTint : 'white')};
`;

File: components/tabs/tab.jsx [hunk 4]
```diff
@@ -27,24 +29,48 @@ export function Tab(props) {
...
+					disabled={disabled}
+					theme={theme}
+					styleOverrides={styleOverrides}
+					selected={selected}
+				>
+					{typeof children === 'function' ? children({ selected, disabled }) : children}
+				</Styled.TabContent>
+			</Styled.Tab>
+		);
+	}
```
bryanrsmith: Here, too, we should either split the components, or find a way to avoid the duplication. You can probably remove the duplication by conditionally selecting the components to render, and unconditionally rendering the component structure. A similar alternative would be to let the consumer pass in the components as props. Then you can export a `SequencedTab` component that returns something like `<TabImpl {...props} TabComponent={styled.SequencedTab} />`, and a similar `Tab` component that injects `styled.Tab`.

```jsx
const Tab = isSequenced ? styled.SequencedTab : styled.Tab;

return <Tab disabled={disabled} {...etc}></Tab>;
```

components/tabs/styled.jsx
```diff
@@ -110,3 +110,78 @@ export const TabPanel = styled.div.attrs({
 
 	${({ selected }) => !selected && 'display: none'};
 `;
+
+export const SequencedTabList = styled.div.attrs(_ => ({ role: 'tablist' }))`
```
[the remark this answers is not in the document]
smaxa20: I was copying the style used [here](https://github.com/Faithlife/styled-ui/pull/191/files#diff-58b5f60f2f7f1e6c16ca9fb6a27a8aafR69) by Timothy.

components/tabs/styled.jsx
```diff
@@ -110,3 +110,78 @@ export const TabPanel = styled.div.attrs({
...
+	'aria-disabled': props.disabled,
+	tabIndex: props.selected ? '0' : '-1',
+}))`
+	display: flex;
+	flex-direction: row;
+	justify-content: center;
+	align-items: center;
+	flex-grow: 1;
+	flex-basis: 0;
+	cursor: ${({ disabled }) => disabled || 'pointer'};
```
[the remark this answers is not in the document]
smaxa20: If there's no cursor specified it shows the arrow, so using a boolean works as intended. The way you suggest works as well and might be clearer to read. I could change it if you want me to.

components/tabs/styled.jsx
```diff
@@ -110,3 +110,78 @@ export const TabPanel = styled.div.attrs({
...
+	display: flex;
+	flex-direction: row;
+`;
+
+export const SequencedTab = styled.button.attrs(props => ({
+	role: 'tab',
+	'aria-selected': props.selected,
+	'aria-controls': `panel:${props.panelId}`,
+	'aria-disabled': props.disabled,
+	tabIndex: props.selected ? '0' : '-1',
```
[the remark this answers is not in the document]
smaxa20: I copied that from [earlier in the file](https://github.com/Faithlife/styled-ui/blob/4272db40dad4de55b93f29caed963fa463a2bf97/components/tabs/styled.jsx#L47) but you're right. Should I change both occurrences?

components/tabs/styled.jsx
```diff
@@ -110,3 +110,78 @@ export const TabPanel = styled.div.attrs({
...
+	width: 24px;
+	min-width: 24px;
+	height: 24px;
+	margin: 0px 8px 0px 14px;
+	font-size: 14px;
+	font-weight: bold;
+	color: ${({ selected }) => (selected ? colors.blueBase : colors.gray52)};
+	background: ${({ completed }) => completed && colors.blueLight};
+
+	@media (max-width: ${mediaSizes.phone}) {
```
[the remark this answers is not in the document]
smaxa20: Ok, I did that a couple places. I'll flip them.

components/tabs/tab-helpers.jsx
smaxa20: Ok that's good to know. But yeah I'll separate them.

components/tabs/tab-helpers.jsx
smaxa20: Ok I'll talk to him, the problem I was facing here is that using the arrows allowed the user to skip steps if they went left from tab 1.

components/tabs/styled.jsx
smaxa20: I can do that, it allows me to take out these attrs and a few other attributes.

components/tabs/tab.jsx
smaxa20: I think I'll just separate them. They have different props and pretty different return statements. Plus it will be easier to read if they're separated.

I tried to be clever with this component and `SequencedTabList` by trying to incorporate them into the other components, but now I see that the way to do this that is the clearest and easiest to read is to separate them.

components/tabs/styled.jsx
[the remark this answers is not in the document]
bryanrsmith: I didn't realize styled-components had deprecated the object form; please disregard my suggestion to pass the object directly.

The underscore thing is a common pattern in C#, where it's common for a `Func<T>` argument to specify non-optional arguments. We should not use that pattern in JS. If you'd prefer to keep the named argument in place, it should have a meaningful name, like `props`.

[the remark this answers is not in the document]
smaxa20: Ok, I changed it to `() =>` and commented on Timothy's PR about it.

components/tabs/tab-utils.js
```diff
@@ -35,13 +35,66 @@ export function useKeyboardNav(selectedIndex, onSelectTab, children) {
 
 			switch (event.key) {
 				case handledKeys.arrowRight: {
-					const nextEnabledIndex = (currentEnabledIndex + 1) % enabledTabIndexes.length;
+					// const nextEnabledIndex = (currentEnabledIndex + 1) % enabledTabIndexes.length;
```
RobertBolender: Commented code should be removed here.

components/tabs/tab.jsx
```diff
@@ -61,3 +62,59 @@ Tab.propTypes = {
 Tab.defaultProps = {
 	styleOverrides: {},
 };
+
+export function SequencedTab(props) {
+	// PropType linting is diabled so out hidden props can be destuctured along with own consumer props
```
RobertBolender: Super nit: "our" hidden props instead of "out".

components/tabs/styled.jsx
```diff
@@ -72,10 +84,9 @@ export const TabContent = styled.span.attrs({ tabIndex: -1 })`
 	white-space: nowrap;
 	min-height: fit-content;
 	display: inline-block;
-	font-size: ${({ styleOverrides }) => styleOverrides.fontSize || thickness.sixteen};
-	width: ${({ styleOverrides }) => styleOverrides.width};
-	padding: ${({ styleOverrides }) =>
-		styleOverrides.padding || `${thickness.eight} ${thickness.sixteen}`};
+	font-size: ${styleOverrides => styleOverrides.fontSize || thickness.sixteen};
+	width: ${styleOverrides => styleOverrides.width};
+	padding: ${styleOverrides => styleOverrides.padding || `${thickness.eight} ${thickness.sixteen}`};
```
RobertBolender: This is surprising to me, do we not need to destructure `styleOverrides` from the `props` object anymore?

smaxa20: `${props => props.styleOverrides.width};` works.
`${({ styleOverrides }) => styleOverrides.width};` works. 
`${styleOverrides => styleOverrides.width};` doesn't work.
`${({ props}) => props.styleOverrides.width};` throws an error.

What I have there doesn't work and I shouldn't have pushed it. I think I had changed it because it was throwing an error for some reason and forgot to change it back. Thank you for catching that.

components/tabs/tab-helpers.jsx
```diff
@@ -30,6 +29,48 @@ TabList.propTypes = {
...
+export function SequencedTabList({ children }) {
+	const { onSelectTab, selectedTabIndex, theme, panelIdsMap } = useTabContext();
+
+	const [touchedTabs, setTouchedTabs] = useState(new Set());
+
+	useEffect(() => {
+		const newTouchedTabs = touchedTabs;
+		newTouchedTabs.add(selectedTabIndex);
+		setTouchedTabs(new Set(newTouchedTabs));
+		// eslint-disable-next-line react-hooks/exhaustive-deps
```
RobertBolender: Disabling this eslint rule feels like a bad idea to me.
I see that you don't want to re-run the `setTouchedTabs` call each time that `touchedTabs` is updated (which would be an infinite loop), but you could add a check to see whether the value had actually changed before calling that set function.

components/tabs/tab-helpers.jsx
```diff
@@ -32,17 +32,30 @@ TabList.propTypes = {
 export function SequencedTabList({ children }) {
 	const { onSelectTab, selectedTabIndex, theme, panelIdsMap } = useTabContext();
 
+	const [touchedTabs, setTouchedTabs] = useState(new Set());
+
+	useEffect(() => {
+		const newTouchedTabs = touchedTabs;
+		newTouchedTabs.add(selectedTabIndex);
```
bryanrsmith: This mutates the `Set` in state. You need to copy it. `Set`s are, unfortunately, not a particularly well suited for storing in state because they use mutable APIs that make updates unnecessarily awkward.

smaxa20: Does `const newTouchedTabs = touchedTabs;` not make a copy? Do I need to do `const newTouchedTabs = new Set(touchedTabs);`?

Are you saying that `Set`s shouldn't be used to store state? In this situation I think that using a `Set` is far better than any other option. If I used an `Array` I would either get a very large `Array` or I would have to loop through it every time to avoid duplicates.

bryanrsmith: > Does const newTouchedTabs = touchedTabs; not make a copy?

No. It creates a new reference to an existing object.

> Do I need to do const newTouchedTabs = new Set(touchedTabs);?

Yep.

> Are you saying that Sets shouldn't be used to store state?

No; I'm saying that React's state management model requires immutability to function properly. If you mutate an object in state React has no way of knowing that state changed, deps arrays won't function properly because the the previous state value and current state value are referentially equal, etc.. Mutating objects stored in React state is never safe. `Set` uses an mutable API-- adding or removing items mutates the `Set`, it doesn't produce a new one. That's awkward because you can't directly use the Set's API methods when the Set is stored in state; you have to remember to copy it every time.

> If I used an Array I would either get a very large Array or I would have to loop through it every time to avoid duplicates.

You can keep the `Set` if you prefer, I'm just noting that the mutable nature of them makes them fairly awkward in React components. People tend to forget to copy them. If performance is your concern, I wouldn't worry about it. Code readability should always come first. Things like looping over arrays rarely cause problems, and we can optimize them we they do.

smaxa20: > adding or removing items mutates the Set, it doesn't produce a new one.

This is very good to know. Does what I do now by creating a copy, adding to that copy, then setting the state with a copy of that copy avoid the problems with mutability?

I feel like using `Set` would be more readable than using `Array` and getting rid of all duplicates because a `Set` is already assumed to have no duplicates. Of course assuming that what I've done actually works as intended under the hood.

components/tabs/tab-helpers.jsx
smaxa20: Ok that's a good idea.

components/tabs/tab-helpers.jsx
bryanrsmith: > Does what I do now by creating a copy, adding to that copy, then setting the state with a copy of that copy avoid the problems with mutability?

Yep, that works.

> I feel like using Set would be more readable than using Array and getting rid of all duplicates because a Set is already assumed to have no duplicates.

I usually use arrays with spread syntax for adding items, `.filter()` for removing, and `.includes()` or `.find()` for testing. `[...new Set(items)]` can quickly remove duplicates if you need to do that explicitly. No objection to the `Set` if you're happy with it.

components/tabs/tab-helpers.jsx
```diff
@@ -30,6 +29,59 @@ TabList.propTypes = {
...
+	useEffect(() => {
+		const newTouchedTabs = new Set(touchedTabs);
+		newTouchedTabs.add(selectedTabIndex);
+
+		const setsAreEqual =
+			newTouchedTabs.size === touchedTabs.size &&
+			[...newTouchedTabs].every(value => touchedTabs.has(value));
+
+		if (!setsAreEqual) {
+			setTouchedTabs(new Set(newTouchedTabs));
```
bryanrsmith: Copying the `Set` again seems unnecessary. I wouldn't bother testing for equality either. Just unconditionally `setTouchedTabs(newTouchedTabs)`.

smaxa20: You're right, a bit of overkill. But without testing for equality this would become an infinite loop [like Robert said](https://github.com/Faithlife/styled-ui/pull/193#discussion_r302669432).

components/tabs/tab-helpers.jsx
smaxa20: Ok I see how that can work. I'm happy with `Set` though.

components/tabs/tab-helpers.jsx
bryanrsmith: Oh. Why do we need `touchedTabs` in the dependency list? Isn't this the only code that modifies them?

smaxa20: Yeah it's the only place they're being modified. But if I exclude it in the dependency list it throws a linter error which @RobertBolender said might be a bad idea (in the same thread that I linked above).

bryanrsmith: Got it. Sorry, I'm not sure I fully understood the intent before. Using the functional update form would let you omit it from the dep list. Something like:

```js
useEffect(() => {
    setTouchedTabs(touchedTabs => new Set(touchedTabs).add(selectedTabIndex)
}, [selectedTabIndex]);
```

smaxa20: Oh I didn't even think about doing it like that. That looks great!

bryanrsmith:
@smaxa20 Could you take a look at the conflict in package.json? This looks great to me & I'm happy to merge it as soon as there's a green button for me to click :-)

smaxa20:
@bryanrsmith the only difference is whether or not the caret prefix appears before the `focus-visible` version number. I think it makes sense to keep it updated on minor versions, but I defer that decision to you.

bryanrsmith:
Locking the version by omitting the caret is usually done deliberately to avoid an update that contains a bug or conflict in the dependency tree. I think you added that one, so if that's not the case here let's go ahead and change it to a caret range.

smaxa20:
Great. Ready to merge.

PR #194, opened 2019-07-10
Ag grid

korbinancell:
Here is a first pass at using ag-grid. A couple things to note:
* The goal of this draft is making sure that the current implementation is extensible and an adequate facade.
* After talking performance with some ChMS people it looks like they are completely handling sorting and filtering client side with excellent results so I have opted to start only supporting client side as most of the consumers I have talked to do not have a good search/ filtering set up.
* I'm not sure how to move forward with additional tables. I have a couple more complex versions that I will add that giving needs and seem to be very reusable, but I am not sure if we want to export the `BaseTable` component and allow consumers to use it, or make them add their tables to styled-ui.


components/grid/simple-table.jsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { useTableState } from './table-helpers';
+import { BaseTable } from './base-table';
+
+export function SimpleTable({
+	onRowClick,
+	children,
```
rmjohnson: What's the use case for children?

components/grid/base-table.jsx
```diff
@@ -0,0 +1,189 @@
...
+			if (initialPageNumber) {
+				api.goToPage(initialPageNumber);
+			}
+		},
+		[setGridApi, setColumnApi, sortModel, filterText, initialPageNumber],
+	);
+
+	const handleSelectionChanged = useCallback(() => {
+		const selectedRows = gridApi.getSelectedRows();
+		onRowClick && onRowClick(selectedRows[0].id);
```
rmjohnson: Is this an ag grid id or a passed in id from `data`?

korbinancell: Good catch, that is an id used in the row data. I meant to change it to pass the entire row object.

components/grid/simple-table.jsx
korbinancell: It is just the react children that are passed along to the `Base-Table` component.

PR #198, opened 2019-07-15
Modal fixes

smaxa20:
[Title color note plus full modal spec](https://app.zeplin.io/project/58a23c2d4597afb248c13a43/screen/5ad9263c2a990c7b7051a3a7?did=5d237095c9bfd83781a3fe91)

### DIFF
--- a/components/modal/default-modal-footer.jsx
+++ b/components/modal/default-modal-footer.jsx
@@ -15,6 +15,7 @@ export const DefaultModalFooter = props => (
 					theme={{ width: props.useFullWidthButtons ? '100%' : null }}
 					tabIndex={props.commitButton.tabindex}
 					onClick={props.commitButton.onClick}
+					disabled={props.commitButton.disabled}
 				>
 					{props.commitButton.text}
 				</Button>
@@ -27,6 +28,7 @@ export const DefaultModalFooter = props => (
 						theme={{ width: props.useFullWidthButtons ? '100%' : null }}
 						tabIndex={props.cancelButton.tabindex}
 						onClick={props.cancelButton.onClick}
+						disabled={props.cancelButton.disabled}
 					>
 						{props.cancelButton.text}
 					</Button>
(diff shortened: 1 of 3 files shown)


components/modal/default-modal-footer.jsx
```diff
@@ -15,6 +15,7 @@ export const DefaultModalFooter = props => (
 					theme={{ width: props.useFullWidthButtons ? '100%' : null }}
 					tabIndex={props.commitButton.tabindex}
 					onClick={props.commitButton.onClick}
+					disabled={props.commitButton.disabled}
```
korbinancell: Should these be added to the propTypes?

smaxa20: They're using the disabled prop from `Button` which is using the disabled prop from the html button so it doesn't need to be added to the propTypes. If you think it would be better to add them in for clarity's sake I could do that, but even `Button` doesn't have it in its propTypes.

korbinancell: Hmm. It should probably be in both places. Otherwise it's not super discover-able.

LawrenceRafferty:
Should this PR also have the font size/weight and line height fix? https://app.zeplin.io/project/5b6c978b008a822d0c73d94d/screen/5cdf066390276b2a9a530c0e?did=5d1f9aff2bdce756beb0d214&cmid=5d2376d6f444a0374b9b6b19

Here is where I'm working around it: https://git.faithlife.dev/Logos/CommunicationsApi/pull/126/files#diff-0ab77b4dab5455833ae8d47ede881f3fR11 (which causes a proptypes runtime warning: Warning: Failed prop type: Invalid prop `title` of type `object` supplied to `ModalHeader`, expected `string`.)

components/modal/default-modal-footer.jsx
smaxa20: I agree. I can add it here for these subtitle buttons, but it's not really in the scope of this PR to add them to `Button` I'll add it there with a different PR.

smaxa20:
@LawrenceRafferty Good catch, I thought I had written down all of the things that you told me about last week, but I missed that one. I'll add it right now.

bryanrsmith:
@smaxa20 Is this good to go? I can't tell if you've resolved the issue Lawrence mentioned.

smaxa20:
@bryanrsmith It's good to go.

PR #200, opened 2019-07-15
Add textarea variation to input

smaxa20:
[spec](https://app.zeplin.io/project/58a23c2d4597afb248c13a43/screen/59ee876e4582a457c39dbefe)

### DIFF
--- a/catalog/text-input/input.md
+++ b/catalog/text-input/input.md
@@ -30,3 +30,14 @@ state: { value: 'Washington' }
 	<Button large primary>Search</Button>
 </InputSpacer>
 ```
+
+### Inline
+
+```react
+state: { value: 'Washington' }
+---
+<InputSpacer>
+	<Input inline value={state.value} onChange={event => setState({ value: event.target.value })} />
+	<Button primary>Search</Button>
+</InputSpacer>
+```
--- a/catalog/text-input/input.md
+++ b/catalog/text-input/input.md
@@ -41,3 +41,14 @@ state: { value: 'Washington' }
 	<Button primary>Search</Button>
 </InputSpacer>
 ```
+
+### Textarea
+
+```react
+state: { value: '' }
+---
+<InputSpacer>
+	<Input textarea placeholder='My favorite state is...' value={state.value} onChange={event => setState({ value: event.target.value })} />
+	<Button small primary>Search</Button>
+</InputSpacer>
+```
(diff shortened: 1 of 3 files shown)


catalog/text-input/input.md
```diff
@@ -30,3 +30,25 @@ state: { value: 'Washington' }
 	<Button large primary>Search</Button>
 </InputSpacer>
 ```
+
+### Inline
```
smaxa20: This variation already existed but I couldn't find anywhere that it was demoed. If we don't want it to be demoed let me know.

components/input/styled.jsx
```diff
@@ -64,3 +64,32 @@ export const variationMap = {
 		}
 	`,
 };
+
+export const Textarea = styled.textarea`
```
smaxa20: I tried to find a way to include this in the `variationMap` but I couldn't figure out a way because it needs a different base html element. If there is a way to add it to the `variationMap` I would be able to remove my changes from component.jsx

korbinancell: Styled components allows for an 'as' prop that changes the base element of a styled component. https://www.styled-components.com/docs/api#as-polymorphic-prop

smaxa20: Ahh that's right, that's perfect for this situation. Thank you!

components/input/component.jsx
```diff
@@ -70,6 +73,7 @@ export const Input = forwardClassRef(
 
 			return (
 				<MappedStyledComponent
+					as={textarea && Styled.Textarea}
```
bryanrsmith: We want this to be styled just like inputs, but multiline, right?.  Can that not be accomplished with `as="textarea"`, with no other changes?

smaxa20: If I understand the spec correctly we want it to be larger than any of the other inputs to show the user that a paragraph is supposed to go there. By just using `as={textarea && 'textarea'}` the consumer has to specify both a size and `textarea` as props to prevent it from looking bad before any resizing. And even then I don't think we want a `large` variation textarea because the text would be too big to be in a paragraph.

bryanrsmith: > the consumer has to specify both a size and textarea as props to prevent it from looking bad before any resizing

There's no reason we can't change the default sizing based on the `as` prop.

> even then I don't think we want a large variation textarea because the text would be too big to be in a paragraph.

Why would we care if a consumer does that? It seems to me like we're adding a lot of code for virtually no benefit over forwarding the `as` prop.

smaxa20: Ok, so my problem was that once I use the `as` prop I wasn't able to add any styles to the textarea, but I was being too restrictive of the consumer so now the `height` and `width` styleOverrides work and now those and the base `Input` styles are the only things that is styling the textarea.

PR #202, opened 2019-07-17
Fix popover styling.

sportnak:
Currently causes the width of the popover to be broken for chms.


bryanrsmith:
Hah. Looks like a nice copy/paste error :-)

PR #207, opened 2019-07-22
Expose rowHeight property

ronnywoods:


components/grid/simple-table.jsx
```diff
@@ -13,6 +13,7 @@ export function SimpleTable({
 	maxRows,
 	rowSelectionType,
 	hideHeaders,
+	rowHeight,
```
korbinancell: It's unfortunate that we have to do this for every new property, but would you add this to `PaginatedTable` as well.

PR #208, opened 2019-07-23
Use layout components in Styled-UI components

smaxa20:


components/Box.js
```diff
@@ -3,19 +3,18 @@ import { layout, flexbox, position, grid, shadow, border, background } from 'sty
...
+
+	${common};
+	${layout};
+	${flexbox};
+	${position};
+	${grid};
+	${shadow};
+	${border};
+	${background};
+`;
```
bryanrsmith: What's the reason for this change?

components/help-box/component.jsx
```diff
@@ -14,55 +16,120 @@ export function HelpBox({
...
-			theme={theme}
-			success={helpBoxProps.success}
-			danger={helpBoxProps.danger}
-			warning={helpBoxProps.warning}
-			minor={helpBoxProps.minor}
 			stacked={stacked}
-			{...filteredProps}
+			backgroundColor={
+				success
+					? globalTheme.colors.green1
```
bryanrsmith: We'll want to use the key name instead of referencing the theme directly, so that it can be customized. `success ? 'green1' : ...`. styled-system's helpers will look it up in the theme available through context.

components/help-box/component.jsx
```diff
@@ -14,55 +16,120 @@ export function HelpBox({
...
+					? globalTheme.colors.green2
+					: danger
+					? globalTheme.colors.red3
+					: warning
+					? globalTheme.colors.yellow3
+					: minor
+					? globalTheme.colors.gray14
+					: globalTheme.colors.blue3
+			}`}
+			borderLeft={`solid 4px ${
```
bryanrsmith: Find a way to avoid duplicating the large nested ternary. An easy solution would be to override the border-left-width. It looks like that's not a prop provided by styled-system; we could add one, or use the `css` prop.

e.g.

```jsx
<Box border={1} borderColor={success ? 'green2' : danger ? 'red3' : warning ? 'yellow3' : minor ? 'gray14' : 'blue3'} css={`border-left-width: 4px`}>
```

components/help-box/component.jsx
```diff
@@ -14,55 +16,120 @@ export function HelpBox({
...
+					: warning
+					? globalTheme.colors.yellow3
+					: minor
+					? globalTheme.colors.gray14
+					: globalTheme.colors.blue3
+			}`}
+			color={globalTheme.colors.flGray}
+			position="relative"
+			display="flex"
+			borderRadius={globalTheme.radii[1]}
```
bryanrsmith: Pass the index of the value you want from the theme scale. `borderRadius={1}`

components/help-box/component.jsx
```diff
@@ -14,55 +16,120 @@ export function HelpBox({
...
-	const { component: HelpBoxVariation, filteredProps } = applyVariations(
-		Styled.HelpBox,
-		Styled.variationMap,
-		helpBoxProps,
-	);
-
 	return (
-		<HelpBoxVariation
+		<Box
 			className={className}
```
bryanrsmith: Looks like you're not directly using the className, so you could let it be spread with the other `helpBoxProps`.

components/help-box/component.jsx
```diff
@@ -14,55 +16,120 @@ export function HelpBox({
...
+						{danger ? <Exclamation /> : success ? <CircleCheck /> : minor ? null : <Info />}
+					</Styled.Icon>
 				))}
-			<Styled.HelpBoxContent>{children}</Styled.HelpBoxContent>
+			<Styled.HelpBoxContent
+				stacked={stacked}
+				display="flex"
+				flex="1"
+				height={large ? '230px' : ''}
+				padding={`14px ${globalTheme.space[5]}px 14px ${globalTheme.space[4]}px`}
```
bryanrsmith: Spacing values should always come from the theme now. Let's swap 14px for 12px. Avoid the string concatenation by using the various props provided by styled-system.

```jsx
padding={4}
paddingRight={5}
```

components/help-box/styled.jsx
```diff
@@ -1,140 +1,86 @@
...
+	flex: none;
+	width: ${props => (props.large ? '42px' : '24px')};
+	height: ${props => (props.large ? '42px' : '24px')};
+	margin: ${`${globalTheme.space[4]}px ${globalTheme.space[0]}px ${globalTheme.space[0]}px
+				${globalTheme.space[5]}px`};
+
+	path {
+		fill: ${props =>
+			props.success
+				? globalTheme.colors.green2
```
bryanrsmith: Here, too, these need to come from the context theme. `props.theme.colors.green2`. We probably need to add a theme provider component that injects the default theme into context when one isn't present for these more complex components.

components/help-box/styled.jsx
```diff
@@ -1,140 +1,86 @@
...
-		text-align: left;
-		line-height: 1.25;
-		font-size: 16px;
-		color: ${colors.flGray};
+		order: 2;
+	}
 
-		height: ${props => (props.large ? '230px' : '')};
-		padding: 14px 16px 14px 12px;
+	${HelpBoxFooter} {
```
bryanrsmith: Everything in this file seems like it could be more cleanly expressed with function components. You shouldn't need the styled component type selectors because you can pass the `stacked` prop directly.

components/help-box/styled.jsx
```diff
@@ -1,140 +1,86 @@
...
-		&::-moz-focus-inner {
-			border: 0;
-			padding: 0;
-		}
+`;
+
+export const CloseIcon = styled(Box)`
+	path {
+		fill: ${props =>
+			props.success
```
bryanrsmith: It would be nice to avoid duplicating this. If we can't find a way to reuse the calculation, is there a helper function that could make this more succinct?

bryanrsmith: Maybe something like...

```js
const variant = getVariant({ success, danger, warning, minor });
const fillColor = fillColors[variant];


function getVariant(obj) {
   for (const [key, value] of Object.entries(obj)) {
        if (value) {
            return key;
        }
    }

    return null;
}

File: components/Box.js (hunk 1 above)
Comment (smaxa20) [reply components/Box.js]: All the other layout components were using this syntax, it was just for consistency.

File: components/help-box/component.jsx (hunk 3 above)
Comment (smaxa20) [reply components/help-box/component.jsx]: That `css` prop is handy, I didn't know that was available.

File: components/help-box/component.jsx (hunk 2 above)
Comment (smaxa20) [reply components/help-box/component.jsx]: Can I do this in the `Styled` file as well? Like [here](https://github.com/Faithlife/styled-ui/pull/208/files?file-filters%5B%5D=.jsx#diff-1bfcc755fb3287dd4905b57785822ceaR48). It doesn't seem to let me.

File: components/help-box/component.jsx (hunk 6 above)
Comment (smaxa20) [reply components/help-box/component.jsx]: I have a need for negative margins in this component. What's the best way to handle those?

File: components/Box.js (hunk 1 above)
Comment (bryanrsmith) [reply components/Box.js]: Ah, whoops.

File: components/help-box/component.jsx (hunk 6 above)
Comment (bryanrsmith) [reply components/help-box/component.jsx]: Margin props should already accept negative numbers for this.

Comment (smaxa20) [reply components/help-box/component.jsx]: I can use negative numbers regularly like `marginLeft="-4px"`, but the theme doesn't have any negative spacing values.

File: components/help-box/styled.jsx (hunk 8 above)
Comment (smaxa20) [reply components/help-box/styled.jsx]: `HelpBoxBody` and `HelpBoxFooter` are components added by the consumer so they would have to pass in the `stacked` prop themselves. We originally set it up this way so the consumer only needs to pass props to `HelpBox` and then they would be passed down to the body and footer. I guess I just don't know how to pass them props if I'm not using them in the render function.

File: components/help-box/component.jsx (hunk 6 above)
Comment (bryanrsmith) [reply components/help-box/component.jsx]: If you pass `marginLeft={-4}`, it will look up the spacing value at index 4, and then negate it.

Comment (smaxa20) [reply components/help-box/component.jsx]: Oh wow that's really nice.

File: components/help-box/component.jsx [hunk 10]
```diff
@@ -1,8 +1,65 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Close, Exclamation, CircleCheck, Info } from '../icons';
-import { applyVariations } from '../utils';
-import * as Styled from './styled';
+import { theme as globalTheme } from '../../theme';
```
bryanrsmith: Theme should pretty much always come from context. Otherwise consumers will be unable to customize the theme.

For stuff like the the `getFill` use case, you could either use a styled system custom prop (which can look up theme data), or retrieve the theme from context with the `useTheme` hook.

https://styled-system.com/custom-props

Something like....
```jsx
import { system } from 'styled-system';

const IconBox = styled(Box)`
    path {
        ${system({ fill: { property: 'fill', scale: 'colors' } })}
    }
`;

<IconBox fill={variation.icon}>
```

Or...
```
const theme = useTheme();

File: components/help-box/component.jsx [hunk 11]
```diff
@@ -1,8 +1,65 @@
...
+function getVariation(variation, obj) {
+	if (variation) {
+		return variation;
+	}
+	return [...Object.entries(obj)].find(entry => entry[1])[0];
+}
+
+function getFill(theme, color) {
+	return `path {
+				fill: ${new Map(Object.entries(theme.colors)).get(color)};
```
bryanrsmith: The Map seems unnecessary. just `theme.colors[color]` should do the trick.

components/help-box/component.jsx
```diff
@@ -1,8 +1,65 @@
...
+		closeIconColor: 'gray34',
+		bulbColor: 'gray14',
+	},
+	original: {
+		bg: 'blue1',
+		borderColor: 'blue3',
+		iconColor: 'blue3',
+		icon: <Info />,
+		closeIconColor: 'blue5',
+		bulbColor: 'blue3',
```
bryanrsmith: No objection if this is a better approach, but all of the variants use the same color for the border, icon, and bulb. You could probably consolidate them into a single `fg` property.

components/help-box/component.jsx
```diff
@@ -14,55 +71,102 @@ export function HelpBox({
...
-		Styled.HelpBox,
-		Styled.variationMap,
-		helpBoxProps,
-	);
+	const chosenVariation =
+		variations[getVariation(variation, { success, danger, warning, minor, original: true })];
+	const childrenWithProps =
+		typeof children === 'string'
+			? children
+			: React.Children.map(children, child => React.cloneElement(child, { stacked: stacked }));
```
bryanrsmith: Use the shorthand `{ stacked }`

catalog/help-box/variations.md
```diff
@@ -6,19 +6,19 @@
 		<HelpBox.Body>This is a helpful alert.</HelpBox.Body>
 		<HelpBox.Footer><Button primary small>CTA</Button></HelpBox.Footer>
 	</HelpBox>
-	<HelpBox danger handleClose={() => true}>
+	<HelpBox variation='danger' handleClose={() => true}>
```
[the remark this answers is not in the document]
smaxa20: Ok great, I'll switch it.

smaxa20:
@bryanrsmith do we want to make it standard to use either the shorthand or longhand version of the space props? (e.g. Should we decide whether to use `marginTop` or `mt`?)

components/accordion/accordion-header.jsx
```diff
@@ -56,25 +64,66 @@ export function AccordionHeader({ ariaLevel, children, renderCustomIndicator, su
...
+								) : null}
+							</Box>
 						</React.Fragment>
-					</Styled.ButtonContentWrapper>
+					</Box>
 				</Styled.Button>
 			</Styled.Heading>
 			{renderCustomIndicator ? (
-				<Styled.Indicator>
+				<Box gridColumn="indicator" gridRow="1" marginTop={5} marginRight={[5, 5, 6]}>
```
bryanrsmith: No objection to this, but you _can_ use `null` to skip breakpoints. `marginRight={[5, null, 6]}` should be equivalent.

components/accordion/accordion-panel.jsx
```diff
@@ -1,19 +1,29 @@
...
+import { Box } from '../Box';
 import { Collapse } from '../collapse';
 import { useAccordionItemContext } from './accordion-util';
-import * as Styled from './styled-panel';
 
 export function AccordionPanel({ children }) {
 	const { isExpanded, headerId, panelId, styleOverrides } = useAccordionItemContext();
 	return (
-		<Styled.PanelContainer>
+		<Box minWidth="0px">
```
bryanrsmith: `Box` has `min-width: 0` set on it by default.

components/accordion/accordion-panel.jsx
```diff
@@ -1,19 +1,29 @@
...
 			<Collapse isOpen={isExpanded}>
-				<Styled.Panel headerId={headerId} panelId={panelId} styleOverrides={styleOverrides}>
+				<Box
+					headerId={headerId}
+					panelId={panelId}
+					styleOverrides={styleOverrides}
```
bryanrsmith: Several of these props don't seem right.

components/check-box/checkbox-content.jsx
```diff
@@ -1,15 +1,56 @@
...
+				backgroundColor={disabled && 'gray8'}
+			>
+				<Box
+					position="absolute"
+					top="0"
+					left="0"
+					width="14px"
+					height="14px"
+					css={`
+						cursor: ${disabled ? 'default' : 'pointer'};
```
bryanrsmith: This doesn't seem right. `cursor: pointer` should only go on interactive element. Should there be a button here somewhere?

components/check-box/checkbox-content.jsx
```diff
@@ -1,15 +1,56 @@
...
+					/>
+				</Box>
+			</Styled.HoverableBox>
+			{title && (
+				<Styled.Label marginLeft="22px" siblingMarginLeft="6px" fontWeight="regular" fontSize={2}>
+					{title}
+				</Styled.Label>
+			)}
+			{children && (
+				<Styled.Label marginLeft="22px" siblingMarginLeft="6px">
```
bryanrsmith: We should try to stick to the spacing scale for margins.

bryanrsmith: I should add a `horizontal` mode to `Stack` for cases like this.

components/check-box/styled.jsx
```diff
@@ -56,42 +31,27 @@ export const CheckboxContainer = styled.button`
...
-		opacity: 0;
+export const HoverableBox = styled(Box)`
+	@media (hover: hover) {
+		&:hover {
+			${system({ hoverBorderColor: { property: 'borderColor', scale: 'colors' } })};
+		}
+	}
+	@media (hover: none) {
+		&:active {
+			${system({ hoverBorderColor: { property: 'borderColor', scale: 'colors' } })};
```
bryanrsmith: You could avoid the duplication by pulling the function returned by `system()` out into a local.

components/accordion/accordion-header.jsx
```diff
@@ -56,25 +64,66 @@ export function AccordionHeader({ ariaLevel, children, renderCustomIndicator, su
...
-							<Styled.ButtonContent>
-								{children ? <Styled.Title>{children}</Styled.Title> : null}
-								{subtitle ? <Styled.Subtitle>{subtitle}</Styled.Subtitle> : null}
-							</Styled.ButtonContent>
+							<Box
+								display="inline-grid"
+								gridTemplateColumns="min-content auto"
+								gridGap={6}
+								alignItems="center"
+								css={{ whiteSpace: 'nowrap' }}
```
bryanrsmith: Seems like this should be on the Text.

components/drop-zone/component.jsx
```diff
@@ -51,15 +53,30 @@ export class DropZone extends PureComponent {
...
+				justifyContent="center"
+				alignItems="center"
+				padding={4}
+				border="2px dashed"
+				borderColor="gray22"
+				background={showHighlight && 'blue1'}
+				fontSize={3}
+				fontWeight="regular"
+				lineHeight="26px"
+				css={{ transition: 'background 1s ease' }}
```
bryanrsmith: Let me know when you find missing properties & I'll add system props for them.

components/drop-zone/component.jsx
```diff
@@ -51,15 +53,30 @@ export class DropZone extends PureComponent {
...
+				flexDirection="column"
+				justifyContent="center"
+				alignItems="center"
+				padding={4}
+				border="2px dashed"
+				borderColor="gray22"
+				background={showHighlight && 'blue1'}
+				fontSize={3}
+				fontWeight="regular"
+				lineHeight="26px"
```
bryanrsmith: Suspicious.

components/files-section/component.jsx
```diff
@@ -107,10 +110,18 @@ export class FilesSection extends PureComponent {
 		const lastFileIndex = files.length - 1;
 
 		return (
-			<Styled.FilesSection>
-				<Styled.Title fontSize={titleFontSize}>{title}</Styled.Title>
-				<Styled.Divider />
-				<Styled.FilesContainer>
+			<Container>
+				<Text
+					display="block"
```
bryanrsmith: Do you want `Paragraph` instead?

components/files-section/component.jsx
```diff
@@ -120,29 +131,45 @@ export class FilesSection extends PureComponent {
...
-							{index < lastFileIndex && <Styled.Divider />}
+							{index < lastFileIndex && <Box borderBottom={1} borderColor="gray14" />}
 						</React.Fragment>
 					))}
-				</Styled.FilesContainer>
+				</Box>
 				{onUploadFiles ? (
 					<DropZone onDrop={onUploadFiles}>
-						<Styled.DropZoneText>{dropZoneText}</Styled.DropZoneText>
+						<Text textAlign="center" fontWeight="regular" lineHeight="26px">
```
bryanrsmith: Does the `textAlign` do anything here?

components/files-section/component.jsx
```diff
@@ -120,29 +131,45 @@ export class FilesSection extends PureComponent {
...
 					</DropZone>
 				) : null}
-			</Styled.FilesSection>
+			</Container>
 		);
 	}
 }
+
+const Container = styled.section`
+	${resetStyles};
```
bryanrsmith: Should we add a `StyleReset` component that could be composed with Box? `<StyleReset as={Box}>` or whatever.

components/files-section/file-item.jsx
```diff
@@ -89,19 +103,36 @@ export class FileItem extends PureComponent {
...
 				) : null}
-			</Styled.FileItem>
+			</FileItemBox>
 		);
 	}
 }
+
+const FileItemBox = styled(Box)`
+	* {
+		${system({ childFlexShrink: { property: 'flex-shrink', scale: 'none' } })};
```
bryanrsmith: I think you want to just omit `scale`. This seems like it would try to look up a scale named "none" on the theme.

components/modal/default-modal-footer.jsx
```diff
@@ -20,7 +26,11 @@ export const DefaultModalFooter = props => (
 				</Button>
 			)}
 			{props.cancelButton && (
-				<Styled.CancelContainer>
+				<Box
+					width={props.useFullWidthButtons && '100%'}
+					marginBottom={props.useFullWidthButtons && 5}
+					marginRight={props.useFullWidthButtons || 5}
```
bryanrsmith: Normally I wouldn't object to this, but for these props I think it might be safer to be explicit with a ternary. It's pretty common for people to be lazy with boolean props, and I could easily imagine someone writing code like `useFullWidthButtons={items.length}` to adjust layout based on the size of some collection. If they did that here, we'd end up passing the numeric prop value to styled-system, where it would get looked up the scale.

components/modal/modal-header.jsx
```diff
@@ -1,17 +1,41 @@
...
+		justifyContent="space-between"
+		alignItems="center"
+		width="100%"
+		marginBottom={4}
+		paddingBottom={5}
+		borderBottom={styleOverrides.bottomBorder ? styleOverrides.bottomBorder : 1}
+		borderColor="gray22"
+	>
+		<Box display="flex" justifyContent="space-between" alignItems="center" width="100%">
+			<Text fontSize={6} fontWeight="semibold" lineHeight="30px" color="gray66">
```
bryanrsmith: For all these, it'd be better to prefer an existing text style. `<Text textStyle="c.16"` or whatever.

components/modal/modal.jsx
```diff
@@ -123,32 +121,49 @@ export class Modal extends React.Component {
...
+						}
+					}}
+					display="flex"
+					flexDirection="column"
+					justifyContent="center"
+					alignItems="center"
+					width="fit-content"
+					height="fit-content"
+					maxWidth="calc(100% - 16px)"
+					maxHeight={['calc(100% - 16px)', 'calc(100% - 16px)', '80%']}
```
bryanrsmith: Here's a spot where a `null` might be nice so you don't have to duplicate that calc expression.

components/accordion/accordion-header.jsx
```diff
@@ -56,25 +64,66 @@ export function AccordionHeader({ ariaLevel, children, renderCustomIndicator, su
...
+							>
+								{children ? (
+									<Text
+										display="grid"
+										color="gray52"
+										letterSpacing="0.5px"
+										fontWeight="semibold"
+										fontFamily="normal"
+										fontSize={3}
+										lineHeight={1}
```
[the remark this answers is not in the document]
smaxa20: Should I just choose the `textStyle` that's most similar then? I've also noticed that none of the `textStyles` have `fontWeight: fontWeight.regular` and as I'm going through the components most text is non-bold.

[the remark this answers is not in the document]
bryanrsmith: > Should I just choose the textStyle that's most similar then?

I think so, unless you have text that's obviously not close. We can ask Amber for feedback or changes as needed. Supposedly what we have in the theme includes every text style we should need.

> I've also noticed that none of the textStyles have fontWeight: fontWeight.regular

`regular` is the default; are they inheriting `bold` somehow?

components/accordion/accordion-header.jsx
```diff
@@ -56,25 +64,66 @@ export function AccordionHeader({ ariaLevel, children, renderCustomIndicator, su
...
+						display="grid"
+						alignItems="center"
+						gridTemplateColumns={hideArrows ? 'auto' : 'min-content auto'}
+						gridColumnGap={4}
+						borderTop={1}
+						borderColor="gray14"
+						background="linear-gradient(180deg, #fafafa, hsla(0, 0%, 100%, 0))"
+						py={5}
+						px={[5, 5, 6]}
+						css={{ lineHeight: '1' }}
```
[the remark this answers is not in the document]
smaxa20: I've had to use it on a box 3 times so far. I wouldn't be opposed to adding it to Box.

components/accordion/accordion-panel.jsx
smaxa20: Those props actually don't need to be included in this tag anymore. I'll just take them out.

components/check-box/checkbox-content.jsx
smaxa20: This makes it so that the pointer only shows up when you're hovering over the actual checkbox, but it doesn't make sense because the button is bigger than that. I'll move it to the button unless there's a reason to only show the pointer when hovering over the box.

components/accordion/accordion-header.jsx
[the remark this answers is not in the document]
smaxa20: > Supposedly what we have in the theme includes every text style we should need.

Sounds good.

> `regular` is the default

I mean on all of the `TextStyles` that we define there are none with `regular` so If I use one of our's it's going to be `semibold`.

> are they inheriting `bold` somehow?

`Text` has `textStyled.c.16` as its default which has `semibold` weight so whenever I use `Text` I'm inheriting that.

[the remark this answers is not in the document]
bryanrsmith: Oh, hah. Now I understand. That seems like a copy/paste error... I wouldn't expect any of the content fonts to be semibold. I'll get that fixed. Sorry.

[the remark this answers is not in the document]
bryanrsmith: Definitely a copy/paste error. Only the headings should have been semibold. I've pushed a fix.

components/accordion/accordion-header.jsx
smaxa20: You added it yesterday to the `typography` helper. I guess I should just use it as if it's included so when this gets merged in it won't need to be changed again.

components/drop-zone/component.jsx
smaxa20: Anything I have to use the `css` prop for?

components/drop-zone/component.jsx
smaxa20: What's suspicious?

components/files-section/component.jsx
smaxa20: I looked into it. `Paragraph` has baked in top and bottom margins so it was either use `Text` and specify `display="block"` or use `Paragraph` and specify `marginTop={0}`. Since this is a title I figured `Text` was a better representation of what it is used as.

components/files-section/component.jsx
smaxa20: Nope.

components/files-section/component.jsx
bryanrsmith: From the user agent styles? Probably better for us to reset that to 0 and make you put it in a Stack for managing spacing.

components/files-section/component.jsx
smaxa20: Maybe. It could be something that we just wrap around every component just to make sure everything is clean when we start styling it.

components/modal/default-modal-footer.jsx
smaxa20: Ok that's fair.

components/files-section/component.jsx
smaxa20: Nope.

components/files-section/component.jsx
smaxa20: I think it's from the underlying html `<p>` tag. I agree that we should reset it.

components/accordion/accordion-header.jsx
smaxa20: Do you mean `whiteSpace`?

components/accordion/accordion-header.jsx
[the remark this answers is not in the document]
smaxa20: Ok that makes sense. Thanks.

components/help-box/component.jsx
```diff
@@ -85,20 +172,43 @@ HelpBox.propTypes = {
 		backgroundColor: PropTypes.string,
 		closeIconColor: PropTypes.string,
 	}),
-	/** Green theme */
+	/** Enum with values: 'success', 'danger', 'warning', and 'minor' */
+	variant: PropTypes.oneOf(['success', 'danger', 'warning', 'minor']),
+	/** Green theme (depricated in favor of the variation prop) */
```
bryanrsmith: Typo: deprecated (more below)

components/popover/popover-base.jsx
```diff
@@ -160,30 +159,38 @@ export class PopoverBase extends Component {
...
+							...style,
+							...(!hideArrow ? Styled.margins[Styled.getPlacement(placement)] : {}),
+						}}
+						delay={delay}
+						hideArrow={hideArrow}
+						styleOverrides={styleOverrides}
+						onAnimationEnd={this.handleTransition}
+						background="white"
+						boxShadow={1}
+						fontSize="c.16"
```
bryanrsmith: "c.16" is a named font style. It can't be used for `fontSize`. Also, I don't think `Box` accepts typography props.

components/popover/popover-base.jsx
```diff
@@ -160,30 +159,38 @@ export class PopoverBase extends Component {
...
-							)}
-						</Styled.PopoverBase>
-					</ThemeProvider>
+					<Box
+						ref={ref}
+						style={{
+							...style,
+							...(!hideArrow ? Styled.margins[Styled.getPlacement(placement)] : {}),
+						}}
+						delay={delay}
```
bryanrsmith: What's this?

components/popover/popover-base.jsx
```diff
@@ -160,30 +159,38 @@ export class PopoverBase extends Component {
...
+						hideArrow={hideArrow}
+						styleOverrides={styleOverrides}
```
bryanrsmith: I think we have some old props to remove here.

components/popover/styled.jsx
```diff
@@ -19,73 +19,15 @@ export const getPlacement = placement => {
 	return placement.split('-')[0];
 };
 
-export const PopoverBase = styled.div`
```
bryanrsmith: Ahhhhhhhhh ☺️

components/tabs/styled.jsx
```diff
@@ -66,47 +66,88 @@ export const Tab = styled.button.attrs({
...
+// })`
+// 	position: relative;
+// 	padding: ${thickness.eight};
+
+// 	&:focus {
+// 		outline: none;
+// 	}
+
+// 	${({ selected }) => !selected && 'display: none'};
+// `;
```
bryanrsmith: Is the commented code in this file ready to be removed?

smaxa20:
> Yeah, just wondering why whiteSpace is set on the Box instead of on the Text children. I guess it avoids a line break between the spans, and is reasonable to put on the container here.

Probably because it applies to both the title and the subtitle, but it works just as well and without having to use the `css` prop if I put it on its `<Text>` tag children, so I'll do that.

(don't know why it's not letting me respond to your comment)

components/accordion/accordion-header.jsx
```diff
@@ -56,25 +65,56 @@ export function AccordionHeader({ ariaLevel, children, renderCustomIndicator, su
...
+								gridTemplateColumns="min-content auto"
+								gridGap={6}
+								alignItems="center"
+								css={{ whiteSpace: 'nowrap' }}
+							>
+								{children ? (
+									<Text
+										display="grid"
+										color="gray52"
+										letterSpacing="0.5px"
```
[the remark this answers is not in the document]
smaxa20: I don't see why not. It changes it a little bit, but it still looks fine to me.

components/check-box/checkbox-content.jsx
```diff
@@ -20,3 +54,29 @@ CheckboxContent.propTypes = {
 	isChecked: PropTypes.bool,
 	disabled: PropTypes.bool,
 };
+
+const hoverBorderColor = { hoverBorderColor: { property: 'borderColor', scale: 'colors' } };
+const HoverableBox = styled(Box)`
+	@media (hover: hover) {
+		&:hover {
+			${system(hoverBorderColor)};
```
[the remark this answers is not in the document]
smaxa20: Ah, I see.

components/check-box/component.jsx
```diff
@@ -73,3 +71,31 @@ Checkbox.defaultProps = {
...
+const CheckboxContainer = styled.button`
+	${resetStyles};
+
+	display: flex;
+	align-items: center;
+	position: relative;
+	border: none;
+	padding: 0px 16px;
+	min-height: 44px;
+	min-width: 44px;
```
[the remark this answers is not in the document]
smaxa20: Yes, I wasn't sure whether to roll it back here as well or to leave it until we merge and deal with it there.

components/files-section/file-item.jsx
```diff
@@ -89,19 +102,33 @@ export class FileItem extends PureComponent {
...
 				) : null}
-			</Styled.FileItem>
+			</FileItemBox>
 		);
 	}
 }
+
+const FileItemBox = styled(Box)`
+	* {
+		${system({ childFlexShrink: { property: 'flex-shrink' } })};
```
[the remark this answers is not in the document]
smaxa20: It should be. Thanks for that catch.

components/popover/popover-base.jsx
smaxa20: The delay before a popover will show up and disappear I believe.

components/popover/popover-base.jsx
smaxa20: You're right. Nice catch.

components/check-box/component.jsx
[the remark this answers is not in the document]
bryanrsmith: We'll need to resolve conflicts before we can merge. Easiest is probably to merge changes from master into this branch periodically and resolve the conflicts as they come up.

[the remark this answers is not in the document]
smaxa20: This branch as in `vnext` or as in this branch that I'm PRing?

components/popover/styled.jsx
smaxa20: Yes it was very nice to be able to delete that one haha.

components/tabs/styled.jsx
smaxa20: Oh yeah, missed that. Thanks.

components/popover/popover-base.jsx
bryanrsmith: It looks like an extra prop that isn't supported by Box. I think I left this before I noticed what was happening and added the other comment about that :-)

smaxa20: Yeah, it wasn't necessary after moving to Boxes. I removed it.

catalog/accordion/variations.md
```diff
@@ -29,7 +30,8 @@ state: { expandedSections: [0] }
 		</Accordion.Item>
 		<Accordion.Item>
 			<Accordion.Header>
-				Section Two Title
+				<Accordion.Title>Section One Title</Accordion.Title>
```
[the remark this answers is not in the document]
smaxa20: Whoops, copy paste error.

catalog/button/variations.md
```diff
@@ -43,30 +43,30 @@ showSource: true
...
+	<Button variant="primaryOutline" size="small" ref={buttonRef}>
 		Target Button
 	</Button>
 </ButtonDemo>
 ```
 
 ### Supported style customizations
 
-Do not use the `style` prop to style this component (it will be ignored). Instead, if there is missing customization that you need for this component, ask to have it added. The `theme` prop can be used to control color variations while the `styleOverrides` prop can be used for other styles.
+Do not use the `style` prop to style this component (it will be ignored). Instead, if there is missing customization that you need for this component, ask to have it added.
```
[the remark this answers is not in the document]
smaxa20: Well it still stands that if there's something missing they can ask to add it and we could do that. I don't know how it handles the `style` prop now though.

components/input/component.jsx
```diff
@@ -56,33 +65,103 @@ export const Input = forwardClassRef(
...
 					value={value || ''}
 					placeholder={placeholder || ''}
 					onChange={this.handleChange}
 					onClick={onClick}
 					onKeyPress={this.handleKeyPress}
 					ref={forwardedRef}
-					{...filteredProps || {}}
+					height="32px"
+					padding={3}
+					textStyle={variant === 'large' || large ? 'ui.18' : 'ui.16'}
```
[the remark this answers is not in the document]
smaxa20: It's only awkward because how I do the variation doesn't allow consumers to change anything styled specific to a variation like `textStyle="ui.18"` would be for `large`. Also I wanted to use `textStyle` instead of `fontSize` etc. 

If we change to using Styled-System variations api it would be able to become better I think.

components/loading-spinner/component.jsx
```diff
@@ -1,41 +1,78 @@
...
-	const { component: MappedStyledComponent, filteredProps } = applyVariations(
-		Styled.Spinner,
-		Styled.variationMap,
-		restProps,
+export function LoadingSpinner({ variant, small, medium, large, height, width, ...restProps }) {
+	return (
+		<Spinner
+			variant={getVariation(variant, { small, medium, large, none: true })}
+			display="inline-block"
+			width={height || width}
```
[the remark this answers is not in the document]
smaxa20: We want it to always be a circle so I use the value for height or width (whichever is specified) for both height and width. I had to put the same value first in the `or` or else if they specify both height and width it could end up being an oval.

components/loading-spinner/component.jsx
```diff
@@ -1,41 +1,78 @@
...
+			height={height || width}
+			marginLeft="5px"
```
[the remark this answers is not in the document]
smaxa20: Nope, not that I can tell. Changing it to `{2}`

PR #209, opened 2019-07-23
Allow non-admin users to select groups when desired

dustymack:
Add resource strings where needed


components/group-selector/large/search-result.jsx
```diff
@@ -170,3 +246,7 @@ export class SearchResult extends React.PureComponent {
 		);
 	}
 }
+
+SearchResult.defaultProps = {
```
bryanrsmith: Since this is a class component, property initializer syntax would be more appropriate.

components/group-selector/large/component.jsx
```diff
@@ -48,6 +68,16 @@ export class LargeGroupSelector extends React.Component {
 		scrollWidthDelta: 0,
 	};
 
+	componentDidMount() {
+		if (this.props.showInPlace) {
+			window.addEventListener('scroll', this.handleScroll);
```
bryanrsmith: This doesn't seem like a safe assumption. What if the scroll ancestor is not window?

catalog/group-selector/variations.md
```diff
@@ -124,52 +134,203 @@ state: {
...
-		groups={state.groups}
-		selectedGroupId={state.selectedGroupId}
-		groupSearchResults={state.groupSearchResults || state.groups}
-		groupSelectorView={"groups"}
 		onGetStartedClick={() => {alert("Lets get started.")}}
 		onClaimGroupClick={() => {alert("Claim the group!")}}
 		onJoinGroupClick={() => {alert("Should Join the group, or request")}}
 		onAdminRequestClick={() => {alert("Admin access requested")}}
-		isMobile={false}
+		showInPlace
```
bryanrsmith: What does this prop mean? The documentation isn't helping me.

components/group-selector/large/component.jsx
```diff
@@ -305,3 +340,21 @@ export class LargeGroupSelector extends React.Component {
...
+	authorizedMembershipLevels: ['admin'],
+	authorizedGroupKinds: ['church'],
+	resources: {
+		title: 'Find Your Church',
+		subTitle: 'in the Faithlife Church Directory',
+		requestButtonText: 'Request Admin',
+		joinButtonText: 'Join Group',
+		claimButtonText: 'Claim',
+		selectButtonText: 'Get Started',
+		dontSeeChurchText: "Don't see your church?",
```
bryanrsmith: We should probably use the correct typographical quotes in the default strings.

dustymack:
> I can't tell from reading; is this a breaking change for existing consumers of the component?

The intention is to be fully backwards compatible. Some of the new copy applies to existing installations, but that is intended to be a global improvement.

catalog/group-selector/variations.md
```diff
@@ -7,7 +7,7 @@ state: {
 		name: 'Michaels Test Church',
 		groupId: 1,
 		kind: 'church',
-		avatarUrl: 'https://avatars.logoscdn.com/451/6830451_f896e509e7504b33812f0f6bf086ffa8.jpg',
+		avatarUrl: '',
```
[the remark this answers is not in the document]
dustymack: This isn't a real change, just a bad `diff``.  I moved the `LargeGroupSelector` versions to the top in order to show the new scrolling behavior when `showInPlace=true`

catalog/group-selector/variations.md
```diff
@@ -36,16 +36,25 @@ state: {
 		kind: 'church',
 		avatarUrl: '',
 		relationshipKind: 'participant',
+		membershipKind: '',
```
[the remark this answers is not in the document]
dustymack: That is what I was testing with this one, invalid/different `membershipKind`. It looked like that edge case was already in the button behaviors.

catalog/group-selector/variations.md
```diff
@@ -36,16 +36,25 @@ state: {
...
 	],
-	selectedGroupId: 1,
-	mobile: false,
 }
 ---
-<GroupSelectorDemo>
-	<GroupSelector
+<LargeGroupSelectorDemo>
+	<LargeGroupSelector
+		isOpen={false}
```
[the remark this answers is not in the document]
dustymack: Not a change, just a bad `diff`. I believe this is just a starting value, though.

components/group-selector/large/component.jsx
```diff
@@ -35,6 +35,26 @@ export class LargeGroupSelector extends React.Component {
...
 		/** Whether or not to show the "Find Your Church in the Faithlife Group Directory" title */
 		hideTitle: PropTypes.bool,
+		/** Array of membership levels allowed to "Select" group. Defaults to Admin only */
+		authorizedMembershipLevels: PropTypes.arrayOf(PropTypes.string),
+		/** Array of group kinds user is allowed to "Select". Defaults to 'church' only */
+		authorizedGroupKinds: PropTypes.arrayOf(PropTypes.string),
+		/** Flag to use "Select"/"Request" button style instead of "Get Started"/"Join"/"Follow"/"Claim" */
+		useSelectRequestButtonStyle: PropTypes.bool,
+		/** String literals to overload UI elements and for localization */
+		resources: PropTypes.shape({
```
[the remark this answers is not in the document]
dustymack: Happy to rename it. Skeletor habits.

components/group-selector/large/component.jsx
```diff
@@ -56,33 +86,25 @@ export class LargeGroupSelector extends React.Component {
...
-				toggle={this.toggle}
-			/>
-		));
-	}
+	formatStringList = arrayOfStrings => {
+		const result =
+			arrayOfStrings.length === 1
+				? arrayOfStrings[0]
+				: [arrayOfStrings.slice(0, -1).join(', '), arrayOfStrings.slice(-1)[0]].join(
+						arrayOfStrings.length < 2 ? '' : ' or ',
```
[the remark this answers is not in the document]
dustymack: It isn't language localization yet, really. It is the ability to override labels that the actual case calls for. Full language localization is going to require more work. I like anticipating that work if we have a good pattern for it.

components/group-selector/large/search-result.jsx
```diff
@@ -59,24 +64,56 @@ export class SearchResult extends React.PureComponent {
...
+			message = (
+				<Tooltip
+					content={
+						<>
+							<Styled.SearchResultMessage>
+								Claiming this group will make you the admin.
+							</Styled.SearchResultMessage>
+							<br />
+							<Styled.SearchResultMessage>
+								You can invite others to the group, assign additional admins, and transfer admin
```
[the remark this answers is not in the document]
dustymack: It was outside the scope of my current case to do the complete localization. I will review how close I got and if it's worth getting the rest on this one.

catalog/group-selector/variations.md
dustymack: Pre-existent. It adds the component as an in-place element instead of a modal that pops up when clicked. I can see about adding a better description in documentation.

components/group-selector/large/component.jsx
dustymack: True story. Adjusting now.

dustymack:
@bryanrsmith I believe I have addressed all review so far.

VersionHistory.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+### 5.10.1
```
bryanrsmith: The changes you've made are feature additions, not bug fixes, so this should be a minor version change. That said, we make version number changes at release time so that multiple PRs don't get us mixed up.

components/group-selector/large/component.jsx
```diff
@@ -48,6 +86,17 @@ export class LargeGroupSelector extends React.Component {
 		scrollWidthDelta: 0,
 	};
 
+	componentDidMount() {
+		if (this.props.showInPlace) {
+			var scrollableNode = this.getScrollParent(this.node);
+			scrollableNode.addEventListener('scroll', this.handleScroll.bind(this));
```
bryanrsmith: `Function.prototype.bind()` creates a new function, so by inlining this you've registered a handler that cannot be removed (because you don't have a reference to the function).

We don't typically use `.bind()` here, instead preferring arrow function properties.

components/group-selector/large/component.jsx
```diff
@@ -48,6 +86,17 @@ export class LargeGroupSelector extends React.Component {
...
+		}
+	}
+
+	componentWillUnmount() {
+		window.removeEventListener('scroll', this.handleScroll);
```
bryanrsmith: This needs to be updated to match the new didMount logic.

components/group-selector/large/component.jsx
```diff
@@ -48,6 +86,17 @@ export class LargeGroupSelector extends React.Component {
 		scrollWidthDelta: 0,
 	};
 
+	componentDidMount() {
+		if (this.props.showInPlace) {
+			var scrollableNode = this.getScrollParent(this.node);
```
bryanrsmith: Don't use var. Does this produce lint errors? There are several more uses of `var` lower down as well.

components/group-selector/large/component.jsx
```diff
@@ -56,33 +86,25 @@ export class LargeGroupSelector extends React.Component {
...
-				onJoinGroupClick={this.props.onJoinGroupClick}
-				onClaimGroupClick={this.handleClaimGroup}
-				setModalState={this.setModalState}
-				setSelectedGroupId={this.setSelectedGroupId}
-				onGetStarted={this.handleGetStarted}
-				toggle={this.toggle}
-			/>
-		));
-	}
+	formatStringList = arrayOfStrings => {
```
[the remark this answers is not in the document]
bryanrsmith: This has not yet been addressed.

components/group-selector/large/component.jsx
```diff
@@ -56,33 +86,25 @@ export class LargeGroupSelector extends React.Component {
...
+						arrayOfStrings.length < 2 ? '' : ' or ',
+				  );
+		return result.charAt(0).toUpperCase() + result.slice(1);
+	};
 
 	getSearchResults = () => {
 		let groups;
-		if (this.props.groupSearchResults) {
-			groups = this.props.groupSearchResults.map(group => (
+		const groupResults = this.props.groupSearchResults || this.props.groups;
```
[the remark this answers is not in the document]
bryanrsmith: This has not yet been addressed.

components/group-selector/large/component.jsx
```diff
@@ -171,34 +219,65 @@ export class LargeGroupSelector extends React.Component {
 		this.props.onClaimGroupClick(groupId);
 	};
 
-	handleScroll = scrollData => {
-		if (this.props.showInPlace) {
-			return;
+	getScrollParent(element, includeHidden) {
```
bryanrsmith: Move out to a helper function instead of an instance method. Was this copied or adapted from somewhere? If so, please add a comment indicating where.

components/group-selector/large/component.jsx
```diff
@@ -171,34 +219,65 @@ export class LargeGroupSelector extends React.Component {
...
+		var style = getComputedStyle(element);
+		var excludeStaticParent = style.position === 'absolute';
+		var overflowRegex = includeHidden ? /(auto|scroll|hidden)/ : /(auto|scroll)/;
+
+		if (style.position === 'fixed') {
+			return document.body;
+		}
+		for (var parent = element; (parent = parent.parentElement); ) {
```
bryanrsmith: This is hard to read. How would you feel about a while loop instead?

components/group-selector/large/component.jsx
bryanrsmith: Actually, it looks like `handleScroll` is already an error function, so you should be able to just remove the `.bind()`.

components/group-selector/large/component.jsx
```diff
@@ -245,7 +327,7 @@ export class LargeGroupSelector extends React.Component {
 			this.state.modalContent === 'admin' || this.state.modalContent === 'change';
 
 		return (
-			<Styled.LargeGroupSelector>
+			<Styled.LargeGroupSelector ref={node => (this.node = node)}>
```
bryanrsmith: * Be deliberate about returns types. This should have curly braces to clarify that the function doesn't have a meaningful return value.
* But prefer `React.createRef()` over function refs.

components/group-selector/large/create-group.jsx
```diff
@@ -19,34 +20,38 @@ export class CreateGroup extends React.Component {
...
 
 	render() {
+		const {
+			showRequiredStars,
+			newChurchName,
+			newChurchLocation,
+			onChurchNameInputChange,
+			onChurchLocationInputChange,
+			localizedResources,
+		} = this.props;
```
bryanrsmith: 👍 Like this, in every method that accesses `this.props` or `this.state`.

components/group-selector/large/search-result.jsx
```diff
@@ -59,24 +64,56 @@ export class SearchResult extends React.PureComponent {
...
 		if (claimable) {
-			message = <Styled.SearchResultMessage>This is an empty group</Styled.SearchResultMessage>;
+			message = (
+				<Tooltip
+					content={
+						<>
+							<Styled.SearchResultMessage>
+								Claiming this group will make you the admin.
+							</Styled.SearchResultMessage>
+							<br />
```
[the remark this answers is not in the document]
bryanrsmith: This has not yet been addressed.

components/group-selector/large/component.jsx
dustymack: prettier seems to insist on this over `'Don\'t see your church?'`

bryanrsmith: `'Don’t see your church?'`

components/group-selector/large/component.jsx
dustymack: That is what I had originally and my `this` context seemed lost. Trying again as there were other variables at play.

components/group-selector/large/component.jsx
dustymack: It did not product eslint errors. Fixing...

components/group-selector/large/component.jsx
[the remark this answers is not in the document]
dustymack: I can do this here. Not sure why it wasn't that way originally.

components/group-selector/large/search-result.jsx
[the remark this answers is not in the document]
dustymack: I am assuming this is: Use css in place of the `<br/>`?

dustymack:
Not sure how I missed some of those. I think I covered the remaining pieces.

components/group-selector/large/component.jsx
```diff
@@ -35,6 +35,44 @@ export class LargeGroupSelector extends React.Component {
...
+		authorizedMembershipLevels: ['admin'],
+		authorizedGroupKinds: ['church'],
+		localizedResources: {
+			title: 'Find Your Church',
+			subTitle: 'in the Faithlife Church Directory',
+			requestButtonText: 'Request Admin',
+			joinButtonText: 'Join Group',
+			claimButtonText: 'Claim',
+			selectButtonText: 'Get Started',
+			dontSeeChurchText: "Don't see your church?",
```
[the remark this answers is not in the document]
dustymack: Not seeing the previous conversation now in Git, but I responded there. Do you have a preferred syntax here? 
Prettier converts `Don\'t see your church?'` automatically to this format. 

Would you prefer: 
```
dontSeeChurchText: `Don't see your church?`,
```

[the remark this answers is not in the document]
bryanrsmith: You're using a straight quote instead of a curly quote. If you use a curly quote the text will look typographically correct to the user, and you can use the standard quote marks for the string. I responded earlier with the exact code to use.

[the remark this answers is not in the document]
dustymack: Thank you, I will hunt down that earlier comment.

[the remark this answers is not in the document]
dustymack: I did not find your earlier comment, but your description helped me find it, thanks again.

PR #210, opened 2019-07-24
Reset on viewport change

ronnywoods:
Examples don't set isSmallViewport property dynamically (may submit as separate PR); noted the need for these changes based on use in separate project that grid does not adjust to new rowHeight or column visibility properties on resizing of grid without resetting column state and row height.  Thus, resetting column state and row height on grid anytime relevant new props are received.

Changes from previous PR (adding rowHeight) are also still shown on the diff below.

### DIFF
--- a/components/grid/base-table.jsx
+++ b/components/grid/base-table.jsx
@@ -118,8 +118,9 @@ export function BaseTable({
 	const rowCount = data ? data.length : 1;
 	const currentHeaderHeight = hideHeaders ? 0 : headerHeight;
 	const tableHeight =
-		(maxRows && maxRows < rowCount ? maxRows : rowCount) * rowHeight ||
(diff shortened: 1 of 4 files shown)


components/grid/base-table.jsx
```diff
@@ -168,6 +180,7 @@ export function BaseTable({
 						<AgGridColumn
 							{...columnProps}
 							key={fieldName}
+							colId={fieldName}
```
korbinancell: `colId` is inferred to be `field` unless it is specified, so I would rather leave it unsaid unless it is set by a consumer.

ronnywoods: I can't remember now why I felt I needed to add that separately (wish I had thought to include a note)...but worst case consumer can add it, so I'll go ahead and remove it now.

PR #213, opened 2019-07-30
Set 'no rows' height

ronnywoods:
My last PR removed superfluous table padding, but this created an issue when table doesn't have any data (entire table is collapsed and there's no room for 'no rows to show' message).  This PR fixes this by setting a table height of 200px whenever there is no data to display.

Diff will also include changes from my last yet-to-be-merged PR.

### DIFF
--- a/components/grid/base-table.jsx
+++ b/components/grid/base-table.jsx
@@ -118,8 +118,9 @@ export function BaseTable({
 	const rowCount = data ? data.length : 1;
 	const currentHeaderHeight = hideHeaders ? 0 : headerHeight;
 	const tableHeight =
-		(maxRows && maxRows < rowCount ? maxRows : rowCount) * rowHeight ||
-		defaultRowHeight + tableHeightPadding + currentHeaderHeight;
+		(maxRows && maxRows < rowCount ? maxRows : rowCount) * (rowHeight || defaultRowHeight) +
+		tableHeightPadding +
+		currentHeaderHeight;
 	return (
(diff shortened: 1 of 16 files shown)


components/grid/base-table.jsx
```diff
@@ -168,6 +183,7 @@ export function BaseTable({
 						<AgGridColumn
 							{...columnProps}
 							key={fieldName}
+							colId={fieldName}
```
korbinancell: Looks like the changes for the previous pr ended up here. If you fix the colId here then we can just close the previous one,

components/grid/base-table.jsx
```diff
@@ -6,6 +6,7 @@ import * as Styled from './styled';
 
 const defaultRowHeight = 45;
 const headerHeight = defaultRowHeight - 5;
+const noRowsTableHeight = 200;
```
korbinancell: I wonder if you can localize the `No Rows Found` text or whatever it says. Probably not an issue now until we figure out how we want to localize.

components/grid/base-table.jsx
ronnywoods: Just fixed in other PR but I'll go ahead and fix here as well.

components/grid/base-table.jsx
ronnywoods: Yeah, it seemed like this text was buried in some grid settings somewhere...In our project I'm localizing within each `cellComponent` but we're only gonna support english initially anyway, so not too worried about 'No Rows Found' text for now.

PR #220, opened 2019-08-08
Parameter updates

korbinancell:
Would allowing the return to be used as a ref be neat, or not work like it does in my head? There would be some weird side effects if it was used as a dependency I suppose.
Like the following:

```
export function useCopyRefs(refs) {
        const usableRef = useRef();
	const copy = useCallback(
		ref => {
			if (refs) {
				for (const item of refs) {
					if (item) {
						if (typeof item === 'function') {
							item(ref);
						} else {
							item.current = ref;
						}
					}
				}
			}
		},
		[refs],
	);

        copy.current = usableRef.current;
 	return copy;
}
```


bryanrsmith:
Sorry, I'm not understanding your comment. What's your goal, and how would it be used?

PR #221, opened 2019-08-09
Expose minHeight and maxHeight options for table size

ronnywoods:
Max height may be a more desirable property than maxRows in some cases (e.g. if desired height does not align perfectly with row heights), so it would be useful to expose both options for setting max table height.

Min height allows consumers to ensure that table does not shrink below a certain size.  Another option here would be to standardize this within the shared component (i.e. don't allow an option here)...either way, it seems a min height needs to be set, at least on large viewport (otherwise table w/ limited number of rows can get so short that options menu cannot fit on right click).

### DIFF
--- a/components/grid/base-table.jsx
+++ b/components/grid/base-table.jsx
@@ -141,14 +141,13 @@ export function BaseTable({
 			  tableHeightPadding +
 			  currentHeaderHeight
 			: noRowsDefaultTableHeight;
-	const tableHeight =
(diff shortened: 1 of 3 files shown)


ronnywoods:
Turns out it would be really useful to be able to specify minHeight and maxHeight in some form other than pixels, so I'm going to adjust this PR to make both minHeight and maxHeight strings (even though other grid height attributes, like rowHeight, are types as number (of pixels)).  Adjustment to PR on its way...

Issue #230, opened 2019-08-20
Add a "partially checked" state to Checkbox


kbsletten:
With nested check boxes, it's helpful to differentiate "all selected" vs "some selected", but our checkbox component has not in-between state.

PR #232, opened 2019-08-21
Add mixed state to checkboxes

kbsletten:
To support nested options where a checkbox may be a header and need to reflect a partially-checked state, this adds support for a mixed mode including aria changes.

Addresses issue: https://github.com/Faithlife/styled-ui/issues/230

### DIFF
--- a/catalog/checkbox/variations.md
+++ b/catalog/checkbox/variations.md
@@ -18,13 +18,12 @@ state: { isChecked: false }
 
 ```react
 showSource: true
-state: { isMixed: true, isChecked: false }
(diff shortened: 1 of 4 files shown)


kbsletten:
@bryanrsmith, does this seem reasonable?

kbsletten:
@bryanrsmith to avoid a vestigial prop, I've redone it to just allow 'mixed' as a value to `isChecked`, that way we can expand it to `unset` if we ever need it (though, there's no aria-checked value for unset). I've also checked with Amber, and except for the size being wrong (should be 16x16, not 14x14), she thought it looked good.

PR #233, opened 2019-08-21
Add overflow props to popover style overrides

iamthuypham:


components/popover/popover-base.jsx
```diff
@@ -62,6 +62,8 @@ export class PopoverBase extends Component {
 			textAlign: PropTypes.string,
 			width: PropTypes.string,
 			zIndex: PropTypes.number,
+			overflowX: PropTypes.string,
```
korbinancell: There is already an `overflow` style override (looks like it was not added to the proptypes but it does work) does that suit your needs?
If not, then you also need to include both of these in the styled-component styles here: https://github.com/Faithlife/styled-ui/blob/63eaea795212b13b96e61aa8fc5cfcb47fe86823/components/popover/styled.jsx#L73

iamthuypham: @korbinancell Oh, I did not see it is included. Let me try that.

PR #237, opened 2019-09-13
Full width modal content

RobertBolender:
This is needed to support the Smart Media Editor usage inside a Styled-UI modal.

image

The only change to existing consumers is that the border underneath the modal header will extend to the edge of the modal, even on modals that have horizontal padding. I confirmed with Shiloh that this is desired.


bryanrsmith:
I'd like to stop propagating the styleOverrides pattern in favor of more modular designs that can better support our theming strategy and can evolve over time more cleanly. This is slightly awkward when extending existing components, but one reasonable option here would be to allow the consumer to control rendering of the `ModalContent` component. We can modify `ModalContent` to be a `styled(Box)` instead of a `styled.div`, and then add it to the public API. The render method of `Modal` would then change to something like...

```jsx
const doesChildrenIncludeModalContent = React.Children.count(children) === 1 && React.Children.only(children).type === ModalContent;
// etc...
{doesChildrenIncludeModalContent ? children : <ModalContent>{children}</ModalContent>}
```

Then a consumer could optionally choose to render the `ModalContent` wrapper themselves as the root of the children, and pass their own styled-system overrides for padding (or anything else).

What do you think?

My preference for new components like this is to publicly expose all such components from the beginning (for this component, probably Backdrop, Header, Content, and Footer, at least), requiring consumers to opt in to the pieces they want by rendering nested elements rather than drive the entire API through props on a single component. This allows more flexibility for future changes, such as what we're trying to do here, and lets us avoid hacky extensions like the existing `withoutFooter` prop. (If you don't want a footer, you just don't render one).

RobertBolender:
I do like exposing the sub-components, as we do with our `<Dropdown>` and `<Accordion>` components.

For modifying this existing component, would it be acceptable if I were to expose just the `<ModalContent>` component, and defer the rest for later if necessary?

bryanrsmith:
Yes, I think so. We have a difficult task in evolving existing components to a shape we like. Rebuilding them takes time and is difficult to do without breaking consumers. I'm definitely happy to take baby steps if we can keep them all heading in the same direction. Ideally they should all trend toward styled-components APIs with styled-system props for styling needs, and maximum modularity.

PR #243, opened 2019-09-27
Add hover colors for anchor button overrides.

AuresaNyctea:
Currently base `a.hover` styles from faithlife.com are coming through on anchor buttons with the 5.19 update.

I believe this is the correct way to set default hover states on buttons.

### DIFF
--- a/components/button/styled.jsx
+++ b/components/button/styled.jsx
@@ -77,6 +77,11 @@ export const Button = styled.button`
 					primary: {
 						border: '1px solid',
 						color: 'white',
+						'@media (hover: hover)': {
+							'&:hover': {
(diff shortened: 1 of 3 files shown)


components/button/styled.jsx
```diff
@@ -77,6 +77,11 @@ export const Button = styled.button`
 					primary: {
 						border: '1px solid',
 						color: 'white',
+						'@media (hover: hover)': {
+							'&:hover': {
+								color: disabled || 'white',
```
RobertBolender: We should be setting the hover color even when disabled, or we'll have the same color bleeding problem on FLCOM.
I'm not sure whether white or a gray would be appropriate here.

Issue #244, opened 2019-09-27
Product Drawer renders partially offscreen some viewports (SITES-5828)


kevinforrestkeyes:
https://faithlife.atlassian.net/browse/SITES-5828

Product Drawer overflows on viewports under 990px before switching to mobile layout. To fix, I'm going to reposition the product drawer on affected viewport range.

RobertBolender:
It looks like this was resolved in this Skeletor PR: https://git.faithlife.dev/Logos/Skeletor/pull/1823

If that is not the case, please feel free to re-open this issue.

PR #245, opened 2019-09-27
modify ProductDrawerDropdown to stay in viewport on displays

kevinforrestkeyes:
for issue https://github.com/Faithlife/styled-ui/issues/244 from https://faithlife.atlassian.net/browse/SITES-5828

translating the ProductDrawerDropdown a bit to the left between 768px & 992px to make sure it remains fully in view for those viewports, then override that styling when viewport is above 992px

### DIFF
--- a/components/product-drawer/product-drawer-dropdown/styled.jsx
+++ b/components/product-drawer/product-drawer-dropdown/styled.jsx
@@ -33,11 +33,16 @@ export const ProductDrawerDropdown = styled.div`
 		left: auto;
 		bottom: auto;
 		top: auto;
+		transform: translateX(22%);
 		overflow-y: auto;
 		animation-duration: 300ms;
 		animation-timing-function: cubic-bezier(0.33, 0, 0, 1);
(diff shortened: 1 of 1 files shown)


kevinforrestkeyes:
parity with https://git.faithlife.dev/Logos/Skeletor/pull/1823

bryanrsmith:
This seems pretty hacky. Can you provide more details about why a translate is necessary here? I can't see how the width is set, so it seems like there's potential for translate to position improperly. What's the reason for not using left/right instead?

kevinforrestkeyes:
@bryanrsmith I went with `translate` instead of `left` because when I was experimenting with how to accomplish this, the `left` property was having unintended affects on the width of the dropdown which I was unable to account for (until now). I opted for using a `translate` since I was able to achieve the desired affect without altering other aspects of the product drawer. I see now that using `translate` could cause new positioning issues if the width of the drawer were changed. I've determined that using `right` can achieve that same affect without the same pitfalls, and will update my PR shortly. Thanks for the feedback.

Issue #216, opened 2019-08-06
Grid v2 v1


korbinancell:
# This has moved to [FaithlifeEquipment](https://git.faithlife.dev/Logos/FaithlifeEquipment/issues/61)

~A common thread to keep track of questions/issues with the new ag-grid wrapper. Feel free to leave a comment or question and I will add it to the list of things to do.~

~Spec from amber: https://app.zeplin.io/project/58a23c2d4597afb248c13a43/screen/5d56f9a48847a29b54627e3c~

~## Tasks~

- [x]  Rename from Table -> Grid to facilitate adding a simple styled table component in the future.
- [x]  Allow passing a max height directly.
- [x]  Investigate sizeColumnsToFit() does not seem to fire on initial page load in safari.
- [x] Update to spec from amber
- [x] Add simpler way to add props to `BaseGrid` without having to update every grid comp.
- [ ] Export to csv context menu does not work with custom components.
- [ ] Handle server-side row model `SimpleGrid` and `PaginatedGrid`
- [x] Add drag-n-drop to `SimpleGrid`
- [x] Add checkbox support to all grids.
- [ ] Improve no rows found message.
- [ ] Cell editing not supported in TreeGrid.
- [ ] Server side paging
- [ ] Fix "No Row To Show" casing

~## Possible/ As Needed~

- [ ] Switch to svg icons
- [ ] Allow printing mode for client-side row model

ronnywoods:
`sizeColumnsToFit()` does not seem to fire on initial page load in safari.  ref: https://www.flowdock.com/app/logos/fl-comms/threads/lbTgyVKp5yo8o6gPKeKlREZLhJR

ronnywoods:
Just submitted a PR to fix the `sizeColumnsToFit()` issue: https://github.com/Faithlife/styled-ui/pull/218

ronnywoods:
Submitted another PR exposing `maxHeight` and `minHeight`: https://github.com/Faithlife/styled-ui/pull/221

Rationale included in PR description/comments

TaitWeicht:
@korbinancell I noticed that `gridOptions` is [here](https://github.com/Faithlife/styled-ui/blob/vnext/components/grid/base-table.jsx#L23) in the base table, but in the [simple table](https://github.com/Faithlife/styled-ui/blob/vnext/components/grid/simple-table.jsx#L5) there is no way to way to pass `gridOptions` in.
[...]

PR #248, opened 2019-10-02
Add tree grid

korbinancell:
This should solve a few of the issues in https://github.com/Faithlife/styled-ui/issues/216
My plan is after this to go about porting everything to v5


components/grid/base-grid.jsx
```diff
@@ -246,3 +241,34 @@ BaseTable.propTypes = {
 	handleGetRowId: PropTypes.func,
 	children: PropTypes.node,
 };
+
+function useParseChildrenSettings(children, additionalCellComponents = {}) {
```
bryanrsmith: This looks like a standard functions. I don't think it should be named like a hook.

components/grid/tree-grid.jsx
```diff
@@ -0,0 +1,303 @@
...
+	const { gridApi, setGridApi, columnApi, setColumnApi } = useGridState();
+
+	// For styling the drop target rows
+	const previousHoveredRowNode = useRef();
+	const hoveredRowNode = useRef();
+	const draggedNode = useRef();
+	const dragDirection = useRef();
+
+	const heading = React.Children.toArray(children).find(child => child && child.type.isTreeGroup);
+	useEffect(() => {
```
bryanrsmith: I think we could wrap this in a `if (process.env.NODE_ENV !== 'production')`

components/grid/tree-grid.jsx
```diff
@@ -0,0 +1,303 @@
...
+	}, []); // eslint-disable-line react-hooks/exhaustive-deps
+
+	const getShouldShowDropTarget = useCallback(
+		onDirection => ({ rowIndex }) =>
+			hoveredRowNode.current &&
+			hoveredRowNode.current.rowIndex === rowIndex &&
+			draggedNode.current.rowIndex !== hoveredRowNode.current.rowIndex &&
+			dragDirection.current === onDirection &&
+			(!isValidDropTarget ||
+				isValidDropTarget(
```
bryanrsmith: This is hard to read. Consider adding a body and using an `if` statement.

components/grid/tree-grid.jsx
```diff
@@ -0,0 +1,303 @@
...
+
+TreeGrid.rowSelectionOptions = BaseGrid.rowSelectionOptions;
+
+TreeGrid.rowTypes = {
+	folder: 'folder',
+	item: 'item',
+};
+
+TreeGrid.expandedRowsOptions = {
+	// Why ag-grid?
```
bryanrsmith: What does this mean?

components/grid/tree-grid.jsx
```diff
@@ -0,0 +1,303 @@
...
+// depth first for now
+function getTreeWithoutId(id, tree) {
+	const filteredTree = tree.filter(item => item.id !== id);
+	if (filteredTree.length !== tree.length) {
+		return filteredTree;
+	}
+
+	for (const item of filteredTree) {
+		if (item.children && item.children.length) {
+			item.children = getTreeWithoutId(id, item.children);
```
bryanrsmith: Mutating items in this collections seems bad.

components/grid/tree-grid.jsx
```diff
@@ -0,0 +1,303 @@
...
+			return [data, ...tree];
+		}
+		newTree.splice(direction === dragDirections.up ? parentIndex : parentIndex + 1, 0, data);
+
+		return newTree;
+	}
+
+	for (const item of newTree) {
+		if (item.children) {
+			item.children = insertDataIntoTree(data, parentId, direction, item.children);
```
bryanrsmith: Here, too.

components/grid/tree-grid.jsx
```diff
@@ -0,0 +1,303 @@
...
+
+	return newTree;
+}
+
+function getNewPath(parentNode, direction) {
+	const isParentFolder = parentNode.data.treeNodeType === TreeGrid.rowTypes.folder;
+	if (isParentFolder && parentNode.expanded && direction === dragDirections.down) {
+		return [...parentNode.data.path];
+	} else {
+		return [...parentNode.data.path.slice(0, -1)];
```
bryanrsmith: `Array.prototype.slice` creates a shallow copy of the subsection of the array. You don't need to copy it again.

components/grid/tree-grid.jsx
korbinancell: It was a comment to self that slipped through. The different options are weird.

PR #252, opened 2019-10-10
Make deafult ModalContainer padding consistent

AuresaNyctea:
Now that we don't have a bottom margin on the modalheader


bryanrsmith:
Is this what everyone/the spec wants? I assumed the padding specifications came from design, but I didn't confirm.

AuresaNyctea:
There used to be a bottom margin on the header of 12 px which made the spacing look equal all around. now that margin is gone according to @RobertBolender so this is fixing it back to the previous look of having 24px spacing around the whole container.

PR #255, opened 2019-10-11
Add Home/End key handler to Select component

iamthuypham:
- Type a word in Select
- Press Home/End key

Current behavior:  focus on the dropdown items (default by `react-select`)
Expected behavior:  focus on the input, cursor move to begin or end of the input value

### DIFF
--- a/components/text-input-v2/select.jsx
+++ b/components/text-input-v2/select.jsx
@@ -134,6 +134,34 @@ function noOptionsMessage({ inputValue }) {
 	return inputValue ? 'No options' : null;
 }
 
+function handleKeyDown(e) {
+	const evt = { ...e };
+	switch (evt.key) {
+		case 'Home': {
+			evt.preventDefault();
+			if (evt.shiftKey) {
+				evt.target.selectionStart = 0;
+			} else {
+				evt.target.setSelectionRange(0, 0);
+			}
+			break;
+		}
+		case 'End': {
+			evt.preventDefault();
+			const len = evt.target.value.length;
+			if (evt.shiftKey) {
+				evt.target.selectionEnd = len;
+			} else {
+				evt.target.setSelectionRange(len, len);
+			}
+			break;
+		}
+		default: {
+			break;
+		}
+	}
+}
+
 /** Autocomplete control based on react-select */
 export const Select = React.forwardRef(({ components = {}, ...props }, ref) => {
 	const body = useBody();
@@ -145,6 +173,7 @@ export const Select = React.forwardRef(({ components = {}, ...props }, ref) => {
 			theme={selectTheme}
 			components={{ DropdownIndicator, ...defaultComponents, ...components }}
 			noOptionsMessage={noOptionsMessage}
+			onKeyDown={handleKeyDown}
 			menuPortalTarget={body}
 			{...props}
 			styles={selectStyles(props)}
@@ -164,6 +193,7 @@ export const CreatableSelect = React.forwardRef(({ components = {}, ...props }, 
 			formatCreateLabel={node => <span>New entry: {node}</span>}
 			components={{ DropdownIndicator, ...defaultComponents, ...components }}
 			noOptionsMessage={noOptionsMessage}
+			onKeyDown={handleKeyDown}
 			menuPortalTarget={body}
 			{...props}
 			styles={selectStyles(props)}
@@ -184,6 +214,7 @@ export const AsyncCreatableSelect = React.forwardRef(({ components = {}, ...prop
 			components={{ DropdownIndicator, ...defaultComponents, ...components }}
 			formatCreateLabel={node => <span>New entry: {node}</span>}
 			noOptionsMessage={noOptionsMessage}
+			onKeyDown={handleKeyDown}
 			menuPortalTarget={body}
 			{...props}
 			styles={selectStyles(props)}
@@ -202,6 +233,7 @@ export const AsyncSelect = React.forwardRef(({ components = {}, ...props }, ref)
 			theme={selectTheme}
 			components={{ DropdownIndicator, ...defaultComponents, ...components }}
 			noOptionsMessage={noOptionsMessage}
+			onKeyDown={handleKeyDown}
 			menuPortalTarget={body}
 			{...props}
 			styles={selectStyles(props)}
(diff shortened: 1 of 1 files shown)


bryanrsmith:
Has uiux confirmed that this change is desirable for all uses of this component?

If a consumer passes their own `onKeyDown`, the component's interaction behavior will change, which seems undesirable. Should this handler forward the event to an `onKeyDown` prop, if one was provided?

iamthuypham:
@bryanrsmith Make sense. I am not so sure how to achieve that.
Should it be like this?

```
const Select = (props) => {
     const onConsumerKeyDown = props.onKeyDown;
     return (
            <ReactSelect {...props} onKeyDown={e => handleKeyDown(e, onConsumerKeyDown)}/>
     )
}

const handleKeyDown = (e, onConsumerKeyDown) => {
     const evt = {...e};
     switch (evt.key) {
            case 'Home':
                  ...
             case 'End':
                  ...
            default:
                  if (onConsumerKeyDown) {
                       onConsumerKeyDown(e);
                  }
                  break;
     }
}
```

bryanrsmith:
That's one option, yes. This would not allow them to see keydown events when the key was home or end, which also seems a little surprising. 

I suppose it depends on what kind of control you want the consumer to have. You could invoke their handler first and let them return a boolean indicating whether or not they handled the event (which would let a consumer dynamically opt out of the home/end handling). A similar approach would be to check event.defaultPrevented, and use that to let a consumer communicate that they wish to opt out of the default handling. Or you could always invoke their handler and also always handle home and end keys (which would let them handle home/end if they want, but also not allow them to opt out of the default behavior).

iamthuypham:
@bryanrsmith I added the check `event.defaultPrevented`

components/text-input-v2/select.jsx
```diff
@@ -134,9 +134,46 @@ function noOptionsMessage({ inputValue }) {
 	return inputValue ? 'No options' : null;
 }
 
+function handleKeyDown(e, onConsumerKeyDown) {
+	const evt = { ...e };
+	if (evt.defaultPrevented) {
```
bryanrsmith: @iamthuypham Sorry, I don't understand this code. Can you describe why the event would have been prevented at this point? Also, why is the event object copied on line 138?

What I was suggesting was more along the lines of....

```js
if (onConsumerKeyDown) {
    onConsumerKeyDown(e);
    if(e.defaultPrevented) {
        return;
    }
}
```

This would allow the consumer to cancel the default action, informing this handler that it shouldn't run.

components/text-input-v2/select.jsx
```diff
@@ -134,9 +134,46 @@ function noOptionsMessage({ inputValue }) {
...
+		if (onConsumerKeyDown) {
+			onConsumerKeyDown(e);
+		}
+		return;
+	}
+
+	switch (evt.key) {
+		case 'Home': {
+			if (evt.shiftKey) {
+				evt.target.selectionStart = 0;
```
bryanrsmith: Do we have guarantees that the target is an input element? Should we check? What does this code do if the select instances disables search, so there is no input?

bryanrsmith:
Closing due to lack of response or activity.

bryankenote:
@bryanrsmith Would you be willing to reopen this? I would like to keep working on it.

bryanrsmith:
For sure. I'm not opposed to the change; the PR was just getting stale.

bryankenote:
@iamthuypham Can I get permission to push to this?

iamthuypham:
@bryankenote Not sure if your push requires my permission. But if you need to merge, Bryan Smith is authorized to merge.

bryankenote:
@iamthuypham The commits for this pr live on the branch on your remote. To add to this pr I need to either have permission to push to your branch or you can just merge my changes into it if you do not mind (I already created a pr). Otherwise, I can just create a whole new pr for this.

iamthuypham:
@bryankenote Sorry, I forgot this is my branch. https://github.com/iamthuypham/styled-ui/invitations

bryankenote:
@bryanrsmith This should be ready for additional review whenever you get the chance.

components/text-input-v2/select.jsx
```diff
@@ -134,9 +134,45 @@ function noOptionsMessage({ inputValue }) {
...
+			const len = e.target.value.length;
+			if (e.shiftKey) {
+				e.target.selectionEnd = len;
+			} else {
+				e.target.setSelectionRange(len, len);
+			}
+			break;
+		}
+		default: {
+			if (onConsumerKeyDown) {
```
bryanrsmith: Seems like this will double-invoke `onConsumerKeyDown`.

components/text-input-v2/select.jsx
```diff
@@ -134,9 +134,45 @@ function noOptionsMessage({ inputValue }) {
...
+function handleKeyDown(e, onConsumerKeyDown) {
+	const { nodeName, type } = e.target;
+	if (onConsumerKeyDown) {
+		onConsumerKeyDown(e);
+		if (e.defaultPrevented) {
+			return;
+		}
+	}
+	switch (e.key) {
+		case nodeName === 'input' && type === 'text' && 'Home': {
```
bryanrsmith: Using a complex expression for a case label is pretty unusual. Either wrapping the whole switch with a conditional (since both branches test the same condition), or using nested conditionals instead of the switch would be easier to understand.

I don't believe the default case is necessary, so you could probably replace the whole switch with something like,

```js
if (nodeName === 'input' && type === 'text') {
	const selectionIndex = e.key === 'Home' ? 0 : e.key === 'End' ? value.length : null;

	if (selectionIndex !== null) {
		if (e.shiftKey) {
			e.target.selectionEnd = selectionIndex;
		} else {
			e.target.setSelectionRange(selectionIndex, selectionIndex);
		}
	}
}

File: components/text-input-v2/select.jsx [hunk 5]
```diff
@@ -164,6 +201,7 @@ export const CreatableSelect = React.forwardRef(({ components = {}, ...props },
 			formatCreateLabel={node => <span>New entry: {node}</span>}
 			components={{ DropdownIndicator, ...defaultComponents, ...components }}
 			noOptionsMessage={noOptionsMessage}
+			onKeyDown={handleKeyDown}
```
bryanrsmith: `onKeyDown` is going to be overridden by a consumer-specified handler here, because of the positioning of the spread. Don't we want all of these components to handle this event the same as `Select` above?

PR #257, opened 2019-10-14
Seperate export grid component.

korbinancell:


webpack.config.js
```diff
@@ -9,7 +9,7 @@ module.exports = {
 	},
 	entry: {
 		main: './index.js',
-		'ag-grid': './components/grid/index.js',
```
bryanrsmith: This seems like a breaking change. If we want to start calling it 'grid' instead of 'ag-grid', a cleaner approach might be to add files to the root of the repo that re-export from dist. e.g.,

`/grid.js`
```js
module.exports = require('./dist/ag-grid.js');
```

bryanrsmith: (There are no breaking changes to the original ag-grid component; correct?)

korbinancell: The ag-grid components existed in the catalog only. The original grid folder only had the sass file and exported it through the index. Which could be an issue I suppose.

bryanrsmith: @korbinancell Did you see this Q?

bryanrsmith: Ok, we need to avoid breaking changes. Can you restore whatever exports were there before?

bryanrsmith: We can document them as being deprecated, and remove then in 6.0.

korbinancell: I was mistaken the only thing that is missing are the example components that were never exposed to consumers.

PR #260, opened 2019-10-22
Don't lazy load product drawer dropdown.

tobs1701:
Discussed with @dustymack. The lazy-load isn't really saving us much, and it's impeding consumption in brand-bar and Skeletor.

### DIFF
--- a/components/product-drawer/component.jsx
+++ b/components/product-drawer/component.jsx
@@ -1,10 +1,9 @@
-import React, { lazy, Suspense } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
-import { LoadingSpinner } from '../loading-spinner';
 import { PopoverManager, PopoverReference } from '../popover';
 import * as Styled from './styled';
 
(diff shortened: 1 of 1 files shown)


tobs1701:
I fully support finding a solution. For this sprint, I've spent a significant amount of time trying to find one and I'm just trying to get unblocked for the time being.

What if I also lazy-loaded the product drawer in the brand-bar? Use of the drawer is intended to be opt-in; the only site that would be doing so at this point is logos.com.

cc @dustymack

bryanrsmith:
> What if I also lazy-loaded the product drawer in the brand-bar?

Isn't part of the drawer a button that couldn't be lazy loaded without affecting layout?

>  Use of the drawer is intended to be opt-in; the only site that would be doing so at this point is logos.com.

You'll need to avoid statically importing the product drawer into the brand bar if not everyone wants it. You can probably either export a separate brand-bar component that includes the product drawer, or you allow consumers to import the product drawer themselves and pass it to the product drawer at render time.

PR #261, opened 2019-10-22
Grid updates

korbinancell:
Addresses a couple issues in https://github.com/Faithlife/styled-ui/issues/216

* Add drag-n-drop to SimpleGrid
* Add checkbox support to all grids.


catalog/grid/examples.md
```diff
@@ -55,8 +55,8 @@ showSource: true
 ---
 <div>
 	<SimpleGrid data={censusData} maxRows={10}>
-		<GridColumn displayName="Area Description" fieldName="areaDesc" groupByColumn width={200} />
-		<GridColumn displayName="Name" fieldName="value" defaultSort={GridColumn.sortOptions.ascending} />
+		<SimpleGrid.GroupColumn displayName="Name" fieldName="value" />
```
rmjohnson: Why does this one need to be a `SimpleGrid.GroupColumn` instead of just a `GridColumn`?

catalog/grid/examples.md
```diff
@@ -110,3 +110,108 @@ function PopulationChange({ value }) {
...
+	</SimpleGrid>
+</div>
+```
+
+```code
+lang: js
+---
+import { useCellEditor } from '@faithlife/styled-ui/grid';
+
+const IncrementButton = React.forwardRef(({ value }, ref) => {
```
rmjohnson: Can we not just `IncrementButton` from `cell-editors.jsx` ? Maybe I'm just not understanding what this is for.

components/grid/base-grid.jsx
```diff
@@ -148,13 +182,17 @@ export function BaseGrid({
...
 				getRowNodeId={handleGetRowId || getRowNodeId}
 				suppressRowClickSelection={suppressRowClick}
 				onCellClicked={suppressRowClick ? handleCellClicked : null}
 				getRowHeight={getRowHeight}
+				onRowClicked={suppressRowClick ? null : handleRowClicked}
+				onCellEditingStopped={handleCellEdit}
+				suppressContextMenu
+				animateRows
 				reactNext
+				colResizeDefault="shift"
```
rmjohnson: lol

components/grid/base-grid.jsx
```diff
@@ -171,18 +209,25 @@ export function BaseGrid({
...
 						<AgGridColumn
 							{...additionalColumnOptions}
 							{...columnProps}
+							rowDrag={showDragHandle && index === 0}
 							key={fieldName}
 							headerName={displayName}
 							field={fieldName}
 							sortable={isSortable}
 							cellRenderer={cellComponent ? fieldName : null}
+							cellEditor={editorComponent ? `${fieldName}Editor` : null}
```
rmjohnson: nit: should this `Editor` be in shared code? (So if we want to change it we only need to remember to change it in one spot)

components/grid/grid-helpers/aggregation-group-column.jsx
```diff
@@ -2,7 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { GridColumn } from './grid-column';
 
-export function TreeGroupColumn({
+// This is maintained separately from the GridColumn because while there is overlap they do have some differences.
+export function AggregationGroupColumn({
```
rmjohnson: Is this going to be a breaking change for consumers?

components/grid/grid-helpers/grid-helpers.jsx
```diff
@@ -0,0 +1,210 @@
...
+		previousHoveredRowNode,
+		hoveredRowNode,
+		draggedNode,
+		dragDirection,
+		getShouldShowDropTarget,
+	};
+}
+
+export function useCellEditor(ref, cellValue, isPopover = true) {
+	const [showAsPopover] = useState(isPopover);
```
rmjohnson: What's the advantage of using `useState` here if we have no setter? (as opposed to `useRef`)

components/grid/tree-grid.jsx
```diff
@@ -224,30 +170,28 @@ TreeGrid.rowTypes = {
...
 
-TreeGrid.expandedRowsOptions = {
-	all: -1,
-	none: 0,
-	topLevel: 1,
-};
+TreeGrid.expandedRowsOptions = BaseGrid.expandedRowsOptions;
 
-TreeGrid.GroupColumn = TreeGroupColumn;
+TreeGrid.GroupColumn = AggregationGroupColumn;
```
rmjohnson: Ah, I assume we can change this because the old `TreeGroupColumn` was never referenced by consumers (see earlier comment).

catalog/grid/examples.md
korbinancell: Under the hood they are different things with almost the same props (but not exactly) and almost the same shape (but not exactly).

catalog/grid/examples.md
korbinancell: I just added this as a usage example for `useCellEditor`

components/grid/tree-grid.jsx
korbinancell: Yep it is never directly exported, so a name change is fine.

components/grid/grid-helpers/aggregation-group-column.jsx
korbinancell: see below

components/grid/grid-helpers/grid-helpers.jsx
korbinancell: I don't have to type showAsPopover.current. 🤷‍♂ doesn't seem like there is any huge benefit either way.

korbinancell: To me it also communicates more clearly that this shouldn't change or be changed.

catalog/grid/examples.md
rmjohnson: But what's special about this column that it requires being a SimpleGrid.GroupColumn?

korbinancell: It is defining what is being grouped and the props for those rows.

rmjohnson: I was reading that as `SimpleGrid.GridColumn` 🤦‍♂ Carry on :smile:

PR #264, opened 2019-10-24
Name Accordion Wrapper and give it a bottom border

Seth-Klein:


components/accordion/accordion.jsx
```diff
@@ -48,13 +49,13 @@ export function Accordion({
 	);
 
 	return (
-		<Box {...props} onKeyDown={handleKeyboardNav}>
```
bryanrsmith: `Box` accepts theme-aware style props for border. You should be able to keep the existing Box, and use `<Box borderBottom={1} borderColor="gray14"`

Seth-Klein: cool, this is my first time working in this area so information like this is helpful. Along those lines, am I right in assuming that some of the props [here](https://github.com/Faithlife/styled-ui/blob/add415df76de5d21814a61c8063636825f177c97/components/accordion/accordion-header.jsx#L148-L155) are not "theme-aware style props" and that's why it has to be a const and not just a Box? (this was the code I was trying to emulate originally, so I want to understand the difference in their use cases)

components/accordion/accordion.jsx
```diff
@@ -1,5 +1,6 @@
 import React, { useCallback, useMemo, useRef, useState } from 'react';
 import PropTypes from 'prop-types';
+import styled from 'styled-components';
```
AuresaNyctea: I don't think this is necessary anymore

components/accordion/accordion.jsx
bryanrsmith: That code is probably a remnant of a refactor. It could also use `Box`. The following are all equivallent:

```jsx
return <Box borderColor="gray14">Hello</Box>;
```

```jsx
return <MyWrapper>Hello</MyWrapper>

...

const MyWrapper = ({children, ...props}) => <Box {...props} borderColor="gray14">{children}</Box>
```

```jsx
return <MyWrapper>Hello</MyWrapper>

...

const MyWrapper = styled(Box)`
    border-color: ${props => props.theme.colors.gray14};
`;
```


`Box` is essentially just a styled component with hundreds of rules similar to ` border-color: ${props => props.theme.colors.gray14}` above (plus, many more css properties like display, width, padding, margin, color, flex, grid, etc.). The difference is that it uses prop values to look up named theme data, making it simpler to use the shared theme.

```jsx
const Box = styled.div`
    border-color: ${props => props.borderColor ? props.theme.colors[props.borderColor] : null};
`;
```

The code I suggested is functionally identical to the code you originally wrote; it's just more succinct because you can lean on `Box` to look up the theme data for you rather than manually writing the function to pull the desired value out of the theme object. Both forms are valid and have their place; I tend to prefer using `Box` for easier access to theme data. You might use the styled-components API (`styled(Box)`) if you need to use pseudo classes, which `Box` does not provide props for. e.g.,

```jsx
const MyWrapper = styled(Box)`
    border-color: ${props => props.theme.colors.gray14};

    &:hover {
        border-color: ${props => props.theme.colors.gray66};
    }
`
```

Further reading: 
* `styled-components` is the library styling library we're using here. It's what provides the tag functions (like, `styled(Box)`) that processes the CSS string with embedded functions and injects stylesheets into the document. It also provides theme support. https://www.styled-components.com/docs/advanced#theming
* `styled-system` is a set of helpers (basically, all those interpolation functions like `${props => props.theme.colors[props.color];}` that pair with styled-components to provide a more structured approach to leveraging the styled-components theming system. `styled-system` is what we used to build `Box` with all the theme-aware styled props. https://styled-system.com/

PR #275, opened 2019-10-28
Update modals

### LINKED ISSUE #265: [RFD] Update to modals
# Modals

We already have a Modal component, but have received new specs for updated spacing and layout: https://app.zeplin.io/project/58a23c2d4597afb248c13a43/screen/5da8b786136c341856fc8020

* Describe the proposed API changes:
We currently expose a ModalContent component that is a Box with default padding on all four sides. Amber has requested that the vertical padding between the ModalHeader and ModalContent, as well as the vertical padding between the ModalContent and ModalFooter not be arbitrarily configurable, it will either be 24px or 16px depending on the variant chosen.

The updated ModalContent Box will default to 0 vertical padding, and default to horizontal padding that matches the Modal's 24px/16px variant.

RobertBolender:
Spec:

Resolves #265 

Consumers can opt-in to the new v6 Modal API by importing from:
`import { Modal } from @faithlife/styled-ui/v6`

Deprecation warnings have been added to inform users of the new API. The /v6 entrypoint will continue to be supported (with a deprecation warning) throughout the lifetime of v6 and won't be removed until v7 is released. This will ensure plenty of time for a smooth transition.

---

BREAKING CHANGES:

1. ~~ModalContent no longer has `overflow-y: auto` by default. Consumers who want their modal content to scroll should add it themselves. `<ModalContent>` accepts a `overflowY` styled-system prop.~~
This decision has been reversed.

2. ModalHeader no longer has a bottom-border. ~~ModalContent no longer has vertical padding by default.~~

3. ~~Modals that are rendered without a ModalFooter *may* need to specify bottom padding on their ModalContent, unless they don't want it (such as for the fullscreen SmartMediaEditor, for example).~~

4. Removed the unused `tabindex` option from modal footer buttons. (zero active uses at this time)


korbinancell:
While you are poking around could I request adding a boolean isDisabled to the `footerProps` buttons object. Currently you have to supply a custom footer just to disable a button.

RobertBolender:
Looks like the DefaultModalFooter already supports disabled the buttons with a prop, we just don't call it out on the propTypes for the Modal. I'll add it in this PR.

https://github.com/Faithlife/styled-ui/blob/4c3f025e410e7b7a2ed964837a665559bfc8c438/components/modal/default-modal-footer.jsx#L76

catalog/modal/variations.md
```diff
@@ -261,17 +329,17 @@ state: { modal: false, value: '' }
...
-		title="Modal with no title border"
-		styleOverrides={{ bottomBorder: 'none' }}
+		title="Modal with full width content"
+		withoutFooter
 	>
-		<div>
-			This modal has no title border!
-		</div>
+		<ModalContent padding={0} overflowY="auto">
+			<img src="https://www.bellinghamherald.com/news/local/l6de4z/picture53186905/alternates/LANDSCAPE_1140/Faithlife%201" alt="Faithlife campus" style={{ display: 'block' }} />
```
bryanrsmith: I'm entertained that we're hot linking a photo of our own campus from the herald. 😂

components/modal/index.js
```diff
@@ -1,3 +1,24 @@
-export { ModalFooter } from './modal-footer';
-export { ModalContent } from './modal-content';
-export { Modal } from './modal';
+import { deprecate } from '../utils';
+import { ModalContent } from './modal-content';
+import { ModalFooter } from './modal-footer';
+import { LegacyModal } from './legacy-modal';
+
+const v6ImportHelpText = `You can opt into this new API by importing { modal } from @faithlife/styled-ui/v6
```
bryanrsmith: `{ Modal }`?

components/utils/deprecate.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+var hits = {};
```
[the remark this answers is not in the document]
RobertBolender: It does.

components/modal/modal-footer-buttons.jsx
```diff
@@ -0,0 +1,93 @@
...
+					marginRight={useFullWidthButtons ? null : 'auto'}
+				>
+					<Button
+						primaryOutline
+						medium
+						onClick={deleteButton.onClick}
+						defaultColor="red4"
+						hoverColor="red3"
+						activeColor="red5"
+						disabledColor="red1"
```
bryanrsmith: This is fine for now, but I intend to remove these props. They're undocumented and were never intended to be exposed publicly. I'll be adding a danger variant to Button, and we'll need to use theme or the css prop to apply colors outside of the variants.

components/modal/modal-header.jsx
```diff
@@ -5,35 +5,50 @@ import { Text } from '../Text';
...
+			<Box display="flex" justifyContent="space-between" alignItems="center" width="100%">
+				<Text textStyle={textStyle} color="gray66">
+					{title}
+				</Text>
+				<Button variant="primaryTransparent" icon={<Close />} onClick={onClose} />
+			</Box>
+			{subtitle && (
+				<Paragraph
+					width="100%"
+					paddingTop={3}
```
bryanrsmith: Would a Stack be a more appropriate way to accomplish this?

components/modal/modal.jsx
```diff
@@ -1,166 +1,100 @@
...
-		this.cancelResizeListener = cancel;
-		this.setState({ canUseDom: true }); // eslint-disable-line
+/** A flexible component built on styled-system primitives. */
+export const Modal = ({
+	isOpen,
+	container,
+	onClose,
+	children,
+	theme,
+	styleOverrides,
```
bryanrsmith: Is this the v6 API? We should remove this prop, if so.

components/utils/deprecate.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+var hits = new Set();
```
bryanrsmith: Still have that `var` :-)

We should figure out why this is allowed.

components/modal/modal-footer-buttons.jsx
```diff
@@ -0,0 +1,93 @@
...
+			)}
+			{deleteButton && (
+				<Box
+					width={useFullWidthButtons ? '100%' : null}
+					marginBottom={useFullWidthButtons ? 5 : null}
+					marginRight={useFullWidthButtons ? null : 'auto'}
+				>
+					<Button
+						primaryOutline
+						medium
```
bryanrsmith: These are deprecated props. `variant="primaryOutline" size="medium"`

components/modal/modal-footer-buttons.jsx
```diff
@@ -0,0 +1,93 @@
...
+	cancelButton: PropTypes.shape({
+		onClick: PropTypes.func.isRequired,
+		text: PropTypes.string.isRequired,
+		tabindex: PropTypes.string,
+		disabled: PropTypes.bool,
+	}),
+	deleteButton: PropTypes.shape({
+		onClick: PropTypes.func.isRequired,
+		text: PropTypes.string.isRequired,
+		disabled: PropTypes.bool,
```
bryanrsmith: No `tabindex` on this one? Also, did we deliberately name `tabindex` with different casing?

catalog/modal/variations-v6.md
```diff
@@ -0,0 +1,151 @@
...
+<div>
+	<Button variant="primary" size="medium" onClick={() => setState({ modal: !state.modal })}>Open a modal!</Button>
+	<Modal
+		isOpen={state.modal}
+		container="body"
+		onClose={() => setState({ modal: false })}
+		fullscreen
+	>
+		<Modal.Header title="Fullscreen modal" textStyle="h.24" />
+		<Modal.Content overflowY="auto">
```
bryanrsmith: @RobertBolender Would you remind me of why we removed this default?

components/modal/modal-footer-buttons.jsx
RobertBolender: Deprecated props should have warnings 😆

catalog/modal/variations-v6.md
RobertBolender: Having the overflowY not be "visible" will clip the input element focus border if an `<input>` is the first or last child of the `<Modal.Content>`.

My reasoning is that it is *more* common that we have small modals with input elements than that we have modals that will need to scroll. I'm open to being persuaded otherwise.

https://www.flowdock.com/app/logos/styled-ui/threads/aLCffos-m9wRFnpqkJndQOtRiLd

components/modal/modal.jsx
RobertBolender: Looks like we were only using it for zIndex, which is available as a styled-system prop. I'll add a deprecation notice.

components/modal/modal-footer-buttons.jsx
RobertBolender: They should all be `tabIndex` in React, and I see no reason to not include it on the delete button.

bryanrsmith: > They should all be tabIndex in React

Right, so why are we using `tabindex`?

RobertBolender: Looks like you merged the PR: 🙃 
https://github.com/Faithlife/styled-ui/pull/167

RobertBolender: I can deprecate `tabindex` and replace it with `tabIndex`.

catalog/modal/variations-v6.md
RobertBolender: From an offline discussion: we can sidestep that problem if we include vertical padding on the Modal.Content instead of bottom-padding on the Modal.Header and top-padding on the Modal.Footer. I will make that change shortly.

components/utils/deprecate.js
RobertBolender: eslint-config-faithlife has `prefer-const` enabled but not `no-var`
https://eslint.org/docs/rules/no-var#require-let-or-const-instead-of-var-no-var

components/modal/modal.jsx
RobertBolender: Oh, we had been passing the zIndex style-override to the *backdrop*, which we are otherwise not exposing to the consumer.
How would you feel about adding an explicit `zIndex` prop to the Modal that would be passed to the Backdrop instead? I can't think of a scenario where you would want to set the zIndex on the Modal itself in a way that wouldn't affect the Backdrop.

components/modal/modal-footer-buttons.jsx
RobertBolender: Do we intend to remove those deprecated props in v6? I said so in these warnings:
https://github.com/Faithlife/styled-ui/pull/275/files#diff-5c88c3414a1e482bfebb4f8c7fa080baR46

components/modal/modal-footer-buttons.jsx
RobertBolender: Also, there are *zero* active uses of this. Should I just remove it?

RobertBolender: Confirmed with Thuy that these are unused.

components/modal/modal-footer-buttons.jsx
bryanrsmith: Yep, and I'm actually nearly done with a button rewrite; If you don't mind leaving the Button props as-is, I'll add the warnings using your new utility before I push it.

components/modal/modal.jsx
```diff
@@ -1,166 +1,103 @@
...
+	onClose,
+	children,
+	theme,
+	contentPadding,
+	fullscreen,
+	zIndex,
+	...props
+}) => {
+	const [size, containerRef] = useElementSize();
+	const [canUseDom, setCanUseDom] = useState(false);
```
bryanrsmith: What's the purpose of this flag? It seems like it's just used to avoid rendering before the component mounts, but I'm not sure why.

RobertBolender: Correct, it does prevent rendering the modal before the component mounts. I'm not 100% certain of the reason for it, other than the fact that we might be querying the document for a container if a containerID is supplied.

It was added when we added React.createPortal:
https://github.com/Faithlife/styled-ui/commit/e2a3db25011e82fb85fbd44798f1f9aaa586b37c

I'll check if we could remove it.

RobertBolender: @EthanRutherford reports that it is necessary for SSR, React's SSR does not support hydrating Portals:
https://github.com/facebook/react/issues/13097

bryanrsmith: No objection if it's necessary, but I find the current code a bit confusing. It might be cleaner to put the target element in state, and set that in the effect instead of using a ref.

RobertBolender: Sure, I can setState in a useEffect instead of using a Ref.

bryanrsmith: Makes sense. Maybe just a comment, then?

Issue #280, opened 2019-10-31
Override toggle text styles in Product Drawer.


tobs1701:
Recent consumption of the ProductDrawer as part of https://faithlife.atlassian.net/browse/SITES-5126 has revealed the need to override the toggle text styling, specifically:

- text font
- text size
- text color

The goal here is to allow more consistent styling between the Drawer toggle and its consumer. The following doc provides more details: https://docs.google.com/document/d/1fUT08im_Q866O4WC-gCwD8Nw4NoGsMcODKiIGXCFm6s/edit

RobertBolender:
It appears that this issue was resolved by PR #281. If that is not the case, please feel free to re-open this issue.

PR #281, opened 2019-10-31
Add additional style overrides to Product Drawer. SITES-5126

tobs1701:
Also, hide toggle text at smaller viewport width.

Addresses https://github.com/Faithlife/styled-ui/issues/280.

### DIFF
--- a/components/product-drawer/styled.jsx
+++ b/components/product-drawer/styled.jsx
@@ -42,4 +42,5 @@ export const ProductDrawerToggleText = styled.span`
 		font-family: ${({ styleOverrides }) => styleOverrides.toggleTextFontFamily || 'inherit'};
 		font-size: ${({ styleOverrides }) => styleOverrides.toggleTextFontSize || 'inherit'};
 		color: ${({ styleOverrides }) => styleOverrides.toggleTextColor || 'inherit'};
+	}
 `;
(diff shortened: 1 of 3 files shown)


components/product-drawer/styled.jsx
```diff
@@ -36,8 +36,11 @@ export const ProductDrawerToggle = styled.button`
 export const ProductDrawerToggleText = styled.span`
 	display: none;
 
-	@media (min-width: ${mediaSizes.tablet}) {
+	@media (min-width: ${mediaSizes.phone}) {
 		display: inline;
 		margin-left: 8px;
+		font-family: ${({ styleOverrides }) => styleOverrides.toggleTextFontFamily || 'inherit'};
+		font-size: ${({ styleOverrides }) => styleOverrides.toggleTextFontSize || 'inherit'};
+		color: ${({ styleOverrides }) => styleOverrides.toggleTextColor || 'inherit'};
```
bryanrsmith: You only want to apply the style overrides at larger viewports? Why? That seems surprising.

tobs1701: Smaller viewports are `display: none`.

bryanrsmith: That's a decision that could easily change. I'd suggestion limiting the scope of the media queries to the properties that actually need to change. A style like color wouldn't likely ever be affected by the viewport width, so putting it in a media query makes the code harder to understand.

tobs1701: https://github.com/Faithlife/styled-ui/pull/281/commits/05eee6274ce124a3485f5408761ed0dc6f53f5dd

PR #283, opened 2019-11-05
Update placeholder text color

iamthuypham:


iamthuypham:
Waiting for Amber to double check and approve the change.

components/shared-styles/index.js
```diff
@@ -95,6 +95,7 @@ export const inputColors = {
 	inputBorderColor: colors.borderColor,
 	inputFocusedBorderColor: '#278ed4',
 	inputFocusedShadowColor: '#d0e6f6',
+	inputPlaceholderColor: '#C7C7C7',
```
bryanrsmith: We shouldn't duplicate this in two theming systems. Shared styles will eventually be removed, so it'd be great to move away from it and use theme instead.

iamthuypham:
We approved the #c7c7c7 here : https://faithlife.com/messages/24703?threadId=801163

iamthuypham:
@bryanrsmith I removed the `inputColors` in select.jsx

PR #287, opened 2019-11-08
Update input disabled styles.

korbinancell:


components/input/Input.jsx
```diff
@@ -133,7 +133,9 @@ ${({ theme, styleOverrides = {} }) => css`
 	}
 
 	&:disabled {
-		opacity: 0.5;
+		background-color: ${theme.colors.gray4};
+		color: ${theme.colors.gray34};
+		border-color: ${theme.colors.gray14};
```
RobertBolender: I need to refresh my memory on the difference between these two lines (is there one?):

https://github.com/Faithlife/styled-ui/pull/287/files#diff-c3bfa34cc5484bc90836a4fe35b6ce57R127

https://github.com/Faithlife/styled-ui/pull/287/files#diff-c3bfa34cc5484bc90836a4fe35b6ce57R138

RobertBolender: Actually, I think I remember the difference:

`border-color: ${theme.colors.gray14};` will always use the styled-ui theme as imported up on line 6.

`border-color: ${({ theme }) => theme.colors.gray14};` would allow some potential consumer to supply their own theme with their own value for `gray14` if they so desired for some reason.

bryanrsmith: Yes, we should not be directly importing theme. If you want to reduce syntax, wrap the whole css string in a function. 

```jsx
styled.input(({theme}) => css`
    color: ${theme.colors.gray14};
`)

Comment (korbinancell) [reply components/input/Input.jsx]: @bryanrsmith would it make sense to go ahead and add an `input` property to the theme same as the `button` one? It would definitely be nice for giving.

Comment (bryanrsmith) [reply components/input/Input.jsx]: Yep 👍

File: components/input/Input.jsx [hunk 2]
```diff
@@ -121,69 +121,71 @@ ${({ theme, styleOverrides = {} }) => css`
...
 	}
 
 	&:read-only {
-		background: ${({ theme }) => theme.colors.gray8};
+		background: ${theme.colors.input.readOnlyBackgroundColor};
 	}
 
 	&::placeholder {
-		color: ${({ theme }) => theme.colors.inputPlaceholderColor};
+		color: ${theme.colors.inputPlaceholderColor || theme.colors.input.placeholderColor};
```
bryanrsmith: Why are we preferring the old location here?

korbinancell: So it would be non breaking if someone is overriding theme colors already

theme/core.js
```diff
@@ -91,10 +91,10 @@ const colors = {
 	yellow5: '#b58300',
 
 	borderColor: '#C7C7C7',
-	inputBorderColor: '#C7C7C7',
+	inputBorderColor: '#a8a8a8',
 	inputFocusedBorderColor: '#278ed4',
 	inputFocusedShadowColor: '#d0e6f6',
-	inputPlaceholderColor: '#C7C7C7',
+	inputPlaceholderColor: '#a8a8a8',
```
RobertBolender: Placeholder text should remain `#C7C7C7`

korbinancell: I remember why we didn't merge this initially now. Do you know if Amber ever got around to updating the spec?

RobertBolender: Not yet, I just asked her for an update though.

components/input/Input.jsx
bryanrsmith: But because `theme.colors.inputPlaceholderColor` still exists, we're never going to fall through this condition, right?

korbinancell: Unless a custom theme doesn't add it, yea. Do we have a pattern for deprecating theme properties? I can pr to anyone using it I spose.

korbinancell: @bryanrsmith is it fine to just remove the old theme members?

bryanrsmith: In this case I think it's fine. You'll need to update the other uses of it within styled-ui.

components/input/Input.jsx
```diff
@@ -121,69 +121,72 @@ ${({ theme, styleOverrides = {} }) => css`
...
+		&::placeholder {
+			color: ${theme.colors.input.disabledPlaceholderColor};
+			font-style: italic;
+		}
+	}
+
+	${createVariant({
+		variants: {
+			small: {
+				padding: '8px',
```
bryanrsmith: styled-system now supports theme keys in variants; we should prefer to use them over hard coding and statically referencing the theme.

PR #288, opened 2019-11-08
Add svg export

korbinancell:
This currently breaks anything that uses icons...
If we decide we want to go this route then it should be easy enough to fix.

The icons folder needs some clean up and possibly replacement by actual svgs.


RobertBolender:
> This currently breaks anything that uses icons...

Internal uses of icons, or external uses of icons? Broken in what way?

korbinancell:
Internal.
I removed the existing url-loader for svgs (which might not be correct) so stuff like product drawer is sad.
image

.svgo.yml
```diff
@@ -0,0 +1,51 @@
...
+  - removeUselessDefs: true
+  - removeUselessStrokeAndFill: true
+  - removeViewBox: false
+  - removeXMLNS: false
+  - removeXMLProcInst: true
+  - sortAttrs: true
+  - transformsWithOnePath: true
+  - prefixIds: 'styled-ui'
+
+js2svg:
```
bryanrsmith: What's the purpose of these options?

.svgo.yml
```diff
@@ -0,0 +1,51 @@
...
+  - removeUnknownsAndDefaults: true
+  - removeUnusedNS: true
+  - removeUselessDefs: true
+  - removeUselessStrokeAndFill: true
+  - removeViewBox: false
+  - removeXMLNS: false
+  - removeXMLProcInst: true
+  - sortAttrs: true
+  - transformsWithOnePath: true
+  - prefixIds: 'styled-ui'
```
bryanrsmith: Do we need something unique to the file in there to avoid generated IDs conflicting between svg files? e.g., `#mask1`

catalog.config.js
```diff
@@ -0,0 +1,14 @@
...
+	{
+		test: /\.svg$/,
+		use: ['@svgr/webpack'],
+	},
+];
+
+module.exports = {
+	webpack: catalogWebpackConfig => {
+		const modifiedConfig = { ...catalogWebpackConfig };
+		modifiedConfig.module.rules[0].oneOf = [...newRules, ...modifiedConfig.module.rules[0].oneOf];
```
bryanrsmith: Seems like you could unshift more succinctly.

bryanrsmith:
I think before we release this we should delete all (most of?) our inlined icons and pull in the whole icons folder from Google Drive. That way we can be reasonably sure our icon set is complete, and have a clear path forward when icons get updated.

.svgo.yml
korbinancell: They get their own unique id, this is just a prefix to ensure there are no collisions with others (like fl.com)

RobertBolender:
Anything blocking this PR from moving forward? 😅

PR #289, opened 2019-11-13
Accept styled-system props in Tabs, add Modal example to FilePicker component.

RobertBolender:


catalog/file-picker/variations.md
```diff
@@ -33,39 +35,127 @@ state: {
 	>
 		<TabManager>
 			<TabList>
-				<Tab>
+				<Tab paddingX={4}>
```
korbinancell: It feels really strange to have to include these styles on every single usage of the component to stay consistent. Maybe we should re-export the component or something similar.

RobertBolender: I intend to write a shared helper component for the AmberPickerModal that hides these particular styles. I like the idea of having the Tabs live in Styled-UI as the generic customizable component, and having the most specific AmberPickerModal in the shared repo with a simpler API surface.

PR #290, opened 2019-11-13
Use v6 Modal, CSS Grid in Share Dialog.

RobertBolender:


RobertBolender:
This fixes a regression with the v6 Modal update.
Amber will follow-up with a new design for the Share Dialog soon:
https://faithlife.atlassian.net/browse/IXD-1509

PR #293, opened 2019-11-15
Fix hide and viewport dependent columns.

korbinancell:


components/grid/base-grid.jsx
```diff
@@ -180,6 +171,7 @@ export function BaseGrid({
 						hide,
 						isLargeViewportOnly,
 						isSmallViewportOnly,
+						hasInteractableElement,
```
rmjohnson: Why do we need to pull this out?

korbinancell: If you pass a prop to the columns that it doesn't recognize it will complain about it in the console, even in production. I saw it and popped it in while I was in the area.

PR #300, opened 2019-12-02
Mobile tooltip

### LINKED ISSUE #239: Inferred input tooltips don't work on mobile (by current design)
It's conditionally rendered [here](https://github.com/Faithlife/styled-ui/blob/1bf2d2486dcf943e5fee8b90bb87039b80017ff2/components/popover/tooltip.jsx#L31-L47). Seems like we could use that same `isMobile` logic to update the click handler on mobile to show the tooltip on the first click and confirm on the second click.

bryankenote:
fixes #239

### DIFF
--- a/components/popover/tooltip.jsx
+++ b/components/popover/tooltip.jsx
@@ -57,7 +57,7 @@ export function Tooltip(props) {
 			>
 				{children}
 			</PopoverReference>
-			<Popover {...otherProps} isOpen={tooltipIsOpen || isOpen}>
+			<Popover {...otherProps} isOpen={isOpen === undefined ? tooltipIsOpen : isOpen}>
 				{content || text}
 			</Popover>
 		</PopoverManager>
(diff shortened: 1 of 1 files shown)


bryanrsmith:
As far as I know tooltips were deliberately hidden on mobile at Amber's request. I wasn't directly involved, but I believe the thought process was that tooltips are never a good experience on mobile, and critical content should not be hidden in tooltips anyway.

bryankenote:
That makes sense. This was created to resolve a ChMS case, but we may need to reconsider using a tooltip specifically on mobile for this scenario. @sportnak Did we ever get a spec for this? This is for CHMS-66 in Jira.

sportnak:
So the inferred inputs have a light bulb inside of them - there's no real way to provide that content another way. @bryankenote does tapping the light bulb allow the option to be selected? If it doesn't, that should be the first step (which might not be an issue in styled-ui). If it does then we can move forward with that until we get another report requesting the tooltip show up on mobile.

bryankenote:
@sportnak Tapping the lightbulb on mobile does set the gender to selected and no longer just inferred. There is no strong indication that this will happen, or has happened after, but it works as you have described.

sportnak:
Sounds good. If we get feedback on the lightbulb or setting those values, we can move forward from there.

PR #302, opened 2019-12-04
Import resize-observer-polyfill to prevent crash on mobile

mdombrowski1997:


bryanrsmith:
This polyfill should be provided by the application.

sportnak:
@bryanrsmith can we get some more context on why the polyfill should be provided by the application? That seems somewhat contrary to the nature of this project. Including it here would mean that all the consuming projects wouldn't have to import this themselves. Shouldn't this project work on all browsers?

kbsletten:
This seems a little backward to me, @bryanrsmith. Isn't ResizeObserver a nonstandard API? Even modern browsers don't all support this feature and we're going to require every user of this package to polyfill in order to support most non-chrome ones? This seems a bit like pre-module web development to me. Even from an ergonomic perspective, Babel won't polyfill this and there's no advertisement that it's likely to crash on the latest version of popular browsers.

https://github.com/babel/babel/issues/7328

bryanrsmith:
> Isn't ResizeObserver a nonstandard API? Even modern browsers don't all support this feature and we're going to require every user of this package to polyfill in order to support most non-chrome ones?

No, it's a standard API supported by most modern browsers. https://caniuse.com/#feat=resizeobserver

The main reason for keeping it in the application is to keep bundle sizes smaller. Not all consumers of this code need to support the browsers lacking an implementation, and we don't want to bloat the bundle with code that will never run in certain apps. It also avoids issues with duplicate polyfills in the application. If every library using ResizeObserver bundled their own polyfill, the end application would ultimately ship several implementations of the same API (and still have to add their own in order to safely use it in application code).

I agree that communication on polyfill requirements is lacking. I'll make sure that's improved.

sportnak:
Gotcha, I didn't realize that it was a standard API. Thanks for helping to improve communications around the polyfill. I think it definitely makes sense to avoid having multiple packages that include the polyfill. :+1:

PR #308, opened 2020-01-08
Demonstrate border-box issue in Switch component.

RobertBolender:
Demo: https://deploy-preview-308--faithlife-styled-ui.netlify.com/#/Switch/variations

This PR demonstrates how the Switch component currently renders when the consuming environment has a global CSS style:

```css
*, :before, :after {
  box-sizing: border-box;
}
```

(such as on faithlife.com)

Here is the PR to fix it: https://github.com/Faithlife/styled-ui/pull/309

PR #309, opened 2020-01-08
Set box-sizing on Switch :after element.

RobertBolender:
This prevents an issue when the Switch component in an environment that
globally sets:

*, :before, *:after {
	box-sizing: border-box;
}

... such as faithlife.com

A demonstration of the issue is available here:
https://deploy-preview-308--faithlife-styled-ui.netlify.com/#/Switch/variations
https://github.com/Faithlife/styled-ui/pull/308


bryanrsmith:
💯

PR #311, opened 2020-01-09
Make UtilityButton inherit font-family.

mitchknife:
- This makes it so consumers don't have to set this for every app in order to get their main font to show up by default.


bryanrsmith:
Thanks!

PR #316, opened 2020-02-04
Fix dependency infinite loop.

sportnak:


RobertBolender:
Unfortunately, the build still thinks that `inboundElements` should be in the dependency array, but `inboundElements` will always be a fresh object on every call to `useBasicMap`.

We could destructure the assignment as we did the last time this bug bit us:

https://github.com/Faithlife/styled-ui/commit/ef4e1a221d09d014130af90c72453f5180c81658#diff-fca3f0fae182bc508fcc8192a9f137ceR117

PR #318, opened 2020-02-19
Close DatePickerInput after selecting a date. GIVE-1759

rmjohnson:


rmjohnson:
@bryanrsmith, can you review? I'm not sure what the process is like for "behavior breaking changes" like this one but it *feels* like a change everyone would want though.

rmjohnson:
@bryanrsmith, wanted to make sure this didn't fall off your radar.

PR #319, opened 2020-03-02
Allow conditionally hiding the circle on sequenced tabs.

sportnak:
I need the styling form sequenced tabs, but don't need the numbers or check marks. The non-sequenced tabs have a totally different styling. So the options are 1. allow the other tabs to have the styling changed to look like this one or 2. allow hiding the circle from the sequenced ones.

https://app.zeplin.io/project/5d3b6ca65fc7e30e1d32a69c/screen/5e264cdc037cc36bc48c9b61

### DIFF
--- a/components/tabs/tab.jsx
+++ b/components/tabs/tab.jsx
@@ -56,6 +56,7 @@ export function SequencedTab({
 	selected,
 	onSelectTab,
 	panelId,
+	hideCircle,
 	...props
 }) {
 	const tabRef = useRef();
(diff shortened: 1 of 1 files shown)


RobertBolender:
From the comments on that Zeplin spec, I believe the Sequenced Tab designs were not yet final.
I see this slightly-more-recent spec from Amber, and I intend to follow up with her:

https://app.zeplin.io/project/58a23c2d4597afb248c13a43/screen/5e2a33fa1d43768074ddf8e4

bryanrsmith:
Sounds good Robert. If we want to be able to hide the indicator, let's take an optional `renderIndicator` prop that can return null (or whatever react node it wants).

sportnak:
Sounds good. Looking to ship some code around this this week so I'll just keep an eye on this thread.

RobertBolender:
I met with Amber today and confirmed:

1. That the Responses page is using outdated styles for tabs.
2. That we have a new spec for Sequenced Tabs.
3. That we will soon have a new spec for Page Tabs, which will be used on this Responses page. That spec is expected within the next day or two, and I will keep you posted. It will be available as a style variant on the Styled-UI Tabs component, rather than Sequenced Tabs.

RobertBolender:
The `page` tab variant will be available once this PR is merged: https://github.com/Faithlife/styled-ui/pull/322

PR #320, opened 2020-03-05
Adjust Slider thumb positioning.

RobertBolender:
Matches thumb positioning from before v5.46.1.

### DIFF
--- a/components/slider/component.jsx
+++ b/components/slider/component.jsx
@@ -342,6 +342,8 @@ export class Slider extends PureComponent {
 					<Thumb
 						isSliding={isSliding}
 						isHovered={isHovered}
+						isAtTrackStart={activeStopIndex === 0}
+						isAtTrackEnd={activeStopIndex === stopCount - 1}
 						position={getPercentage(activeStopIndex, stopCount - 1)}
 						label={labels[activeStopIndex]}
 						disabled={disabled}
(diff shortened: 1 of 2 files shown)


components/slider/component.jsx
```diff
@@ -371,24 +373,30 @@ function getPercentage(index, max) {
...
-});
+const Thumb = React.memo(
+	({ isSliding, isHovered, isAtTrackStart, isAtTrackEnd, position, label, disabled }) => {
+		const thumb = <Styled.Thumb active={isSliding} hovered={isHovered} disabled={disabled} />;
+
+		const isPopupOpen = !!(isHovered && (label || label === 0));
+		return (
+			<Styled.ThumbAnchor
+				style={{
+					left: isAtTrackStart ? '7px' : isAtTrackEnd ? 'calc(100% - 6.5px)' : `${position}%`,
```
bryanrsmith: Why 6.5? `right: 7px` seems simpler.

RobertBolender: 6.5 restores the slider to be pixel-perfect to how it was before. There might be a *cleaner* way to accomplish the same result, but that's what I was able to find yesterday.

RobertBolender: Ah, `right: 7px` with `position: absolute` does do what we want. I had been missing that it was currently `position: relative`. I'll use that.

PR #322, opened 2020-03-10
Tabs

korbinancell:
https://app.zeplin.io/project/58a23c2d4597afb248c13a43/screen/5dd32ad3c3f03caf26992655
https://app.zeplin.io/project/58a23c2d4597afb248c13a43/screen/5e60381982a344183b17b9a8

Implement tabs that allow variations. This does not include overflow handling.


theme/tabs.js
```diff
@@ -0,0 +1,117 @@
...
+		},
+	},
+	page: {
+		'&>*:not(:last-child)': {
+			marginRight: 2,
+		},
+	},
+};
+
+const selected = {
```
korbinancell: This seems kinda strange. I am open to different ways of handling the styling between selected and not.

components/tabs/component.jsx
```diff
@@ -1,46 +1,158 @@
...
 		if (selectedTab !== null && selectedTab !== undefined) {
-			setSelectedTabIndex(selectedTab);
+			setCurrentTab(selectedTab);
 		}
 	}, [selectedTab]);
 
-	const handleSelectTab = useCallback(
-		tabIndex => {
+	const onSelectTab = useCallback(
+		newTabIndex => () => {
```
RobertBolender: Is the extra `=> ()` here a mistake?

RobertBolender:
Would it be appropriate to use the `deprecateComponent` util function on the legacy tabs?

components/tabs/component.jsx
```diff
@@ -1,46 +1,158 @@
...
 	useEffect(() => {
 		if (selectedTab !== null && selectedTab !== undefined) {
-			setSelectedTabIndex(selectedTab);
+			setCurrentTab(selectedTab);
 		}
 	}, [selectedTab]);
 
-	const handleSelectTab = useCallback(
-		tabIndex => {
+	const onSelectTab = useCallback(
```
RobertBolender: FYI: the prior naming was based on a Faithlife convention, at least in use by the FLCOM team.
We use the `on...` prefix for props that expect a handler function, and the `handle...` prefix for the function that gets passed in to that prop.

This is not a hard objection.

korbinancell: 🤦‍♂. You are correct

components/tabs/component.jsx
korbinancell: It's so later on we can pass just a callback to the tabs through currying so they don't even have to know their index.  https://github.com/Faithlife/styled-ui/pull/322/files/6aa29ea540564a202eda5304bdd79a96aee27864#diff-a9fab4bd077f8483a828c01aed7bc7aaR119

theme/tabs.js
```diff
@@ -0,0 +1,117 @@
...
+	'modal-false': {},
+	'modal-true': {
+		'&::before': {
+			content: '""',
+			position: 'absolute',
+			top: 0,
+			left: 0,
+			width: 'calc(100% - 2px)',
```
RobertBolender: This calculated width appears to be relying on `box-sizing: content-box`, but that box-sizing results in this :before element being 6px tall, which is too tall. I recommend using `box-sizing: border-box` and adjusting this to `width: 100%`.

RobertBolender: Reminder: Faithlife.com has a global `*, *:before, *:after { box-sizing: border-box; }` rule.
Maybe we should consider adding that to Styled-UI?

components/tabs/component.jsx
```diff
@@ -1,46 +1,158 @@
...
+TabManager.defaultProps = {
+	variant: 'modal',
 };
+
+const Tab = React.forwardRef(({ ...props }, ref) => {
+	const id = useId();
+	return <TabCore id={`tab:${id}`} ref={ref} {...props} />;
+});
+
+Tab.displayName = 'Tab';
```
bryanrsmith: Pass a named function to forwardRef instead.

theme/core.js
```diff
@@ -125,6 +125,31 @@ colors.button = {
 	focusShadow: colors.blue3,
 };
 
+colors.tab = {
```
bryanrsmith: 👍

theme/tabs.js
```diff
@@ -0,0 +1,118 @@
...
+			borderRadius: '3px 3px 0 0',
+		},
+		backgroundColor: 'tab.modalBackgroundActive',
+		borderLeft: 1,
+		borderRight: 1,
+		borderLeftColor: 'tab.modalBorder',
+		borderRightColor: 'tab.modalBorder',
+		borderBottom: 1,
+		borderBottomColor: 'tab.modalBackgroundActive',
+		marginBottom: '-1px',
```
RobertBolender: Super nit: consider adding this `marginBottom: -1px` to non-selected tabs as well.
I understand that this negative marginBottom was intended so that the Tab would hide the border from the Tab.Panel behind it.
On my laptop monitor, at 100% browser zoom and 125% monitor scaling, I can see 1px of white background between the non-selected Tabs and the Tab.Panels below.

image

sportnak:
@korbinancell  @bryanrsmith any objection to merging this? It's needed for custom forms.

korbinancell:
Skipping the api changes and just adding the variants in https://github.com/Faithlife/styled-ui/pull/323

PR #324, opened 2020-04-14
Use wrapper button in CalendarDate.

PatrickNausha:
This PR fixes a bug where clicking any date in a date picker immediately closes the picker.

## Steps to reproduce
1. Go to https://faithlife.github.io/styled-ui/#/date-picker/variations
2. Under "Default Date Range Picker," click "Select Dates."
3. Click a day in the miniature calendar.
    * Expected: The picker should remain open so I can choose an end date.
    * Actual: The picker immediately closes.

## Problem and solution
There's a little bit of an anti-pattern here where we're styling an HTML `button` element by swapping out different components that render it. This leads to [DOM node destruction](https://reactjs.org/docs/reconciliation.html#elements-of-different-types) on button clicks. In our case, we're replacing a `CalendarWeekDay` with a `CalendarWeekDaySelected` on click. Soon afterward, [we check to see if the element clicked is a child of the popup](https://github.com/Faithlife/styled-ui/blob/85b271bd0f1d23e92f757c49fe543f46d154174a/components/shared-hooks/use-focus-away-handler.js#L18) to decide whether we should close the popup. That logic is written with the assumption that that the element that received focus is still in the DOM.

This is also not a preferred pattern for accessibility reasons: Tab key navigation does not work as expected when you're removing the focused element on "click."

To solve the above problems, this PR uses a very plain-looking button and instead swaps out its children to make the button appear how it should.

### DIFF
--- a/components/date-picker/calendar-date/component.jsx
+++ b/components/date-picker/calendar-date/component.jsx
@@ -25,16 +25,18 @@ export class CalendarDate extends Component {
 		this.props.dateFunctions.getMonth(date1) === this.props.dateFunctions.getMonth(date2) &&
 		this.props.dateFunctions.getYear(date1) === this.props.dateFunctions.getYear(date2);
 
-	renderCalendarWeekDay = (selectedDate, selectedDateRange, date, isCurrentMonth) => {
-		let CalendarWeekday = Styled.CalendarWeekDay;
-		const { isValid, isBefore } = this.props.dateFunctions;
+	getDayComponent = () => {
+		const isCurrentMonth = this.isCurrentMonth();
+		const { date, dateFunctions, selectedDate, selectedDateRange, validate } = this.props;
+		const { isValid, isBefore } = dateFunctions;
 
+		let CalendarWeekday = Styled.CalendarWeekDay;
 		if (selectedDate) {
 			if (selectedDate && isValid(selectedDate) && this.areDatesEqual(selectedDate, date)) {
 				CalendarWeekday = Styled.CalendarWeekDaySelected;
 			}
 
-			if (!isCurrentMonth || (this.props.validate && !this.props.validate(date))) {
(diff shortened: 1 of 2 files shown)


components/date-picker/calendar-date/component.jsx
```diff
@@ -53,43 +55,43 @@ export class CalendarDate extends Component {
...
 
 		if (this.areDatesEqual(currentDate, date)) {
 			currentDayDot = <Styled.CalendarWeekDayCurrentDay />;
 		}
 
 		return (
-			<CalendarWeekDay
+			<Styled.CalendarWeekDayButton
 				onClick={this.setSelectedDate}
-				tabIndex="-1"
```
PatrickNausha, line 85: My assumption here was that `tabIndex="-1"` was a hack to make `div`s pretend they were buttons. I don't see any reason not to let someone tab through the days in the calendar if they'd like to.

components/date-picker/calendar-date/component.jsx
```diff
@@ -25,16 +25,18 @@ export class CalendarDate extends Component {
 		this.props.dateFunctions.getMonth(date1) === this.props.dateFunctions.getMonth(date2) &&
 		this.props.dateFunctions.getYear(date1) === this.props.dateFunctions.getYear(date2);
 
-	renderCalendarWeekDay = (selectedDate, selectedDateRange, date, isCurrentMonth) => {
-		let CalendarWeekday = Styled.CalendarWeekDay;
-		const { isValid, isBefore } = this.props.dateFunctions;
+	getDayComponent = () => {
```
PatrickNausha, line 28: I didn't see an obvious reason why `renderCalendarWeekDay` both took arguments and read from props. I assumed that decision was not intentional and made it only read from props for simplicity.

PR #329, opened 2020-04-20
Use relative values for the check indicator

JeremyEinfeld:
The absolute positioning of the indicator resulted in misalignment under some browsers and conditions:

image

The use of flex and padding allows the size of the check indicator to consistently fill the space within the radio button.


JeremyEinfeld:
It seems that this doesn't resolve this issue.

PR #332, opened 2020-04-22
Made dropdown styleable to support FLTV user interface design

travisbader:
It didn't seem like there was a need to do much more than what was done here to get the styling supported.  If you want me to get deeper into adding properties with styled-system lmk.

### DIFF
--- a/components/check-box/component.jsx
+++ b/components/check-box/component.jsx
@@ -6,7 +6,7 @@ import { isSystemTheme } from '../../theme';
 import { DefaultThemeProvider } from '../DefaultThemeProvider';
 
 /** Styled checkbox control with consistent styling across platforms */
-const Checkbox = React.forwardRef(function Checkbox(props, ref) {
+const Checkbox = React.forwardRef(function Checkbox(props) {
 	const {
 		onClick,
 		onMouseUp,
@@ -22,7 +22,7 @@ const Checkbox = React.forwardRef(function Checkbox(props, ref) {
 	} = props;
 
 	const isCoreTheme = isSystemTheme(theme);
-	const themedProps = { ...checkboxProps, ...(isCoreTheme ? {} : theme) };
+	const themedProps = { ...(isCoreTheme ? {} : theme) };
 
 	/* eslint-disable react/prop-types */
 	const handleMouseUp = e => {
--- a/components/check-box/component.jsx
+++ b/components/check-box/component.jsx
@@ -1,6 +1,5 @@
-import React, { Component } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
-import { colors as sharedColors } from '../shared-styles';
 import { CheckboxContent } from './checkbox-content';
 import * as Styled from './styled';
 import { isSystemTheme } from '../../theme';
--- a/components/check-box/component.jsx
+++ b/components/check-box/component.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useRef } from 'react';
 import PropTypes from 'prop-types';
 import { CheckboxContent } from './checkbox-content';
 import * as Styled from './styled';
@@ -31,14 +31,16 @@ const Checkbox = React.forwardRef(function Checkbox(props) {
 		}
 
 		if (!props.disableAutoBlur) {
-			blur();
+			componentRef.current.blur();
 		}
 	};
+
+	const componentRef = useRef()
 
 	return (
 		<DefaultThemeProvider>
 			<Styled.CheckboxContainer
-				ref={this}
+				ref={componentRef}
 				onClick={onClick}
 				onMouseUp={handleMouseUp}
 				type={type}
--- a/components/check-box/component.jsx
+++ b/components/check-box/component.jsx
@@ -3,71 +3,89 @@ import PropTypes from 'prop-types';
 import { colors as sharedColors } from '../shared-styles';
 import { CheckboxContent } from './checkbox-content';
 import * as Styled from './styled';
+import { isSystemTheme } from '../../theme';
+import { DefaultThemeProvider } from '../DefaultThemeProvider';
 
 /** Styled checkbox control with consistent styling across platforms */
-export class Checkbox extends Component {
-	static propTypes = {
-		/** Handler passed to native `button` */
-		onClick: PropTypes.func.isRequired,
-		onMouseUp: PropTypes.func,
-		title: PropTypes.string,
-		isChecked: PropTypes.oneOf([true, false, 'mixed']),
-		theme: PropTypes.shape({
-			primary: PropTypes.string,
-			border: PropTypes.string,
-			disabledBackground: PropTypes.string,
-			disabledBorder: PropTypes.string,
-		}),
-		type: PropTypes.string,
-		children: PropTypes.node,
-		/** See the docs for how to override styles properly  */
-		className: PropTypes.string,
-		/** Disables automatic blur */
-		disableAutoBlur: PropTypes.bool,
-		disabled: PropTypes.bool,
-	};
+const Checkbox = React.forwardRef(function Checkbox(props, ref) {
+	const {
+		onClick,
+		onMouseUp,
+		title,
+		isChecked,
+		theme,
+		type,
+		children,
+		className,
+		disableAutoBlur,
+		disabled,
+		...checkboxProps
+	} = props;
+
+	const isCoreTheme = isSystemTheme(theme);
+	const themedProps = { ...checkboxProps, ...(isCoreTheme ? {} : theme) };
 
 	/* eslint-disable react/prop-types */
-	onMouseUp = e => {
-		if (this.props.onMouseUp) {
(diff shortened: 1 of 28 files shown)


theme/core.js
```diff
@@ -174,6 +174,14 @@ colors.popover = {
 	background: colors.backgroundPrimary,
 };
 
+colors.dropdown = {
+	background: colors.white,
+	backgroundHover: colors.gray4,
+	foreground: colors.black,
+	separatorColor: colors.gray14,
+	foregroundDisabled: colors.gray14
```
RobertBolender, line 182: The other 4 values in this PR match their previous values, except `foregroundDisabled` which used to be gray22. Was that an intentional change?

components/dropdown/styled.jsx
```diff
@@ -1,11 +1,13 @@
 import styled from 'styled-components';
 import { colors, thickness } from '../shared-styles';
 import { resetStyles } from '../utils';
+import { theme } from '../../theme';
```
RobertBolender, line 4: Importing the theme directly like this will prevent consumers from being able to provide their own theme.
Prefer to use the `theme` object that styled-system will provide as a prop.

Here's an earlier discussion of this for reference:
https://github.com/Faithlife/styled-ui/pull/287#discussion_r344324475
https://github.com/Faithlife/styled-ui/pull/287#discussion_r348092016

o2x7: ok, i wasn't quite sure how that worked.  Ill change, thanks

theme/core.js
o2x7: UI spec here: https://faithlife.atlassian.net/browse/FLTV-1251.  Good catch! I wasn't intending to change from 22->12.  Ill correct

components/dropdown/styled.jsx
```diff
@@ -64,7 +75,8 @@ export const MenuSeparator = styled.hr.attrs(() => ({
 	'aria-orientation': 'horizontal',
 }))`
 	border: 0;
-	border-top: 1px solid ${colors.gray14};
+	border-top: 1px solid
+		${theme.colors.dropdown ? theme.colors.dropdown.separatorColor : colors.gray14};
```
bryanrsmith, line 79: "color" is redundant since we're already in the color scale. `theme.colors.dropdown.separator` seems fine.

bryanrsmith:
> Could you also add an example of that style customization to `catalog/dropdown/variations.md`?

Customization would happen by injecting a new theme object. I don't think it's appropriate to get into this in individual component documentation. We need a separate page documenting how this works, and then component docs probably only need to describe the theme keys they use. I'm fine with omitting all of that from this PR.

bryanrsmith:
More context for those interested: I've been working on full theme customizability in Smart Media Editor largely by using this pattern of referencing component-specific colors from the theme object. The interesting theme bits are at https://git.faithlife.dev/Logos/smart-media-editor/blob/master/packages/smart-media-editor/components/ThemeProvider.jsx. I intend to push much of this work into styled-ui once it stabilizes.

o2x7:
This should be ready to rock, got it integrated into FLTV

bryanrsmith:
1. Can we rename the `theme` object that we're accepting as a prop and sticking in context? Looks like it wasn't used in menuItem, but it is used in menu-checkbox. Unfortunately we have to continue accepting the old `theme` prop, but the code is confusing with two `theme` objects being passed around, so if we can rename it in the top-level dropdown that would be nice.

2. It looks to me like the menu checkbox does not pull colors from the styled system theme, but it does accept colors from the old theme prop. We need to update it as well to accept either theming strategy.

o2x7:
Got your feedback accounted for.  Ready for re-review @bryanrsmith

o2x7:
Some alternative theme colors for your use when you review the PR if you want to see visual changes in the categories UI:


`
colors.dropdown = {
        /* FLTV */
        background: 'black',
        backgroundHover: 'black',
        foreground: 'white',
        separator: colors.gray14,
        foregroundDisabled: colors.gray22,
    };
        
    colors.checkbox = {
        /* FLTV */
        background: colors.white,
        primary: colors.blueBase,
        border: '#95908f',
        disabledBackground: colors.gray8,
        disabledBorder: colors.gray22,
    };
`

components/check-box/component.jsx
```diff
@@ -3,71 +3,89 @@ import PropTypes from 'prop-types';
...
+		type,
+		children,
+		className,
+		disableAutoBlur,
+		disabled,
+		...checkboxProps
+	} = props;
+
+	const isCoreTheme = isSystemTheme(theme);
+	const themedProps = { ...checkboxProps, ...(isCoreTheme ? {} : theme) };
```
bryanrsmith, line 25: Why are we spreading props onto this object? It doesn't look like this was happening before.

components/check-box/styled.jsx
```diff
@@ -3,17 +3,20 @@ import { resetStyles } from '../utils';
 
 export const CheckboxDiv = styled.div`
 	position: absolute;
-	border: solid 1px ${props => props.theme.border};
+	border: solid 1px
+		${({ theme, themeOverrides }) => themeOverrides?.border || theme.colors.checkbox.border};
```
bryanrsmith, line 7: Are you deliberately ignoring values like empty string in the themeOverrides? The nullish coalescing operator is more commonly paired with optional chaining: `themeOverrides?.border ?? theme.colors.checkbox.border`.

components/dropdown/component.jsx
```diff
@@ -1,19 +1,28 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { DropdownCore } from './dropdown-core';
+import { DefaultThemeProvider } from '../DefaultThemeProvider';
+import { isSystemTheme } from '../../theme';
+
+const Dropdown = React.forwardRef(function Dropdown(props, ref) {
```
bryanrsmith, line 7: Where is `ref` used?

components/dropdown/styled.jsx
```diff
@@ -1,16 +1,19 @@
...
 	padding: ${thickness.four} 0;
 
+	background-color: ${({ theme, themeOverrides }) =>
+		themeOverrides?.background || theme.colors.dropdown.background};
 	display: flex;
 	flex-direction: column;
 `;
 
-export const MenuItem = styled.button.attrs(({ role, isDisabled }) => ({
+export const MenuItem = styled.div.attrs(({ role, isDisabled }) => ({
```
bryanrsmith, line 15: This doesn't seem right. Why did the type of these elements change?

components/dropdown/styled.jsx
```diff
@@ -1,16 +1,19 @@
 import styled from 'styled-components';
 import { colors, thickness } from '../shared-styles';
 import { resetStyles } from '../utils';
+import { Box } from '../Box';
```
bryanrsmith: This looks unused.

components/check-box/component.jsx
```diff
@@ -3,71 +3,89 @@ import PropTypes from 'prop-types';
...
-			this.props.onMouseUp(e);
+	const handleMouseUp = e => {
+		if (props.onMouseUp) {
+			props.onMouseUp(e);
 		}
 
-		if (!this.props.disableAutoBlur && this.componentRef.current) {
-			this.componentRef.current.blur();
+		if (!props.disableAutoBlur) {
+			blur();
```
bryanrsmith, line 34: Where is this defined?

o2x7: I am not sure - and I had the same question, it was there before I touched the file.

o2x7: I wasn't sure of the use of disableAutoBlur. this might crash if that is set though

components/dropdown/styled.jsx
o2x7: `styled.button` the theme property does not automatically propagate down to the children of the control. wth `styled.div` it does.  Robert helped me discover that and work through the change.

components/check-box/styled.jsx
o2x7: I wasn't aware of the difference.  I noticed `||` in use previously and it seemed to work.  Thanks for letting me know the difference , Ill modify to use `??`

components/dropdown/styled.jsx
RobertBolender: When we were investigating this, we had observed the different behavior between a `styled(Box)` and a `styled.button`, but I was not aware of any reason for changing from a `button` to a `div`.

o2x7: hmm, I know I was getting the same issue in this case.  Let me test it again and see if I can swap back to a `styled.button`

o2x7: `styled.button` is working fine.  I'll revert.

components/check-box/component.jsx
bryanrsmith: The old code has `this.componentRef.current.blur();`, and it looks like that functionality has been lost in this change. We will need to restore it.

PR #333, opened 2020-05-06
Workaround for server-side rendering issue.

timothy-b:
This is a workaround for https://github.com/reactjs/react-transition-group/issues/618.

Message me for repro info.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
 		"react-popper": "^1.3.3",
 		"react-scrollbar": "^0.5.4",
 		"react-select": "^2",
-		"react-transition-group": "~4.2.1",
+		"react-transition-group": "^4.4.1",
 		"reactstrap": "^6.1.0",
 		"styled-system": "^5.1.1"
 	},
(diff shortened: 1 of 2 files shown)


bryanrsmith:
> Message me for repro info.

Instead, put the repro info here so there's a record of it.

The issue you linked to states that the bug is resolved in 4.4.1. Why are we pinning to an older version instead of updating?

timothy-b:
> > Message me for repro info.
> 
> Instead, put the repro info here so there's a record of it.
> 
> The issue you linked to states that the bug is resolved in 4.4.1. Why are we pinning to an older version instead of updating?

Oh nice, they just fixed it a few hours ago. Updated: https://github.com/Faithlife/styled-ui/pull/333/commits/1043fc5c25d29a9bedb222c6014abfbeaf95359d. To repro the issue, pull `master` from Faithlife and delete your node_modules, then run `yarn` and `yarn build`. You can then either `yarn start` and visit the local site, or run ` node --require .\src\Faithlife.Web\Scripts\dist\server\server-start.js` for a more helpful error message.

PR #335, opened 2020-05-07
Update DatePicker, Add min/max dates

schyler-evans:
Add two new optional attributes to DatePicker and DatePickerInput: minDate and MaxDate.

If supplied, these will limit the datepicker to dates greater than minDate (if supplied) and less than maxDate (if supplied)

### DIFF
--- a/components/date-picker/component.jsx
+++ b/components/date-picker/component.jsx
@@ -96,12 +96,12 @@ export class DatePicker extends Component {
 		this.setMonth(this.props.dateFunctions.addMonths(this.state.currentMonth, 1));
 
 	canDecrementMonth = weeks => {
-		var firstWeek = weeks[0];
+		const firstWeek = weeks[0];
 		return !this.props.minDate || this.props.minDate < firstWeek[0];
 	};
 
 	canIncrementMonth = weeks => {
-		var lastWeek = weeks[weeks.length - 1];
+		const lastWeek = weeks[weeks.length - 1];
 		return !this.props.maxDate || this.props.maxDate > lastWeek[lastWeek.length - 1];
 	};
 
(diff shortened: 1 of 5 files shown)


components/date-picker/component.jsx
```diff
@@ -93,26 +95,42 @@ export class DatePicker extends Component {
 	incrementMonth = () =>
 		this.setMonth(this.props.dateFunctions.addMonths(this.state.currentMonth, 1));
 
+	canDecrementMonth = weeks => {
+		var firstWeek = weeks[0];
+		return !this.props.minDate || this.props.minDate < firstWeek[0];
+	};
+
+	canIncrementMonth = weeks => {
+		var lastWeek = weeks[weeks.length - 1];
```
bryanrsmith, line 104: Use `const`. I'm not sure why this is allowed by our lint config, actually.

components/date-picker/component.jsx
```diff
@@ -93,26 +95,42 @@ export class DatePicker extends Component {
...
 		} = this.props;
 		const { currentMonth, weeks } = this.state;
 
 		return (
 			<Fragment>
 				<Styled.Header>
-					<Styled.ChangeMonth onClick={this.decrementMonth} tabIndex="-1">
-						<Caret style={{ transform: 'scaleX(-1)', color: colors.gray66 }} />
-					</Styled.ChangeMonth>
+					{this.canDecrementMonth(weeks) && (
```
bryanrsmith, line 123: Won't this cause layout shifts? Consider disabling the buttons rather than removing them.

package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@faithlife/styled-ui",
-	"version": "5.54.2",
+	"version": "5.55.0",
```
bryanrsmith, line 3: We change version numbers are release time. You can revert this.

components/date-picker/component.jsx
schyler-evans: Good call!

PR #338, opened 2020-05-13
Add popover v6 implementation

EthanRutherford:
Updates react-popper to v2, and introduces a new version of `Popover` which uses the new hook-based api.

The old popover is left mostly as-is, and maps a handful of older props to the newer equivalents.
The exception is the `modifiers` prop, as the modifiers api has changed. However, a github search shows only one consumer even uses a modifier, so I am willing to contribute a fix to that consumer.

Questions you may have:
Q: Where's the v6 PopoverBase equivalent?
A: PopoverBase appears to exist solely to provide an unstyled version of the popover. Since we expose `usePopover`, a consumer can simply use that with a `Box` to get an unstyled version.

Q: What about the Tooltip component?
A: I think the new popover covers the same usecases as tooltip. If not, however, it should be pretty straightforward to add one that uses the new popover.

Q: Why does the api use `useState` for the popper and arrow refs, instead of `useRef`?
A: the short answer is, this is what's shown in the [react-popper docs](https://popper.js.org/react-popper/v2/#example).
The long answer is, the `usePopper` api accepts dom elements, and not ref objects. I'm not *entirely* sure why this is, but I *think* it has to do with changes to ref objects not typically causing `useEffect` effects to rerun. and usePopper has to dispose and/or construct the internal `popper` instance any time these references change (to and from `null`, or to a different dom element).
As best I can tell, this wouldn't work with `ref`s, because deps are checked *before* render, and refs are updated *after* render. So, as unfortunate as it is, initializing the popper takes two rerenders.

element states start out null. A setstate is fired when we render an element (which may not be on the first render, it could be down the road), which triggers a rerender. useEffect sees that the the elements changed, and so will fire the effect after render. The effect will create a popper instance, and the popper instance will calculate styles. Finally, the popper instance will trigger one last rerender to pass the calculated styles to the component.

### DIFF
--- a/components/popover-v6/index.jsx
+++ b/components/popover-v6/index.jsx
@@ -1,0 +1,113 @@
+/* eslint eqeqeq: ["error", "always", {"null": "never"}] */
+import React, { useState, useEffect } from 'react';
+import ReactDOM from 'react-dom';
+import PropTypes from 'prop-types';
+import { usePopper } from 'react-popper';
+import { PopoverContainer, PopoverArrow } from './styled';
+
+export function usePopover(reference, options) {
+	const [popperElement, setPopper] = useState();
+	const [arrowElement, setArrow] = useState();
+	const { styles, attributes, ...rest } = usePopper(reference, popperElement, {
+		...options,
+		modifiers: [
+			...(options?.modifiers ?? []),
+			{ name: 'arrow', options: { element: arrowElement } },
+		],
+	});
+
+	return {
+		popperElement,
+		popperProps: { ...attributes.popper, style: styles.popper, ref: setPopper },
+		arrowProps: { ...attributes.arrow, style: styles.arrow, ref: setArrow },
+		...rest,
+	};
+}
+
+const arrowOffset = { name: 'offset', options: { offset: [0, 10] } };
+export function Popover({
+	reference,
+	placement,
+	modifiers,
+	strategy,
+	container,
+	hideArrow,
+	onFocusAway,
+	children,
+	...restProps
+}) {
+	const { popperElement, popperProps, arrowProps } = usePopover(reference, {
+		placement,
+		modifiers: hideArrow ? modifiers : [arrowOffset].concat(modifiers ?? []),
+		strategy,
+	});
+
+	useEffect(() => {
+		if (popperElement != null && onFocusAway instanceof Function) {
+			const onClick = event => {
+				if (!popperElement.contains(event.target)) {
+					onFocusAway();
+				}
+			};
+
+			document.addEventListener('click', onClick);
+			return () => document.removeEventListener('click', onClick);
+		}
+	}, [onFocusAway, popperElement]);
+
+	const popover = (
+		<PopoverContainer {...restProps} {...popperProps}>
+			{children}
+			{!hideArrow && <PopoverArrow {...arrowProps} />}
+		</PopoverContainer>
+	);
+
+	if (container != null) {
+		return ReactDOM.createPortal(popover, container === 'body' ? document.body : container.current);
+	}
+
+	return popover;
+}
+
+Popover.propTypes = {
+	/** the element or virtual element to anchor to */
+	reference: PropTypes.oneOfType([
+		PropTypes.element,
+		PropTypes.shape({
+			getBoundingClientRect: PropTypes.func.isRequired,
+			contextElement: PropTypes.element,
+		}),
+	]),
+	/** where on the target the popover anchors */
+	placement: PropTypes.oneOf([
+		'auto',
+		'auto-start',
+		'auto-end',
+		'top',
+		'top-start',
+		'top-end',
+		'bottom',
+		'bottom-start',
+		'bottom-end',
+		'right',
+		'right-start',
+		'right-end',
+		'left',
+		'left-start',
+		'left-end',
+	]),
+	/** modifiers allow custom behavior, see https://popper.js.org/docs/v2/modifiers */
+	modifiers: PropTypes.arrayOf(
+		PropTypes.shape({
+			name: PropTypes.string.isRequired,
+		}),
+	),
+	/** whether popover should use absolute or fixed positioning */
+	strategy: PropTypes.oneOf(['absolute', 'fixed']),
+	/** where to inject the popover, defaults to inline */
+	container: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
+	/** whether or not to show the arrow */
+	hideArrow: PropTypes.bool,
+	/** handler useful for closing popover when clicking away */
+	onFocusAway: PropTypes.func,
+};
--- a/components/popover-v6/index.jsx
+++ b/components/popover-v6/index.jsx
@@ -1,4 +1,3 @@
-/* eslint eqeqeq: ["error", "always", {"null": "never"}] */
 import React, { useState } from 'react';
 import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
@@ -52,7 +51,7 @@ export function Popover({
 		</PopoverContainer>
 	);
 
-	if (container != null) {
+	if (container !== null && container !== undefined) {
 		return ReactDOM.createPortal(popover, container === 'body' ? document.body : container.current);
 	}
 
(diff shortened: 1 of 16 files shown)


components/popover-v6/index.jsx
```diff
@@ -0,0 +1,113 @@
...
+}
+
+const arrowOffset = { name: 'offset', options: { offset: [0, 10] } };
+export function Popover({
+	reference,
+	placement,
+	modifiers,
+	strategy,
+	container,
+	showArrow = true,
```
RobertBolender, line 34: Nit: consider using a `hideArrow` boolean, rather than a boolean prop that defaults to true.

components/popover/popover-base.jsx
```diff
@@ -85,6 +78,7 @@ export class PopoverBase extends Component {
 	_timeout = null;
 
 	componentDidMount() {
+		console.warn('Popover v1 is deprecated, please migrate to popover v2');
```
RobertBolender, line 81: Nit: consider using our deprecate utility to avoid repeated warnings:
https://github.com/Faithlife/styled-ui/blob/master/components/utils/deprecate.js

components/popover/popover-base.jsx
```diff
@@ -145,20 +139,21 @@ export class PopoverBase extends Component {
...
-			computeStyle: {
-				...modifiers.computeStyle,
-				gpuAcceleration: false,
-			},
-		};
+		const popperModifiers = [];
+		if (modifiers instanceof Array) {
+			popperModifiers.push(...modifiers);
+		} else {
+			console.warn(
```
RobertBolender, line 146: Nit: consider using our deprecate utility to avoid repeated warnings:
https://github.com/Faithlife/styled-ui/blob/master/components/utils/deprecate.js

EthanRutherford: Ah, neat, I was not aware of that utility 👍

components/popover-v6/index.jsx
```diff
@@ -0,0 +1,113 @@
...
+	const { popperElement, popperProps, arrowProps } = usePopover(reference, {
+		placement,
+		modifiers: hideArrow ? modifiers : [arrowOffset].concat(modifiers ?? []),
+		strategy,
+	});
+
+	useEffect(() => {
+		if (popperElement != null && onFocusAway instanceof Function) {
+			const onClick = event => {
+				if (!popperElement.contains(event.target)) {
```
RobertBolender, line 48: Popover v1 used the `useFocusAwayHandler` hook to mark Popover content as "inbounds" for focus handling. Was it an intentional decision to not use that hook for Popover v2?
One visible difference is that it was possible to click on a v1 Popover without it blurring, but clicking on a v2 Popover will dismiss the popover.

EthanRutherford: Clicking on a v2 popover should not dismiss it.

EthanRutherford: useFocusAwayHandler is a bit.. odd.
It assumes that something becomes *focused* when the popover is opened, which is only the case when the popover is attached to clicking a button. Hovering, for example, does *not* cause anything to become focused, and therefore the "focus away" event won't fire when you click somewhere, because nothing was focused to begin with.

The only true visible difference between this and `useFocusAway` is that popup v2 will not dismiss when *tabbing* away from a clicked button.

I'm open to alternative solutions, but I'm not entirely sure what they would be.

bryanrsmith: > popup v2 will not dismiss when tabbing away from a clicked button

As far as I can remember, this is the reason `useFocusAway` exists, and a not insignificant amount of effort went into ensuring that popups would not be broken for keyboard users.

EthanRutherford: I have attempted a rewrite of focusAwayHandler, commit incoming.

EthanRutherford: I've rewritten `useFocusAway`, borrowing from the original implementation to provide a simpler implementation which supports v2. In my testing, behavior is the same as before in all cases.

components/popover-v6/index.jsx
```diff
@@ -0,0 +1,104 @@
+/* eslint eqeqeq: ["error", "always", {"null": "never"}] */
```
bryanrsmith, line 1: Please avoid unnecessarily disabling lint rules that don't match personal preferences.

components/popover-v6/index.jsx
bryanrsmith: @korbinancell would you mind taking a quick look at Ethan's changes? My work with this code has been limited.

components/popover-v6/index.jsx
EthanRutherford: This is not simply personal preference: I believe the rule as currently configured leads to writing bugs. I believe I have a strong case for changing the rule to the above settings.

(though I do not believe a github comment is necessarily the best medium to discuss)

bryanrsmith: > (though I do not believe a github comment is necessarily the best medium to discuss)

Yes. Open an issue in the coding guidelines repo. In the meantime, please confirm to guidelines.

components/popover-v6/index.jsx
EthanRutherford: I will note, while it does *appear* as if clicking inside the pop-up dismisses it for the demos that have multiple popups, this is actually just a side effect of the fact that all of the poppers in that demo share the same `isOpen` state. So, when clicking on the content of a pop-up, one of the *other popups* will detect the focus change, which is not inside its own bounds, and toggle off the shared state. I could make this more clear by using separate `isOpen` state values for each popper.

bryanrsmith:
@korbinancell W/o objection, I'll merge this tomorrow morning. Let me know if you need more time.

korbinancell:
@bryanrsmith I will take a look this afternoon

components/utils/focus-away-handler.jsx
```diff
@@ -1,16 +1,10 @@
-import React, { useEffect } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import { useFocusAwayHandler } from '../shared-hooks';
 
 export function FocusAwayHandler({ children, onFocusAway }) {
 	const targetRef = useFocusAwayHandler(onFocusAway);
 
-	useEffect(() => {
```
korbinancell, line 8: Bryan was right about this breaking keyboard nav. When the popover is opened we need to focus something in the popover (especially when it is in a portal) so that you can navigate through the popover normally as well as handling tabbing out of the popover and refocusing the popover reference.

EthanRutherford: Are you sure? As far as I can tell, tabbing away currently functions exactly as expected, and closes the popover.

EthanRutherford: The new code essentially behaves this way: the popover itself is made focusable (using the same tabIndex=-1 trick the previous code used to make the popoverManager focusable).

Then, the handlers will detect any change in focus on the page at all. If the focus has changed to anything that is not the popover (or a child of the popover), the popover closes.

EthanRutherford: I haven't tested in browsers other than Chrome, but I wouldn't imagine the behavior to be different between browsers. If it is, or I'm misunderstanding something, let me know.

korbinancell: Hmm. I'm in Chrome as well, I pulled down your pr and added a button to the `using a container prop` example. When I activate the popover with the keyboard focus stays on the button and doesn't move to the popover.

EthanRutherford: That is true, but that appears to be what the current code does as well.

EthanRutherford: Which is not to say that I can't fix it if that was the desired behavior :)

EthanRutherford: (also a note: this particular component was broken prior to this PR, because the `useFocusAwayHandler` previously returned an object with a targetRef property, not a targetRef directly. This wrapper is not used by anything in styled-ui, and presumeably doesn't have any external consumers either, else we'd have probably heard about it not working.)

EthanRutherford: @korbinancell I have refined useFocusAwayHandler, and I believe it should now behave exactly as you have described, focusing the popup when it is opened, and returning focus to the previously focused element when it is closed. Please give it another pass when you have a chance.

PR #340, opened 2020-05-22
Fix minor regressions in Popover

EthanRutherford:
Fixes a couple regressions reported by the ChMS team.

To repro and test fix:
Regression 1: I had removed the wrapper div in PopoverManager because I thought it was no longer necessary, but certain flex properties on the parent caused visual differences without the wrapper. The wrapper div is now back (so I suppose just verify the div is in the dom, where it wasn't before)

Regression 2: Clicking a popover button to close an open popover would flash it closed and back open. I *think* I remember this being the case even before my changes, but regardless it should now be fixed. Verify that clicking the button should now actually close popovers.


components/utils/focus-away-handler.jsx
```diff
@@ -1,22 +0,0 @@
-import React from 'react';
```
EthanRutherford, line 1: This component wasn't referenced anywhere, and did not appear to be exported from the package either, so I think it's dead code.

PR #342, opened 2020-05-26
Fix disabled checkbox styling.

rmjohnson:
I noticed that the disable checkbox started looking like the normal checkbox recently. Eventually was able to track it down to this. I assume the string template inside the string template doesn't pass through the props again?


rmjohnson:
@bryanrsmith, can I get a merge on this? I'm shipping a new feature that will be potentially confusing with the current styling.

bryanrsmith:
Sorry! I was waiting on the checks to pass and I forgot. I wish there was a button to auto-merge when all checks pass.

bryanrsmith:
v5.57.4 is building now and should be published shortly.

PR #351, opened 2020-07-03
Scroll long lists in ParameterSelect.

danielstephenson:
@korbinancell

### DIFF
--- a/components/parameter-sentence/parameter-select.jsx
+++ b/components/parameter-sentence/parameter-select.jsx
@@ -56,7 +56,15 @@ export function ParameterSelect({
 							</Styled.Button>
 						)}
 					</ListboxToggle>
-					<ListboxMenu styleOverrides={{ zIndex: 10, padding: '4px 0', width }}>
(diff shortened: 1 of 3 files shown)


components/parameter-sentence/parameter-select.jsx
```diff
@@ -56,7 +56,15 @@ export function ParameterSelect({
...
 						)}
 					</ListboxToggle>
-					<ListboxMenu styleOverrides={{ zIndex: 10, padding: '4px 0', width }}>
+					<ListboxMenu
+						styleOverrides={{
+							zIndex: 10,
+							padding: '4px 0',
+							width,
+							overflow: styleOverrides.overflow ?? 'auto',
+							maxHeight: styleOverrides.maxHeight ?? '300px',
```
danielstephenson, line 65: @korbinancell I don't necessarily need to add new overrides to `ParameterSelect` *and* add the new defaults, but I'll need one or the other, whichever way you prefer.

PR #352, opened 2020-07-09
add optional prop for number input to always show increment buttons

kevinforrestkeyes:
this is part of the work being done on [SITES-6418](https://faithlife.atlassian.net/browse/SITES-6418) where we would like to use the NumberInput for the quantity selector, but it has been requested that the buttons are always shown within the input, not just on hover/focus.

### DIFF
--- a/components/input/NumberInput.jsx
+++ b/components/input/NumberInput.jsx
@@ -1,4 +1,5 @@
 import React, { useCallback, useRef, useMemo } from 'react';
+import PropTypes from 'prop-types';
 import { css } from 'styled-components';
 import { Box } from '../Box';
 import { Input } from './Input';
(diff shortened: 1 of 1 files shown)


components/input/NumberInput.jsx
```diff
@@ -72,10 +75,12 @@ const NumberInput = React.memo(
 
 NumberInput.defaultProps = {
 	theme,
+	alwaysShowButtons: false,
```
RobertBolender, line 78: Super nit: this line is unnecessary because a boolean prop defaults to `false`

kevinforrestkeyes:
@RobertBolender I've removed the unnecessary default prop, if it looks all good to you, could you merge this for me?

PR #364, opened 2020-09-14
Add DatePickerInput Config children

korbinancell:
Possible alternative to https://github.com/Faithlife/styled-ui/pull/363

### DIFF
--- a/components/date-picker-input/component.jsx
+++ b/components/date-picker-input/component.jsx
@@ -1,5 +1,6 @@
 import React, { useRef, useState, useEffect, useCallback } from 'react';
 import PropTypes from 'prop-types';
+import { getConfigProps } from '../utils';
 import { Popover } from '../popover-v6';
 import { PlacementOptionsProps } from '../popover/popper-helpers';
 import { Calendar as CalendarIcon } from '../icons';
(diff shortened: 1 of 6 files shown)


components/date-picker-input/component.jsx
```diff
@@ -94,9 +95,21 @@ export function DatePickerInput({
...
+		child => child.type.childConfigComponent,
+	);
+	const popoverProps = configChildren.find(
+		child => child.type.childConfigComponent === DatePickerInputPopover.childConfigComponent,
+	)?.props;
+	const buttonProps = configChildren.find(
+		child => child.type.childConfigComponent === DatePickerInputButton.childConfigComponent,
+	)?.props;
+
+	console.log(React.Children.toArray(children), configChildren, buttonProps);
```
RobertBolender, line 107: Super nit: leftover console.log

components/date-picker-input/component.jsx
```diff
@@ -200,3 +215,16 @@ DatePickerInput.propTypes = {
...
 	}),
 };
+
+function DatePickerInputPopover(props) {
+	return <div />;
+}
+DatePickerInputPopover.childConfigComponent = 'DatePickerInputPopover';
+
+function DatePickerInputButton(props) {
+	return <div />;
```
RobertBolender, line 224: Nit: would this also function if you return `null` instead of an empty div?

components/date-picker-input/styled.jsx
```diff
@@ -13,7 +13,9 @@ export const Container = styled.div`
 	color: ${colors.flGray};
 `;
 
-export const CalendarButton = styled(UtilityButton)`
+export const CalendarButton = styled(UtilityButton).attrs(({ color }) => ({
+	color: color ?? 'gray52',
```
RobertBolender, line 17: Nit: should this be a semantic theme color property to support dark-mode or other theme contexts?

RobertBolender (commented):
It's kind of magic, but I kind of like it

bryanrsmith (commented):
lgtm. Accepting `Date` instances via props is probably not ideal because many consumers will either construct them at render time and pass a new object on every render, or annoyingly have to memoize them. It's too late for that, though, since the old implementation established the pattern.

I generally prefer to actually render the descendent component being used for configuration rather than manipulating `children`. (e.g. make `DatePickerInput.Button` render the button instead of collecting its props and discarding the element). I think it makes the code easier to follow and it often gives consumers more flexibility by letting them render the component in any part of the subtree instead of requiring it to be a direct child. That does require context to route internal data, though, and I'm totally fine with keeping this implementation here if it's easier.

korbinancell:
@bryanrsmith I think I would rather stick with the configuration as is, I feel it is easier for us and the consumer so they won't have to add context and what not just to change the width of the popover or such. Perhaps a mix of these is the best way to go, especially in components that make sense to be a little more free form.

PR #367, opened 2020-09-15
Center the date picker arrow buttons

lawrencesin:
Aligned the date picker's previous and next month's buttons centered according to the [spec](https://app.zeplin.io/project/58a23c2d4597afb248c13a43/screen/5bedfd6811d4fc12b2f92d4a).

Link to Issue: https://github.com/Faithlife/styled-ui/issues/368

### DIFF
--- a/components/date-picker/styled.jsx
+++ b/components/date-picker/styled.jsx
@@ -4,6 +4,7 @@ import { colors } from '../shared-styles';
 export const ChangeMonth = styled.button`
 	display: flex;
 	justify-content: space-around;
+	align-items: center;
 	border: none;
 	margin: 0;
 	padding: 0;


components/date-picker/styled.jsx
```diff
@@ -4,6 +4,7 @@ import { colors } from '../shared-styles';
 export const ChangeMonth = styled.button`
 	display: flex;
 	justify-content: space-around;
+	align-items: center;
```
korbinancell, line 7: It may just be the client im using, but is the spacing off here? Maybe using spaces instead of tabs?

lawrencesin: It was a tab, but the file was saved in CRLF instead of LF line endings

korbinancell (approved):
I'll merge when the build finishes

PR #370, opened 2020-09-23
Expose resize property to textarea.

smaxa20:


RobertBolender (approved):

components/input/Input.jsx
```diff
@@ -134,6 +139,8 @@ const StyledInput = styled.input(
 	border-color: ${theme.colors.input.border};
 	color: ${theme.colors.input.foreground};
 
+	resize: ${resize};
```
bryanrsmith, line 142: This is fine if you prefer it, but you can also tell styled system to map the property for you: `${system({ resize: true })}` will tell it to map the `resize` prop to the css property of the same name. We do this in a few other places (like components that deal with typography, where we add support for things like `textTransform`).

PR #373, opened 2020-09-25
Select text on focus in DatePickerInput.

smaxa20:


bryanrsmith:
`Input` does accept a boolean `selectOnFocus` prop that can do this for you. `<Input selectOnFocus {...others} />`

PR #374, opened 2020-09-29
SUI-3 Updates Accordion to new spec w/ example

mdombrowski1997:
## [SUI-3](https://faithlife.atlassian.net/browse/SUI-3)

* [Thread with discussion](https://www.flowdock.com/app/logos/styled-ui/threads/RuF6-PPZ7Gb-338iT0sZo3Lo_P2)
* [Zeplin spec](https://zpl.io/Vkdodz8)

### DIFF
--- a/components/accordion/accordion-header.jsx
+++ b/components/accordion/accordion-header.jsx
@@ -1,6 +1,7 @@
 import React, { useCallback, useEffect, useRef } from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
+import { colors } from '../shared-styles';
 import { Text } from '../Text';
 import { Box } from '../Box';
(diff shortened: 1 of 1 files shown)


components/accordion/accordion-header.jsx
```diff
@@ -100,7 +102,7 @@ export function AccordionHeader({
 							<Text
 								textStyle="ui.14"
 								display="block"
-								color="accordion.sectionHeaderSubtitle"
+								color="inherit"
```
mdombrowski1997, line 105: Is there a way to pass `:hover` styles to this component?  We only need `inherit` on hover so it can change to blue, we still want the lighter gray when not hovering, but this makes it lose that.

PR #379, opened 2020-11-03
Dropdown v6 Update

korbinancell:


rmjohnson:
My non-designer eyes can't tell for certain but it *looks* like the arrow isn't vertically aligned (feels a little closer to the top than the bottom).
image

components/check-box/styled.jsx
```diff
@@ -3,7 +3,7 @@ import { resetStyles } from '../utils';
 import { Box } from '../Box';
 
 export const CheckboxDiv = styled(Box)`
-	position: absolute;
+	position: ${({ position }) => position || 'absolute'};
```
rmjohnson, line 6: `??` ?

components/dropdown/dropdown-children.jsx
```diff
@@ -0,0 +1,207 @@
...
+	{ disabled, href, children, ...menuItemProps },
+	ref,
+) {
+	return (
+		<Box as="li" role="none">
+			<MenuItem
+				as="a"
+				ref={ref}
+				onClick={null}
+				onKeyDown={null}
```
rmjohnson, line 170: This seems to be intentional (and I'm not sure on the accessibility rules around this) but I was a little surprised that I couldn't use enter to "click" on the link in the dropdown.

components/dropdown/legacy-dropdown/index.js
```diff
@@ -0,0 +1,44 @@
...
+const depWarning = (name, newName) =>
+	`${name} will be moving to ${newName} in Styled-UI version 6.\n${v6ImportHelpText}`;
+
+const Dropdown = deprecateComponent(
+	DropdownComponent,
+	`Prefer the Styled-UI version 6 Dropdown.\n${v6ImportHelpText}`,
+);
+const DropdownMenu = deprecateComponent(
+	DropdownMenuComponent,
+	depWarning('DropdownMenu', 'Dropdown.Menu'),
```
rmjohnson, line 24: Definitely read this as `derpWarning` at first :joy:

components/dropdown/utils.js
```diff
@@ -0,0 +1,95 @@
...
+					return;
+			}
+		},
+		[onToggleMenu, onKeyboardNav],
+	);
+
+	return handleKeyboardActivate;
+}
+
+export function useKeyboardNavigate(onCloseMenu, onKeyboardNav) {
```
rmjohnson, line 58: Fancy!

theme/core.js
```diff
@@ -185,8 +185,8 @@ colors.select = {
 
 colors.dropdown = {
 	background: colors.white,
-	backgroundHover: colors.gray4,
-	foreground: colors.black,
+	backgroundHover: colors.gray8,
```
rmjohnson, line 188: I trust this was intentional / needed.

rmjohnson (approved):

rmjohnson:
@RobertBolender, my current project is blocked on this component so I'm hoping to get it shipped here soon. Do you have bandwidth to review today?

components/dropdown/dropdown-menu.jsx
```diff
@@ -1,21 +1,55 @@
...
+				width={width || Styled.defaultMenuWidth}
+				padding={0}
+				onKeyDown={onKeyPress}
+				{...popoverProps}
+			>
+				{React.Children.map(children, (child, index) =>
+					React.isValidElement(child) && child.type.isFocusableChild
+						? React.cloneElement(child, {
+								ref: registerItem(index),
+								keyboardHovered: focusedItemIndex === index,
```
RobertBolender, line 48: "keyboardHovered" strikes me as strange, is this different from "focused"?
Mouse cursors can hover, keyboards cannot.

components/dropdown/dropdown.jsx
```diff
@@ -0,0 +1,98 @@
...
+			onToggleMenu();
+			splitRef.current ? splitRef.current.focus() : toggleRef.current?.focus();
+		}
+	}, [isOpen, onToggleMenu]);
+
+	const registerItem = useCallback(index => {
+		return ref => {
+			let newList = [...itemList.current];
+			if (itemList.current.length < index) {
+				newList = newList.slice(0, index);
```
RobertBolender, line 37: I'm not understanding the intent of this conditional statement.
[Array.slice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice) already handles the case where the "end" parameter is greater than the length of the array:

> If end is greater than the length of the sequence, slice extracts through to the end of the sequence (arr.length).

components/dropdown/dropdown-menu.jsx
```diff
@@ -1,21 +1,55 @@
...
+			<Popover
+				id={menuId}
+				as="ul"
+				margin={0}
+				role="menu"
+				reference={toggleRef.current}
+				onFocusAway={onCloseMenu}
+				placement="bottom-start"
+				hideArrow
+				width={width || Styled.defaultMenuWidth}
```
RobertBolender, line 39: If this `defaultMenuWidth` const is only used here and not in `styled.jsx`, how would you feel about just hard-coding it here?

theme/core.js
RobertBolender: Discussed in Flow here: https://www.flowdock.com/app/logos/styled-ui/threads/2G-oohRCEJEeskb1UYx7h5GXr7Z

components/utils/get-config-props.js
```diff
@@ -4,3 +4,11 @@ export function getConfigProps(children, name) {
 	const childrenList = React.Children.toArray(children);
 	return childrenList.find(child => child.type.childConfigComponent === name)?.props ?? {};
 }
+
+export function getConfigChild(children, name) {
+	const childrenList = Array.isArray(children) ? children : React.Children.toArray(children);
+	const childIndex = childrenList.findIndex(child => child.type?.childConfigComponent === name);
+	return childIndex !== -1
+		? [childrenList[childIndex], childrenList.filter((_, index) => index !== childIndex)]
+		: [null, children];
```
RobertBolender, line 13: If I'm reading this correctly, the return value of this `getConfigChild` function is an array with two elements:
1. Either a single config component, if it exists, or else null
2. An array of all the children components that are not the config component.
Is that accurate? Would you consider adding a comment to that effect?

RobertBolender (approved):
I left a few non-blocking comments, but otherwise looks good to me 👍

components/dropdown/dropdown.jsx
korbinancell: Yep, and that is what it is being used for. I was just trying to avoid some unnecessary work of slicing and extending the array.

components/dropdown/dropdown-menu.jsx
korbinancell: I usually prefer avoiding magic numbers and mixing css and js so I tried to kill two birds with one stone with keeping the styling constants in a separate file.

RobertBolender: I still find it strange for a const that is only used in one location to be defined in a separate location, but I'll leave it as a non-blocking objection.

PR #380, opened 2020-11-09
Add responsive button sizes to modals.

korbinancell:


components/modal/modal-content.jsx
```diff
@@ -1,24 +1,28 @@
...
-			{...props}
-		>
-			{children}
-		</Box>
+		<ThemedBox paddingY={paddingY ?? contentPadding} overflow="hidden">
+			<Box
+				maxWidth="100%"
+				width={['100vw', 'auto']}
+				css={`
+					overflow-wrap: break-word;
```
RobertBolender, line 16: We could consider adding this as a supported styled-system prop, similar to how we did here:

https://github.com/Faithlife/styled-ui/blob/1168be9b99e19d67033c4c3b645addb62d272db1/theme/system.js#L40

RobertBolender (approved):

PR #381, opened 2020-11-09
SUI-13 - Add Avatar Selects

zmckinnon:
# [SUI-13](https://faithlife.atlassian.net/browse/SUI-13)

I am working on a People lookup component for Church Management.  This component uses an extremely common UX pattern of showing avatars for both the drop down options and the selected options.

### Demo
Check this out here: https://deploy-preview-381--faithlife-styled-ui.netlify.app/#/text-input/select?a=avatar-select

### Screenshot
Screen Shot 2020-11-09 at 1 15 08 PM

### DIFF
--- a/components/text-input-v2/select.jsx
+++ b/components/text-input-v2/select.jsx
@@ -330,22 +330,6 @@ export const AsyncSelect = React.forwardRef(({ components = {}, ...props }, ref)
 	);
 });
 
-export const AvatarSelect = React.forwardRef(({ components = {}, ...props }, ref) => {
-	return <Select ref={ref} {...props} components={{ ...avatarComponents }} />;
-});
-
-export const AvatarCreatableSelect = React.forwardRef(({ components = {}, ...props }, ref) => {
-	return <CreatableSelect ref={ref} {...props} components={{ ...avatarComponents }} />;
-});
-
-export const AvatarAsyncCreatableSelect = React.forwardRef(({ components = {}, ...props }, ref) => {
-	return <AsyncCreatableSelect ref={ref} {...props} components={{ ...avatarComponents }} />;
-});
-
-export const AvatarAsyncSelect = React.forwardRef(({ components = {}, ...props }, ref) => {
-	return <AsyncSelect ref={ref} {...props} components={{ ...avatarComponents }} />;
-});
-
 function useBody() {
 	const [body, setBody] = useState(null);
 	useEffect(() => {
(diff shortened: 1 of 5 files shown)


components/text-input-v2/select.jsx
```diff
@@ -282,6 +330,22 @@ export const AsyncSelect = React.forwardRef(({ components = {}, ...props }, ref)
 	);
 });
 
+export const AvatarSelect = React.forwardRef(({ components = {}, ...props }, ref) => {
```
korbinancell, line 333: Instead of re-wrapping and re-exporting out wrappers we should consider just exporting the `avatarComponents` and lettings consumers pass those in themselves. I am also in favor of just making these the default option and multi option components, because it seems that they won't break any existing selects.

cc @RobertBolender thoughts?

korbinancell (changes requested):

components/text-input-v2/select.jsx
korbinancell: Also thanks for your contribution to styled-ui 🎉

zmckinnon: I considered this as well.  It's more of a documentation thing if we go that direction.  Interested in more people's thoughts.

RobertBolender: For the internal `avatarComponents` themselves, these look beautifully clean and are an excellent use of our theme 👍 

For the shape of the exports, my preferences strongly align with the two options that Korbin recommended:
A: Export `avatarComponents` and let consumers pass them in to the various `components` props.
B: Make these `Option` and `MultiValue` components the defaults for all our selects (if we can verify this won't break existing consumers).

components/text-input-v2/select.jsx
```diff
@@ -282,6 +330,22 @@ export const AsyncSelect = React.forwardRef(({ components = {}, ...props }, ref)
...
+	return <Select ref={ref} {...props} components={{ ...avatarComponents }} />;
```
RobertBolender, line 334: If you do stick with this pattern of export, each of these components should spread their respective `components` props:
`components={{ ... avatarComponents, ...components }}`

RobertBolender (changes requested):

components/text-input-v2/select.jsx
zmckinnon: I did A.  I am tempted to do B, but it feels like it COULD be breaking if the data has an `avatar` property, so I left this out.  If we ever want to change this in the future, should be easy peasy.

RobertBolender (approved):

PR #386, opened 2020-11-18
Fix overriding of styles on disabled buttons

### LINKED ISSUE #382: [BUG] Cannot override styles on disabled button.
**Describe the bug**
The v6 `Button` does not let me override the disabled styles with styled-system props.

**Actual Behavior**

```
<Button
	disabled
	size="small"
	icon={<Checkmark color="#62bb46" />}
	backgroundColor="green1"
	color="foregroundPrimary"
	border="none"
>
	{localization.registered}
</Button>
```
The above markup produces a button that looks like this:
Screen Shot 2020-11-10 at 10 39 16 AM

**Expected behavior**

I would instead expect the button to look like the following.
Screen Shot 2020-11-10 at 10 37 39 AM

**Additional context**
I can work around this by using the non-v6 button, but non-v6 forces an 18px width and height on the icon. A CSS selector targeting the SVG or a `span` wrapping the SVG works around the 18px dimension problem. I'd love to instead avoid custom styles or wrapper hacks and use a disabled v6 `Button`.

tywmick:
Hello! First, a quick question: am I allowed to contribute to this repo without being a Faithlife employee? I ask because [`package.json`](https://github.com/Faithlife/styled-ui/blob/b554c0dd2447d8981fb3d36c0f9f67ecc1ee0d86/package.json#L6) says `"license": "UNLICENSED"`, and the reason I'm interested in this repo is because I'm applying for a job and wanted to start familiarizing myself with the codebase. I promise I won't use Faithlife Styled UI myself; I'm just browsing and committing for educational purposes. 😊

Anyway, if my contributing is okay, I believe I've fixed #382! Comparing the v6 `Button` to `LegacyButton` I saw that `LegacyButton` defined default styles for disabled buttons based on the `disabled` prop, not the `:disabled` CSS pseudo-class. What's happening in the v6 `Button` is that using the Styled System props only changes the properties of the main `button` selector, but if the button is disabled, the increased specificity of `button:disabled` causes it to still win out.

So this PR swaps the `:disabled` pseudo-class out for a `disabled` prop in the v6 `Button` styles.

Screen Shot 2020-11-16 at 10 43 21 PM

### DIFF
--- a/components/button/Button.js
+++ b/components/button/Button.js
@@ -7,18 +7,6 @@ import { common, typography } from '../../theme/system';
 import { buttonSizes, buttons } from '../../theme/buttons';
 import { theme } from '../../theme';
 
-const sizeVariant = variant({
-	prop: 'size',
-	scale: 'buttonSizes',
-	variants: buttonSizes,
-});
-
-const buttonVariant = variant({
-	prop: 'variant',
-	scale: 'buttons',
-	variants: buttons,
-});
-
 const ButtonCore = styled.button.attrs(({ active }) => ({ className: active ? 'active' : null }))`
 	box-sizing: border-box;
 	font-family: inherit;
@@ -55,10 +43,21 @@ const ButtonCore = styled.button.attrs(({ active }) => ({ className: active ? 'a
 		margin-right: ${props => (props.hasChildren ? props.theme.space[2] : '')};
 	}
 
-	${sizeVariant}
-	${buttonVariant}
-	${textStyle};
+	${variant({
+		prop: 'size',
+		scale: 'buttonSizes',
+		variants: buttonSizes,
+	})};
 
+	${({ disabled }) => css`
+		${variant({
+			prop: 'variant',
+			scale: 'buttons',
+			variants: buttons(disabled),
+		})};
+	`}
+
+	${textStyle};
 	${common};
 	${typography};
 	${layout};
--- a/theme/buttons.js
+++ b/theme/buttons.js
@@ -21,152 +21,178 @@ export const buttonSizes = {
 	},
 };
 
-export const buttons = {
-	primary: {
-		color: 'button.primaryForeground',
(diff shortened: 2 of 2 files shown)


korbinancell:
Hello @tywmick, great idea and good catch. Pr's are welcome, but because of the project being unlicensed and not having a contributors guide we will need to hold off on merging until we sort those out.
In the meantime, I can take a look at your pr and see about getting some code review going.

bryanrsmith:
Hi @tywmick,

Thanks for the PR! I agree that this resolves the issue as reported, but I actually think the existing behavior is preferable. 

* There are other button states that exhibit this issue in addition to `:disabled`, such as `:hover`. We could take a similar approach for the other states as well, but it would add complexity, and possibly cause styled components to generate a lot more classes.
* Allowing style props to take precedence over the default styling for these states transfers the burden of managing the states to the consuming code, making it easy to create buttons that don't behave as expected in certain states. If a button wants custom styles for the disabled case, but not for the enabled case, they would have to manually manage all the style props relative to the button state.
* I don't think this is a common need (if it were, we would add variants to cover those use cases), and while it's a little unfortunate that we don't have a convenient mechanism for targeting pseudo classes via normal props, I'm happy for this use case to be just slightly harder to accommodate. Developers that need to change the disabled styling can already customize it with the `css` prop, or by wrapping `Button` in their own styled-component with styled-system variants.

Happy to discuss further if anyone disagrees.

tywmick:
@bryanrsmith Totally makes sense! I honestly like the existing behavior better myself, too. 👍🏼

bryanrsmith:
Sounds good. Without objection, I'll go ahead and close this PR as well as the issue.

@tywmick despite the decision we reached, I do appreciate the clear communication and generally solid PR. Thank you!

tywmick:
@bryanrsmith My pleasure!

PR #388, opened 2020-11-20
Add a focus style to Checkbox

### LINKED ISSUE #371: [BUG] Checkboxes are missing a focus state
**Expected behavior**
When using the tab key to focus a checkbox, there should be a visible indication of where the focus is.

Spec: https://app.zeplin.io/project/58a23c2d4597afb248c13a43/screen/5eaca8f4fd3f6c19c9331e91

tywmick:
I don't have access to the spec referenced in the issue, but how does this look to fix #371?

I grabbed the border color and box shadow straight [from the `Input` component](https://github.com/Faithlife/styled-ui/blob/6043ef87d8d5df05f63915ac24c668abdf22c98f/components/input/Input.jsx#L151-L153). Here's the new focus section of the checkbox styles:

```
&:focus {
	outline: none;

	& ${CheckboxDiv} {
		border-color: ${({ theme }) => theme.colors.input.borderFocused};
		box-shadow: 0 0 0 2px ${({ theme }) => theme.colors.input.shadowFocused};
	}
}
```

And how it looks on the page:

Checkbox in new focus state

Should these values be customized for checkboxes at all, or does it look good as is? If it's fine to leave as is, should the checkbox styles continue referencing [`theme.colors.input`](https://github.com/Faithlife/styled-ui/blob/6043ef87d8d5df05f63915ac24c668abdf22c98f/theme/core.js#L157-L167) directly, or should I create new `borderFocused` and `shadowFocused` fields in [`theme.colors.checkbox`](https://github.com/Faithlife/styled-ui/blob/6043ef87d8d5df05f63915ac24c668abdf22c98f/theme/core.js#L194-L200)?

### DIFF
--- a/components/check-box/styled.jsx
+++ b/components/check-box/styled.jsx
@@ -55,6 +55,11 @@ export const CheckboxContainer = styled.button`
 
 	&:focus {
 		outline: none;
+
+		& ${CheckboxDiv} {
+			border-color: ${({ theme }) => theme.colors.input.borderFocused};
+			box-shadow: 0 0 0 2px ${({ theme }) => theme.colors.input.shadowFocused};
+		}
 	}
 `;
 
--- a/components/check-box/styled.jsx
+++ b/components/check-box/styled.jsx
@@ -57,8 +57,8 @@ export const CheckboxContainer = styled.button`
 		outline: none;
 
 		& ${CheckboxDiv} {
-			border-color: ${({ theme }) => theme.colors.input.borderFocused};
-			box-shadow: 0 0 0 2px ${({ theme }) => theme.colors.input.shadowFocused};
+			border-color: ${({ theme }) => theme.colors.checkbox.borderFocused};
+			box-shadow: 0 0 0 2px ${({ theme }) => theme.colors.checkbox.shadowFocused};
 		}
 	}
 `;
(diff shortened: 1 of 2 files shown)


rmjohnson:
>or should I create new borderFocused and shadowFocused fields in theme.colors.checkbox?

@tywmick this would be my assumption but I'll defer to @korbinancell and @RobertBolender

korbinancell (approved):
Good call on adding it to the theme.

tywmick:
Mostly what I was concerned about with copying `borderFocused` and `shadowFocused` to `theme.colors.checkbox` is that in `theme.colors.input`, those two fields are given unique color codes rather than one of the standard blues defined earlier in the file:

https://github.com/Faithlife/styled-ui/blob/6043ef87d8d5df05f63915ac24c668abdf22c98f/theme/core.js#L157-L167

Which isn't bad in itself, but then I thought that if we want the checkbox colors to remain harmonious with the input colors, if we copy those string values over to `theme.colors.checkbox`, I could envision a scenario in which someone changes the input colors but (completely understandably) doesn't notice that the checkbox colors have the same values, so they leave the checkbox colors alone. So here's the main change with the latest commit I pushed:

```js
colors.checkbox = {
	...
	borderFocused: colors.input.borderFocused,
	shadowFocused: colors.input.shadowFocused,
};
```

That way, no one has to remember to change both values, and _[oh I guess I can stop explaining it now that you've merged 😆]_

PR #389, opened 2020-11-21
Retrieve selected date period with DatePeriodPicker callback

### LINKED ISSUE #199: DatePeriodPicker should allow tokens on datePeriods
Currently when consuming `DatePeriodPicker`, if you want to pass in `datePeriods`, and then respect the correct `displayName` when `setSelectedDate`, you have to do a lookup on the `start` and `end` dates passed into `setSelectedDate`. e.g.
```js
let newDatePeriod: IDatePeriod = Object.values(datePeriods).find(
	dp =>
		(dateFunctions.isSameDay(dateRange.start, dp.dateRange.start) &&
			dateFunctions.isSameDay(dateRange.end, dp.dateRange.end)) ||
		(dateRange.start === dp.dateRange.start && dateRange.end === dp.dateRange.end)
);
```
It would be nice if instead the callback function when a date period is picked passed back a token that so we could instead do something like:
```js
let newDatePeriod: IDatePeriod = datePeriods[datePeriodToken];
```

This wouldn't catch the "didn't select a date period but happened to be the same date range selected as a date period" case (that I'm catching above with the `isSameDay` check) but that feels like an acceptable trade off.

[Original discussion that prompted this](https://git.faithlife.dev/Logos/ChurchManagement/pull/35/files#r139854)

tywmick:
I have an idea that might close #199. The impression I got from what @rmjohnson wrote is that it'd be convenient if `DatePeriodPicker`'s `setSelectedDate` could have access to the display name of a selected date period, so how does adding a second argument to that callback sound?

I changed the `onClick` prop in `Styled.DatePeriod` to...

```jsx
onClick={() => {
  setSelectedDate(datePeriod.dateRange, datePeriod.displayName);
}}
```

...and now you can add a second argument to your `setSelectedDate` function to capture that display name, like this:

```jsx
setSelectedDate={(dateRange, displayName) => {
  setState({ selectedDateRange: dateRange, selectedDatePeriodName: displayName })
}}
```

And this change won't break anything, because any existing callbacks with one argument won't care if you pass them a second argument—they'll just ignore it.

Here's a demo of [the new version in action](https://tymick.me/styled-ui/#/date-picker/variations?a=date-period-picker-two-argument-callback), and here's a demo of it [with a one-argument callback](https://tymick.me/styled-ui/#/date-picker/variations?a=date-period-picker-one-argument-callback) passed in, to show that it doesn't break.

-----

I also made a big change to the prop description, since when I was first looking at the component, it took me a while to understand that `setSelectedDate` was a callback used to retrieve the selected date when it has been set, not a function to set which date is selected. Here's the new description:

> A callback that retrieves the currently selected date range and (optionally) the display name of the selected date period whenever the the selected dates change.

That may only make things clearer in my own head, though, so let me know if that's too drastic a change and I can scale it back.

### DIFF
--- a/components/date-period-picker/component.jsx
+++ b/components/date-period-picker/component.jsx
@@ -30,7 +30,7 @@ export class DatePeriodPicker extends PureComponent {
 		}),
 		/** Function to parse a date of format M/d/yyyy into a date object. See https://date-fns.org/v2.0.0-alpha.25/docs/parse for details */
 		parseDate: PropTypes.func.isRequired,
-		/** A callback that retrieves the currently selected date range and (optionally) the display name of the selected date period whenever the the selected dates change. */
+		/** A callback that retrieves the currently selected date range and (optionally) the index of the selected date period whenever the the selected dates change. */
 		setSelectedDate: PropTypes.func.isRequired,
 		/** Takes a date as a parameter and returns false if that date is invalid */
 		validate: PropTypes.func,
--- a/components/date-period-picker/component.jsx
+++ b/components/date-period-picker/component.jsx
@@ -30,7 +30,7 @@ export class DatePeriodPicker extends PureComponent {
 		}),
 		/** Function to parse a date of format M/d/yyyy into a date object. See https://date-fns.org/v2.0.0-alpha.25/docs/parse for details */
 		parseDate: PropTypes.func.isRequired,
-		/** A callback accepting the currently selected date range and (optionally) the display name of the selected date period as arguments, which will be called whenever the the selected dates change. */
+		/** A callback that retrieves the currently selected date range and (optionally) the display name of the selected date period whenever the the selected dates change. */
 		setSelectedDate: PropTypes.func.isRequired,
 		/** Takes a date as a parameter and returns false if that date is invalid */
 		validate: PropTypes.func,
--- a/components/date-period-picker/component.jsx
+++ b/components/date-period-picker/component.jsx
@@ -144,7 +144,7 @@ export class DatePeriodPicker extends PureComponent {
 			<Styled.Container>
 				{datePeriods.map((datePeriod, index) => (
 					<Styled.DatePeriod
-						key={index}
+						key={datePeriod.displayName}
 						onClick={() => {
 							setSelectedDate(datePeriod.dateRange, index);
 						}}
--- a/components/date-period-picker/component.jsx
+++ b/components/date-period-picker/component.jsx
@@ -104,13 +104,33 @@ export class DatePeriodPicker extends PureComponent {
 				this.setState({ inputValues: { [input]: value } });
 			}
 
-			this.props.setSelectedDate(selectedDate, null);
+			this.props.setSelectedDate(selectedDate, this.getDatePeriodIndex(selectedDate));
 		}
 	}, this.props.debounce);
 
 	handleInputValueChange = (value, input) => {
 		this.setState(state => ({ inputValues: { ...state.inputValues, [input]: value } }));
 		this.parseAndUpdateDate(value, input);
+	};
+
+	getDatePeriodIndex = ({ start, end }) => {
+		if (start === undefined || end === undefined || start === null || end === null) {
+			return null;
+		}
+
+		for (let i = 0; i < this.props.datePeriods.length; i++) {
+			if (
+				isSameDay(start, this.props.datePeriods[i].dateRange.start) &&
+				isSameDay(end, this.props.datePeriods[i].dateRange.end)
+			) {
+				return i;
+			}
+		}
+		return null;
+
+		function isSameDay(date1, date2) {
+			return new Date(date1).setHours(0, 0, 0, 0) === new Date(date2).setHours(0, 0, 0, 0);
+		}
 	};
 
 	render() {
@@ -156,7 +176,9 @@ export class DatePeriodPicker extends PureComponent {
 					<DatePicker
 						asDateRangePicker
 						selectedDateRange={selectedDateRange}
-						setSelectedDate={dateRange => setSelectedDate(dateRange, null)}
+						setSelectedDate={dateRange =>
+							setSelectedDate(dateRange, this.getDatePeriodIndex(dateRange))
+						}
 						validate={validate}
 						dateFunctions={dateFunctions}
 					/>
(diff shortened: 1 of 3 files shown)


tywmick:
Also, I'd love to add logic to catch that "didn't select a date period but happened to be the same date range selected as a date period" case that @rmjohnson talked about. Would that be valuable, or would it be best to only consider a named date period to have been selected when the name itself is clicked?

tywmick:
Also, if that new prop description is good, should I go ahead and change [the one in `DatePicker`](https://github.com/Faithlife/styled-ui/blob/6043ef87d8d5df05f63915ac24c668abdf22c98f/components/date-picker/component.jsx#L49-L50) to match?

rmjohnson:
This looks great to me. I'll defer to @korbinancell and @RobertBolender for approval.

components/date-period-picker/component.jsx
```diff
@@ -126,7 +126,7 @@ export class DatePeriodPicker extends PureComponent {
 					<Styled.DatePeriod
 						key={datePeriod.displayName}
 						onClick={() => {
-							setSelectedDate(datePeriod.dateRange);
+							setSelectedDate(datePeriod.dateRange, datePeriod.displayName);
```
korbinancell, line 129: I like the idea of returning a key to set apart the date ranges, but using localized text (`datePeriod.displayName`) can lead to issues when switching culture/ language and should generally be avoided. I propose adding another property to the `datePeriods` list called something like `key` or `periodKey` that we pass back to the consumer.

korbinancell (commented):
> Also, I'd love to add logic to catch that "didn't...

My only issue here is it's hard to be sure what date functions you are going to be passed. As long as it is null safe this sounds good to me. I assume you can just assume that all date-fns functions will be available as a start.

> And if that new prop description is good, should I go ahead...

SGTM

components/date-period-picker/component.jsx
tywmick: Ooh, what about just adding the index to the map callback like this?

```js
{datePeriods.map((datePeriod, index) => (
  <Styled.DatePeriod
    key={index}
    onClick={() => {
      setSelectedDate(datePeriod.dateRange, index);
    }}
  >
    {datePeriod.displayName}
  </Styled.DatePeriod>
))}
```

That way the consumer can use that index directly on the `datePeriods` array they already have, and we don't have to mutate the `datePeriods` prop itself.

tywmick:
So since the [`dateFunctionProps`](https://github.com/Faithlife/styled-ui/blob/6043ef87d8d5df05f63915ac24c668abdf22c98f/components/date-picker/date-function-props.js) type doesn't currently require an `isSameDay` function, I decided to roll my own with vanilla JS `Date` methods. Is that a good idea, or should we add an `isSameDay` requirement to `dateFunctionProps`? Would changing a prop type in that way be a breaking change?

Here's the new `getDatePeriodIndex` method:

```js
getDatePeriodIndex = ({ start, end }) => {
  if (start === undefined || end === undefined || start === null || end === null) {
    return null;
  }

  for (let i = 0; i < this.props.datePeriods.length; i++) {
    if (
      isSameDay(start, this.props.datePeriods[i].dateRange.start) &&
      isSameDay(end, this.props.datePeriods[i].dateRange.end)
    ) {
      return i;
    }
  }
  return null;

  function isSameDay(date1, date2) {
    return new Date(date1).setHours(0, 0, 0, 0) === new Date(date2).setHours(0, 0, 0, 0);
  }
};
```

And here's the [demo](https://tymick.me/styled-ui/#/date-picker/variations?a=date-period-picker-two-argument-callback) if you want to see it in action. 👍🏼

components/date-period-picker/component.jsx
```diff
@@ -122,11 +142,11 @@ export class DatePeriodPicker extends PureComponent {
 
 		return (
 			<Styled.Container>
-				{datePeriods.map(datePeriod => (
+				{datePeriods.map((datePeriod, index) => (
 					<Styled.DatePeriod
-						key={datePeriod.displayName}
+						key={index}
```
RobertBolender, line 147: Using indexes as keys can have undesired side effects.
https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318
Was `datePeriod.displayName` not guaranteed to be unique? If it is, let's continue to use it. If not, we can *add* the index to the end of the string.

RobertBolender, line 147:
```suggestion
						key={datePeriod.displayName}
```

tywmick: Oh, I'd never thought about what would happen if a list needs to be able to change after rendering—that _is_ a habit I should get out of. Thanks for the info!

The only situation I can see in which `datePeriod.displayName` would _not_ be unique is if the dev made a mistake, because seeing two identically labeled buttons has got to be confusing to a user (and unnecessary here).

RobertBolender (approved):

components/date-period-picker/component.jsx
rmjohnson: >The only situation I can see in which datePeriod.displayName would not be unique is if the dev made a mistake, because seeing two identically labeled buttons has got to be confusing to a user (and unnecessary here).

@RobertBolender, @tywmick, to this end, are you opposed to adding some validation to this component that the consumer did *not* provide duplicate display names? I'd rather us fail fast for a developer with a clear error message than them run into unexpected behavior because of duplicate `key`s.

tywmick:
(I thought the "close" button on the Android app just now was for closing a particular navigation pane, not closing the PR. 🤦🏼‍♂️ Oops.)

components/date-period-picker/component.jsx
tywmick: Ooh! Yeah, one thing I meant to say in my previous comment but forgot was that one advantage of using the display name as the key would be that React would send a key warning if someone used duplicate display names. But you're right, @rmjohnson, a more descriptive error message would be much more useful.

What would be best here, a console warning, console error, or throwing an error in development?

RobertBolender: In this case, the duplicate keys will cause a clear (enough) warning message for us.
I'd be open to additional validation, but would be fine considering that out of scope for this PR and merging this one as-is.

My one hesitation *away* from failing fast in this case is my hypothetical case where a runtime-generated list *doesn't* fail for the developer but fails for the user, where we might end up showing them confusing results that *might* "just work" if we hadn't interrupted their experience with an error message. It's hard to evaluate this scenario without something more concrete.

How would you feel about a filtering the list for duplicates, keeping only the first for each display string, and logging a warning/error for any duplicates? (Not blocking for this PR, it can be a follow-up)

tywmick: > How would you feel about a filtering the list for duplicates, keeping only the first for each display string, and logging a warning/error for any duplicates? (Not blocking for this PR, it can be a follow-up)

I like it! 👌🏼

PR #390, opened 2020-11-24
Add "loading" prop to Button

EthanRutherford:
Adds a `loading` prop to Buttons which A. hides the content and shows a loading spinner instead, and B. disables the Button.
Spec: https://app.zeplin.io/project/58a23c2d4597afb248c13a43/screen/5fb5971e1de8db803db4be02

Possible breaking changes:
I had to add `position: relative` to the button. I'm decently confident this won't be an issue, as I can't think of any scenario where a child of a button would be positioned relative to some *ancestor* of the button. Buttons shouldn't typically have particularly complex children anyway, as any child of the button would be part of the click target of the button, so I really don't think there are any reasonable use cases for absolutely positioned children to "escape" the button.

I also had to add a wrapper div around the `{children}` of the button, so that I could do `visibility: hidden` on it. `children` *can* be just a plain string, which would be rendered as a text node, and thus not be styleable without a wrapper element. I don't expect this to break anyone, but it's worth noting.

### DIFF
--- a/catalog/button/variations-v6.md
+++ b/catalog/button/variations-v6.md
@@ -9,34 +9,32 @@ Buttons allow users to command the computer to take some action. Buttons, like a
 ```react
 showSource: true
 ---
-<div>
-	<ButtonDemo>
-		<Button variant="primary" size="medium">
-			Primary
-		</Button>
-		<Button variant="secondary" size="medium">
-			Secondary
-		</Button>
-		<Button variant="minor" size="medium">
-			Minor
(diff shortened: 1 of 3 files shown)


RobertBolender:
> I also had to add a wrapper div around the {children} of the button, so that I could do visibility: hidden on it.

I'm trying to track it down in the history, but I believe this *has* been an unexpected breaking change in the past.

RobertBolender (approved):

Issue #378, opened 2020-10-12
v6 Roadmap


RobertBolender:
## v6 goals

- [x] Can be installed cleanly on top of v5
- [x] Can have multiple versions installed without conflicts (no global CSS files, etc)
- [x] Should be installed as a direct dependency rather than a peer dependency

## v6 roadmap

- [x] Refactor Listbox/Dropdown components to Listbox/Menu/Dropdown (see https://github.com/Faithlife/styled-ui/issues/411 @korbinancell)
- [x] Remove all references to any `styleOverrides` props, replacing the functionality with styled-system props
  - [x] Accordion (https://github.com/Faithlife/styled-ui/pull/391)
  - [x] DatePickerInput (https://github.com/Faithlife/styled-ui/pull/392)
  - [x] Dropdown (https://github.com/Faithlife/styled-ui/pull/379)
  - [x] FilesSection (https://github.com/Faithlife/styled-ui/pull/395)
  - [x] Input (https://github.com/Faithlife/styled-ui/pull/392)
  - [x] Listbox (https://github.com/Faithlife/styled-ui/pull/432, see also https://github.com/Faithlife/styled-ui/issues/411)
  - [x] ModalBackdrop (only used internally, not exported)
  - [x] ParameterSentence (https://github.com/Faithlife/styled-ui/pull/410)
  - [x] SimpleToast (https://github.com/Faithlife/styled-ui/pull/400)
  - [x] Slider (https://github.com/Faithlife/styled-ui/pull/404)
  - [x] Tab? (https://github.com/Faithlife/styled-ui/pull/409)
- [x] Remove all references to legacy `theme` props, replacing the functionality with our global styled-system theme
  - [x] Checkbox (https://github.com/Faithlife/styled-ui/pull/412)
  - [x] Helpbox (https://github.com/Faithlife/styled-ui/pull/414)
  - [x] LoadingSpinner (https://github.com/Faithlife/styled-ui/pull/415)
  - [x] Radio (https://github.com/Faithlife/styled-ui/pull/416)
  - [x] SimpleToast (https://github.com/Faithlife/styled-ui/pull/400)
- [x] Consider updating exported API for parent/child components
  Example pattern: `<Modal>`, `<Modal.Content>` instead of `<Modal>` and `<ModalContent>`
  - [x] Tabs (https://github.com/Faithlife/styled-ui/pull/417)
- [x] Consider using the ChildConfig API pattern for these components:
[...]

PR #391, opened 2020-11-27
Create v6 Accordion with Styled System

tywmick:
As a step toward the v6 release (#378), how's this look for replacing the `styleOverrides` prop in `Accordion`?

The only property in `styleOverrides` used in `Accordion` was `panelPadding`, which set `padding` on all nested `Accordion.Panel`s, so here are the main replacements:

1. Added Styled System's [variant API](https://styled-system.com/variants) to `Accordion.Panel` to define the default and `'minimal'` padding styles, then added the [`space`](https://styled-system.com/table#space) field to allow overriding `padding` on individual panels.

   ```js
   export const Panel = styled(Box).attrs(({ headerId, panelId }) => ({
     role: 'region',
     'aria-labelledby': `accordion-header-${headerId}`,
     id: `accordion-panel-${panelId}`,
   }))(
     variant({
       variants: {
         default: {
           padding: 6,
           paddingTop: 4,
         },
         minimal: {
           padding: 4,
           paddingTop: 0,
         },
       },
     }),
     space,
   );
   ```

2. Added a [custom](https://styled-system.com/custom-props) `panelPadding` prop to `Accordion`, applying its effects [only to nested](https://styled-components.com/docs/advanced#referring-to-other-components) `Panel`s from `Accordion.Panel`. This way it's still possible to set the `padding` of all panels at the `Accordion` level instead of having to set each `Accordion.Panel`'s `padding` separately.

   ```js
   const AccordionBox = styled(Box)`
     ${Panel} {
       ${system({
         panelPadding: {
           property: 'padding',
           scale: 'space',
         },
       })}
     }
   `;
   ```

-----

Should this PR target a new branch instead of `master` since it's a breaking change, or will v6 be the next release (so it doesn't matter)?


tywmick:
Come to think of it, should I be creating new files for these new v6 versions with breaking changes (this and #392 so far) instead of contemplating a new branch? I just reread the whole "December 18th - all updated components will be available at the /v6 import" part on #378.

RobertBolender:
> Come to think of it, should I be creating new files for these new v6 versions with breaking changes (this and #392 so far) instead of contemplating a new branch? I just reread the whole "December 18th - all updated components will be available at the /v6 import" part on #378.

The pattern we've been using is to add a new file and export it at the /v6 entrypoint. While we will be making the v6 breaking change soon, we want to keep the v5 versions free of breaking changes until that point so that we're always in a shippable state.

Examples of recent v6 additions:
[Popover](https://github.com/Faithlife/styled-ui/pull/338/files#diff-1e4450d023992ca245ef5d344ff3b2a9898701b8f62367b10abf6449dc76bb99R3)
[Dropdown](https://github.com/Faithlife/styled-ui/pull/379/files#diff-1e4450d023992ca245ef5d344ff3b2a9898701b8f62367b10abf6449dc76bb99R4)

tywmick:
I went with the "legacy" naming scheme, renaming old files to `legacy-*` and exporting the v5 version from `components/accordion` as `LegacyAccordion`.

How's this look?

components/accordion/accordion.jsx
```diff
@@ -89,12 +90,20 @@ Accordion.propTypes = {
...
+	panelPadding: styledPropType,
+	...Box.propTypes,
 };
 
-Accordion.defaultProps = {
-	expandedSections: [],
-	styleOverrides: {},
-};
+const AccordionBox = styled(Box)`
+	${Panel} {
```
RobertBolender, line 101: This is clever, I like it 👍

tywmick: Thanks! 😊

PR #392, opened 2020-11-27
Update Input and DatePickerInput for v6

tywmick:
Another step in the v6 roadmap (#378).

`styleOverrides` didn't actually need to be _replaced_ in this one, just removed, because `StyledInput` already uses Styled System's [`layout`](https://styled-system.com/table#layout) function. `width` and `height` props are ready to go. 👍🏼

-----

As with #391, should this breaking change PR target a new branch or stick with `master`?


tywmick:
Oops, I just realized I need to change where other components use `Input`'s `styleOverrides` prop. Better make this a draft in the meantime.

tywmick:
Looks like the only component that was using `Input`'s `styleOverrides` was `DatePickerInput`. That component was using the following `styleOverrides`:

- `inputWidth` – passed to `Input`
- `inputBorderColor` – passed to `Input`
- `hideShadow` – used by `Popover`
- `width` – passed to `Popover`
- `padding` – passed to `Popover`
- `border` – passed to `Popover`
- `zIndex` – passed to `Popover`

I decided to create separate props for each of these and pass them explicitly to `Input` or `Popover` accordingly. Conveniently, `Popover` needs no changes, because `DatePickerInput` already imports the v6 `Popover`, which doesn't accept a `styleOverrides` prop.

One question: now that both `Input` and `Popover` use Styled System props, should the names of the `hideShadow`, `width`, `padding`, `border`, and `zIndex` props be changed to include "popover" in the name? I ask because extra props are being passed to `Input` by default, so it may be confusing that those five are sent to `Popover`, not `Input`.

tywmick:
Since `Input` didn't need any new props added (Styled System props were already enabled alongside `styleOverrides`), I decided not to make a separate v6 export, instead just adding a deprecation warning when `styleOverrides` is used in a production environment. And I added some `@todo` comments to remind us to remove `styleOverrides` when v6 is released.

If a separate v6 `Input` export would be better, though, I can switch things around. 👍🏼

I did make a separate v6 `DatePickerInput` export, and while I was working on that, I added a few more changes to the new v6 component:

1. Adjusted the padding on the calendar button so that when the user clicks above, below, or to the right of the calendar icon but still within the input field, it activates the calendar button instead of selecting the input itself.
2. Changed the popover reference from the whole button to just the icon container, so it will be centered correctly if the developer specifies a centered `placement`.
3. Renamed the props to focus on the input instead of the popover. Previously, input-specific style props included "input" in at the beginning of the name, while popover-specific props just used the basic style prop names. The input field feels like the main "thing" in this component, though, and excess props are already being sent to it, so I thought the following prop name changes would make their purposes more intuitive:
   - `inputWidth` → `width`
   - `inputBorderColor` → `borderColor`
   - `placement` → `calendarPlacement`
   - `hideShadow` → `hideCalendarShadow`
   - `width` → `calendarWidth`
   - `padding` → `calendarPadding`
   - `border` → `calendarBorder`
   - `zIndex` → `calendarZIndex`

tywmick:
Fixed the bug part of #401, where `DatePickerInput`'s calendar select popover was rejecting _all_ selected dates if no `validate` prop was passed.

tywmick:
Oh! I just realized that I don't need those explicit input- and popover-specific style props: that's what the child config components are for. 🤦🏼‍♂️ I've removed the following props:

- `calendarPlacement`
- `hideCalendarShadow`
- `calendarWidth`
- `calendarPadding`
- `calendarBorder`
- `calendarZIndex`

components/date-picker-input/component.jsx
```diff
@@ -101,73 +101,63 @@ export function DatePickerInput({
...
 }
 
 DatePickerInput.defaultProps = {
-	placement: 'bottom-start',
-	styleOverrides: {
-		inputWidth: '100%',
-		padding: '16px 20px',
-		zIndex: 3,
-	},
+	width: '100%',
```
RobertBolender, line 160: Consider specifying the default `width` in the destructured props argument for the component and removing this `defaultProps` section entirely.

PR #394, opened 2020-11-28
Remove conflicting v6 Button size prop

### LINKED ISSUE #393: [BUG] Cannot set v6 Button width or height via props
### Describe the bug
When trying to use the `height` prop on a v6 `Button` (during #395), I discovered that Styled System's [layout utility](https://styled-system.com/table#layout) includes a shortcut prop called `size`, which sets both the `width` and `height` properties. The weird thing is that since the styles for `ButtonCore` [include both `sizeVariant` and `layout`](https://github.com/Faithlife/styled-ui/blob/eee81d8fdb743a47cfa32e727f41428cdcf264c4/components/button/Button.js#L70-L76), `Button`'s `size` prop sets both the size variant _and_ the CSS width and height. I would have thought that defining `size` as the prop name for a custom variant would override the layout utility's `size`, but apparently that isn't the case.

Anyway, since the default value of `ButtonCore`'s `size` prop is `'medium'`, this adds `width: medium; height: medium;` CSS to every default v6 `Button`. It also prevents setting `Button`'s width and height via `width` and `height` props—the `size` prop wins out here.

### To Reproduce
1. Use a v6 `Button`.
2. Try to use its `width` and `height` props.
3. Open the element inspector in the web dev tools in your browser.
4. Notice that the CSS `width` and `height` properties of the `button` in the DOM have the value `medium` (or another other value you passed to the `size` prop), not the values you passed to the `width` and `height` props.

### Expected behavior
The values passed to the `width` and `height` props should be reflected in the CSS in the browser.

### Screenshots
Button CSS issues

### Desktop (please complete the following information):
- OS: macOS 10.15.7 (19H2)
- Browsers
  - Firefox 83.0 (64-bit)
  - Chromium 86.0.4240.193 (Official Build) (x86_64)
  - Safari 14.0 (15610.1.28.1.9, 15610)

### Additional context
I've opened an issue in Styled System (styled-system/styled-system#1728) to ask if this is a bug or if this behavior is intended, but in the meantime, I see two alternate ways of fixing this:

1. The alternative implemented in #394, creating a [custom style function](https://styled-system.com/api#system) that adds every `layout` prop except `size`, so `Button`'s `size` prop won't block its `width` and `height`.
2. Changing the name of the size variant prop to something else like `sizeVariant` instead of just `size`. This will probably break more people's current usage of the v6 `Button`, though.

tywmick:
Fixes #393 by creating a [custom style function](https://styled-system.com/api#system) that adds every `layout` prop except `size`, so `Button`'s `size` prop won't block its `width` and `height`.

### DIFF
--- a/components/button/Button.js
+++ b/components/button/Button.js
@@ -69,15 +69,15 @@ const ButtonCore = styled.button.attrs(({ active }) => ({ className: active ? 'a
 
 	${sizeVariant}
 	${buttonVariant}
-	${textStyle};
+	${textStyle}
 
-	${common};
-	${typography};
-	${layout};
-	${flexbox};
-	${position};
-	${border};
-	${background};
+	${common}
+	${typography}
+	${layout}
+	${flexbox}
+	${position}
+	${border}
+	${background}
 `;
 
 ButtonCore.defaultProps = {
--- a/components/button/Button.js
+++ b/components/button/Button.js
@@ -1,12 +1,13 @@
 import React from 'react';
 import styled, { css } from 'styled-components';
-import { variant, layout, flexbox, position, textStyle, border, background } from 'styled-system';
+import { variant, flexbox, position, textStyle, border, background } from 'styled-system';
 import 'focus-visible';
 import { Box } from '../Box';
 import { common, typography } from '../../theme/system';
 import { buttonSizes, buttons } from '../../theme/buttons';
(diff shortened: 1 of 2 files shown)


RobertBolender:
For now, let's prefer the currently available workaround of manually specifying the `size` prop before either `height` or `width`, as described in the comment on issue #393

PR #395, opened 2020-11-28
Switch to v6 Button in FilesSection

tywmick:
Continuing down the #378 road, this removes use of the `styleOverrides` prop in `FilesSection` by switching to using v6 `Button`s. Correct rendering does depend on the resolution of #393, though, so I'll leave this as a draft for now.

Turns out the v6 equivalent of v5's `<Button variant="primaryTransparent" />` is `<Button variant="link" height="unset" padding={0} />`.


tywmick:
Implemented **@&zwnj;RobertBolender**]'s [workaround](https://github.com/Faithlife/styled-ui/issues/393#issuecomment-735931695) for #393, so everything's rendering correctly now. 👍🏼

tywmick:
Gonna put this back in draft status for #398—if that one's accepted, I'll just revert d0b2c8d.

RobertBolender:
Rather than needing to specify `height="unset" padding={0}` on every usage of `variant=link`, I'd like to explore adding those into the `link` variant itself. This is technically a breaking change, but v6 is the perfect time to do that.

components/button/Button.js
```diff
@@ -54,18 +54,17 @@ const ButtonCore = styled.button.attrs(({ active }) => ({ className: active ? 'a
...
-		isLoading
-			? css`
-					color: transparent !important;
-					& > :not(:first-child) {
-						visibility: hidden;
-					}
-			  `
-			: ''}
+	${({ loading }) =>
+		loading &&
```
RobertBolender, line 61:
```suggestion
	${({ isLoading }) =>
		isLoading &&
```

See: https://github.com/Faithlife/styled-ui/commit/2a5aa6623062eea7d78b2820246874f648e469ab

components/files-section/file-item.jsx
```diff
@@ -71,7 +66,7 @@ export class FileItem extends PureComponent {
 		return (
 			<Styled.FileItem>
 				{onFileClicked ? (
-					<Button onClick={this.handleFileClick} variant="primaryTransparent">
+					<Button onClick={this.handleFileClick} variant="link" height="unset" padding={0}>
```
RobertBolender, line 69:
```suggestion
					<Button onClick={this.handleFileClick} variant="link">
```

components/button/Button.js
tywmick: Whoops, thanks!

RobertBolender (approved):
Looks good to me!
I'll wait to merge this one until we're ready for the v6 breaking change, which should be coming shortly.

RobertBolender:
I've rebased your commits and cherry-picked them onto the v6 PR, so I'll close this one and your commits will get merged in when we're ready for v6.
https://github.com/Faithlife/styled-ui/pull/413/files/ac666c66f8ce54c4f251a6369b2d8a388b6c3d1b..0a1fc6ca5a77f2db4e4496ba39003d5cb35c3790

PR #396, opened 2020-11-29
Remove styleOverrides prop from ModalBackdrop

tywmick:
Continuing down #378, this removes the `styleOverrides` prop from both the class component `ModalBackdrop` and its styled-component sibling, `Backdrop`. The only component currently using either was the [v6 `Modal`](https://github.com/Faithlife/styled-ui/blob/eee81d8fdb743a47cfa32e727f41428cdcf264c4/components/modal/modal.jsx), using the class component, and it wasn't even using the `styleOverrides` prop, just using a `zIndex` prop directly, which `ModalBackdrop` passed straight on to `Box`.

I didn't actually add any Styled System props to either of these; `zIndex` was the only property in use, so I just let the class component pass it on to `Box` and the other component use it to define `z-index` directly in its styled-components style. Should I add other Styled System props?

Also, would it be helpful for me to add the remaining logic from the class component to the styled-component version so the styled-component version can take over the job? I'd be happy to open up a new PR for that if that's the eventual plan anyway.


components/modal-backdrop/styled.jsx
```diff
@@ -11,5 +11,10 @@ export const Backdrop = styled.div`
 	justify-content: center;
 	align-items: center;
 	overflow: hidden;
-	z-index: ${props => props.styleOverrides.zIndex};
+	${({ zIndex }) =>
```
korbinancell, line 14: Looks like Backdrop isn't even used so this whole file can be deleted instead.

korbinancell (changes requested):

PR #398, opened 2020-12-01
Work around v6 Button size prop bug

tywmick:
I think this is just a temporary workaround for #393, but I thought of a way to force `ButtonCore` to define `size` before `width` or `height`! And ensure that an attempt to use the Styled System [layout utility](https://styled-system.com/table#layout) `size` doesn't sabotage the size variant logic.

How's this look for the time being?

```js
const Button = React.forwardRef(
  ({ children, icon, disabled, loading, size, width, height, ...props }, ref) => {
    // Temporary workaround for #393
    let sizeVariant = ButtonCore.defaultProps.size;
    let widthStyle = width;
    let heightStyle = height;
    if (Object.keys(buttonSizes).includes(size)) {
      sizeVariant = size;
    } else if (size !== undefined) {
      if (width === undefined) {
        widthStyle = size;
      }
      if (height === undefined) {
        heightStyle = size;
      }
    }

    return (
      <ButtonCore
        ref={ref}
        size={sizeVariant}
        width={widthStyle}
        height={heightStyle}
        {...props}
        hasChildren={!!children}
        loading={loading}
        disabled={loading || disabled}
      >
        {loading && <LoadingSpinner position="absolute" />}
        {icon}
        {children}
      </ButtonCore>
    );
  },
);
```

### DIFF
--- a/components/button/Button.js
+++ b/components/button/Button.js
@@ -86,19 +86,41 @@ ButtonCore.defaultProps = {
 	variant: 'primary',
 };
 
-const Button = React.forwardRef(({ children, icon, disabled, loading, ...props }, ref) => (
-	<ButtonCore
-		ref={ref}
-		{...props}
-		hasChildren={!!children}
-		loading={loading}
-		disabled={loading || disabled}
-	>
-		{loading && <LoadingSpinner position="absolute" />}
-		{icon}
-		{children}
-	</ButtonCore>
-));
+const Button = React.forwardRef(
+	({ children, icon, disabled, loading, size, width, height, ...props }, ref) => {
+		// Work around size prop bug (#393)
+		let sizeVariant = ButtonCore.defaultProps.size;
+		let widthStyle = width;
+		let heightStyle = height;
+		if (Object.keys(buttonSizes).includes(size)) {
+			sizeVariant = size;
+		} else if (size !== undefined) {
+			if (width === undefined) {
+				widthStyle = size;
+			}
+			if (height === undefined) {
+				heightStyle = size;
+			}
+		}
+
+		return (
+			<ButtonCore
+				ref={ref}
+				size={sizeVariant}
+				width={widthStyle}
+				height={heightStyle}
+				{...props}
+				hasChildren={!!children}
+				loading={loading}
+				disabled={loading || disabled}
+			>
+				{loading && <LoadingSpinner position="absolute" />}
+				{icon}
+				{children}
+			</ButtonCore>
+		);
+	},
+);
 
 const SegmentedButtonGroup = styled(Box).attrs(({ border }) => ({
 	border: border ?? 1,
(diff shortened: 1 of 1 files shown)


components/button/Button.js
```diff
@@ -86,19 +86,41 @@ ButtonCore.defaultProps = {
...
-		loading={loading}
-		disabled={loading || disabled}
-	>
-		{loading && <LoadingSpinner position="absolute" />}
-		{icon}
-		{children}
-	</ButtonCore>
-));
+const Button = React.forwardRef(
+	({ children, icon, disabled, loading, size, width, height, ...props }, ref) => {
```
korbinancell, line 90: Looks like this could be solved much more succinctly with the use of default parameters. Consider:
```javascript
const Button = React.forwardRef(
	(
		{
			children,
			icon,
			size = ButtonCore.defaultProps.size,
			variant = ButtonCore.defaultProps.variant,
			...props
		} = {},
		ref,
	) => (
		<ButtonCore ref={ref} size={size} variant={variant} {...props} hasChildren={!!children}>
			{icon}
			{children}
		</ButtonCore>
	),
);
```

korbinancell (approved):

korbinancell (changes requested):
Whoops I approved instead of asking for changes

components/button/Button.js
tywmick: Ah, so that does fix the "`size` should be defined before `width` and `height`" problem, but the other problem I was trying to mitigate with that extra logic was the fact that the value of the `size` prop is currently being sent to both the custom `sizeVariant` _and_ the `layout` utility (#393):

https://github.com/Faithlife/styled-ui/blob/eee81d8fdb743a47cfa32e727f41428cdcf264c4/components/button/Button.js#L11-L15
https://github.com/Faithlife/styled-ui/blob/eee81d8fdb743a47cfa32e727f41428cdcf264c4/components/button/Button.js#L70-L76

Styled System's [`layout` utility](https://styled-system.com/table/#layout) includes a prop called `size` that sets both `width` and `height`.

Say a user tries to use the `size` prop in the `layout` sense and writes `<Button size="5rem" />` to make a button with a width and height of `5rem`. Currently, it'll set the width and height correctly, but it'll also give `sizeVariant` a value it can't handle, which I think will prevent it from getting the [text, padding, and border styles](https://github.com/Faithlife/styled-ui/blob/eee81d8fdb743a47cfa32e727f41428cdcf264c4/theme/buttons.js#L3-L22) that the `sizeVariant` is supposed to give it. With these changes, when `Button` sees that `'5rem'` isn't a valid `buttonSizes` key, it'll pass `size="medium" width="5rem" height="5rem"` to `ButtonCore` instead.

Not a good long-term fix, for sure. Should we just trust users not to try the `layout` version of `size`, though, to make this component code less confusing? Or should we keep this and I add a few more explanatory comments?

Figuring users didn't want to use the `layout` version of `size` was my motivation for #394, taking out the `layout` `size` prop altogether so we wouldn't have to worry about the duplication.

korbinancell: I say lets not overload size and just let it refer to the size variant. We can just ignore the size layout prop.

korbinancell (approved):

PR #400, opened 2020-12-04
Create v6 SimpleToast with Styled System

tywmick:
Another #378— Here's a v6 `SimpleToast` that replaces the `styleOverrides` prop with Styled System utilities for color, layout, and position. I also replaced an old `theme` prop to use the new Styled System [theme](https://github.com/Faithlife/styled-ui/blob/c6a629f1453e323529e98c5fd911d93dc6ac6ea3/theme/core.js) instead.

The most notable changes are in [`styled.jsx`](https://github.com/Faithlife/styled-ui/pull/400/files#diff-61ce2e7e690369908fcb26c5ef810b37360495a42ee4edf06aaa30980f788a51). In it, since the default styles include a media query, I placed the Styled System primitives inside an `&&` block:

```
&& {
  ${color}
  ${layout}
  ${position}
}
```

This way, if a developer assigns a single value instead of an array to any of those Styled System props, meaning for that single value to apply across all screen width breakpoints, this `&&` specificity will ensure it wins any conflicts with defaults inside in the media query.


components/simple-toast/styled.jsx
```diff
@@ -1,20 +1,21 @@
...
 
-const slideIn = ({ styleOverrides }) => keyframes`
+const slideIn = ({ bottom }) => keyframes`
 	from {
 		bottom: 0;
 	}
 
 	to {
-		bottom: ${styleOverrides.bottomOffset || toastOffset.desktop};
+		bottom: ${(Array.isArray(bottom) ? bottom?.[1] || bottom?.[2] : bottom) || toastOffset.desktop};
```
korbinancell, line 18: This is kinda defeating the purpose of the responsive array props as is. Maybe use something to generate those breakpoints for you like so:

```javascript
const getResponsiveProp = propName => css`
	${({ theme, ...props }) => {
		const propValue = props[propName];
		if (Array.isArray(propValue)) {
			return css`
				${propName}: ${propValue[0]};

				@media screen and (min-width: ${theme.breakpoints.tablet}) {
					${propName}: ${propValue.length === 3 ? propValue[1] : propValue[0]}
				}

				@media screen and (min-width: ${theme.breakpoints.desktop}) {
					${propName}: ${propValue[2] ?? propValue[1]}
				}
			`;
		}

		return `${propName}: ${propValue};`;
	}}
`;
```

Then this line can become 
```javascript
bottom ? getResponsiveProp('bottom') : `bottom: ${toastOffset.desktop}`
```

components/simple-toast/styled.jsx
```diff
@@ -80,19 +82,19 @@ export const ToastContainer = styled.div`
 	}};
 
 	/** Desktop */
-	@media (min-width: ${mediaSizes.tablet}) {
+	@media (min-width: ${themeGet('breakpoints.tablet')}) {
 		padding: 16px;
-		${({ styleOverrides }) => (!styleOverrides.width ? `min-width: ${toastMinWidth}` : '')};
-		height: ${({ styleOverrides }) => styleOverrides.height || toastHeight.desktop};
+		${({ width }) => (!width || !width?.[1] ? `min-width: ${toastMinWidth};` : '')}
```
korbinancell, line 87: This will always apply the min-width for non-array widths. We can just assume if there is a width it will work as planned.

korbinancell (changes requested):

components/simple-toast/styled.jsx
tywmick: Oh, perfect! What had been throwing me off there was that since I hadn't used keyframes much before, I thought that since that `slideIn` function was getting called inside the tablet media query that the CSS generated inside `slideIn` would be placed inside that media query as well. Now I'm remembering that keyframes are defined separately from where they're called 🤦🏼‍♂️ which is the whole point of styled-components' handy `keyframes` function. Learning! ✊🏼

tywmick: Hmm, one question, though: looking at Styled System's [Responsive Styles](https://styled-system.com/responsive-styles) page, it looks like if you sending an array of length 2, say `bottom={['50px', '35px']}`, Styled System would treat that the same as `bottom={['50px', '25px', null]}`. Is that right? If so, sending `bottom={['50px', '25px']}` would result, on a tablet screen size, of the animation continuing all the way to `50px` but then skipping back down to rest at `25px`, right?

If that's the case, would the following be a good change for that section?

```js
if (Array.isArray(propValue)) {
  return css`
    ${propName}: ${propValue[0]};

    @media screen and (min-width: ${theme.breakpoints.tablet}) {
      ${propName}: ${propValue[1]};
    }

    ${propValue.length === 3
      ? css`
          @media screen and (min-width: ${theme.breakpoints.desktop}) {
            ${propName}: ${propValue[2]};
          }
        `
      : ''}
  `;
}
```

korbinancell: Ah, looks like you are correct. 👆 that looks good to me

PR #403, opened 2020-12-04
Fix loading prop to not cause react error message

EthanRutherford:
image
It would appear that there's an actual [loading prop](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-loading), so react in dev-mode was throwing an error because the property is not boolean.
That prop has no meaning outside of an iframe, so I see no need to change the outward api, but changed the internal name to avoid the warning.


michael-sterling:
Nice sleuthing!

RobertBolender (approved):

PR #404, opened 2020-12-04
Create v6 Slider with Styled System

tywmick:
Next stop on #378.

`Slider` didn't require too much updating: the only `styleOverrides` property being used was `backgroundColor`, to [fix the background of inactive track](https://faithlife.github.io/styled-ui/#/slider/variations?a=note-on-background-colors) when the slider is placed over a non-white backround. I just swapped that out with Styled System's [`backgroundColor`/`bg`](https://styled-system.com/table/#color).

Also, since the main `Box` component is already receiving all the excess props, I added `...Box.propTypes` to `Slider`'s `propTypes`.

Would it be helpful if I restructured the `TrackPart`s so that the blue background gradient stops at the thumb (correctly proportioning the gradient for the new stop point, maybe using [Chroma.js](https://gka.github.io/chroma.js/)) and we just give the inactive part of the track `bg="transparent"`? That way developers wouldn't have to worry about explicitly changing the background color prop when using `Slider` over a non-white background. Or is a white default background there still preferred (or the additional color scale computation too expensive)?


components/slider/component.jsx
```diff
@@ -308,7 +303,7 @@ export class Slider extends PureComponent {
 					<TrackPart
 						left={`${getPercentage(activeStopIndex, stopCount - 1)}%`}
 						right={0}
-						bg={(styleOverrides && styleOverrides.backgroundColor) || '#fff'}
+						bg={props?.backgroundColor || props?.bg || '#fff'}
```
korbinancell, line 306: nit prefer ?? over ||

components/slider/component.jsx
```diff
@@ -357,12 +351,11 @@ export class Slider extends PureComponent {
 const TrackPart = ({ children, left, right, ...props }) => (
 	<Box
 		position="absolute"
-		left={0}
-		right={0}
+		left={left || 0}
```
korbinancell, line 354: nit: ?? vs ||

components/slider/component.jsx
tywmick: Ah, that is much better! Thanks!

korbinancell (approved):

PR #406, opened 2020-12-09
Migrate tooltip to use popover v2

EthanRutherford:
Migrates tooltips to use the v2 version of poppers.

### DIFF
--- a/components/popover/tooltip.jsx
+++ b/components/popover/tooltip.jsx
@@ -1,12 +1,12 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, useRef } from 'react';
 import PropTypes from 'prop-types';
 import { useDebouncedCallback } from '../shared-hooks'; // Copied from react-ui. Import from there when made available.
-import { Popover } from './component';
-import { PopoverManager } from './popover-manager';
-import { PopoverReference } from './popper-helpers';
+import { Popover } from '../popover-v6';
+import { Box } from '../Box';
 
-/** Simple tooltip that uses popovers internally. Does not support custom positioning. */
+/** Simple tooltip that uses popovers internally. */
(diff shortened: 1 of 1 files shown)


components/popover/tooltip.jsx
```diff
@@ -27,29 +27,34 @@ export function Tooltip(props) {
 
 	useEffect(() => {
 		setIsOnMobile(window.matchMedia('(hover: none)').matches);
-		return () => {
-			handleMouseLeave.cancel();
-		};
+		return () => handleMouseLeave.cancel();
```
RobertBolender, line 30: Non-blocking nit: [Avoid Returning Meaningless Results from Arrow Functions](https://github.com/Faithlife/CodingGuidelines/blob/master/sections/javascript/javascript.md#avoid-returning-meaningless-results-from-arrow-functions) is in our coding guidelines.

RobertBolender (approved):

components/popover/tooltip.jsx
```diff
@@ -27,29 +27,34 @@ export function Tooltip(props) {
...
 		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, []);
 
+	const referenceProps = {
+		onMouseEnter: handleMouseEnter,
+		onMouseLeave: handleMouseLeave,
+		onClick: () => (isOnMobile ? toggle() : setToolTipIsOpen(false)),
+		ref,
```
RobertBolender, line 38: I know that refs aren't exactly the same as other props, and that they don't function with function components, so will this cause any warnings or errors when passed to the children function on line 46?

EthanRutherford: No, because we're calling the children function as a function, and not rendering it as if it were a function component.
`children(...)` vs `<children {...} />`

PR #410, opened 2020-12-13
Create v6 ParameterSentence components with Styled System

tywmick:
Rolling along on #378 🚗

Just a draft at the moment. `ParameterSentence` and `ParameterInputBox` are ready, but working on `ParameterSelect` raised some questions about `Listbox` that need to be resolved first. I'm going to bring that up in a new issue (#411).

In the meantime, though, one change in here that may seem unrelated at first glance is in [`components/utils/filter-props.js`](https://github.com/Faithlife/styled-ui/pull/410/files#diff-86353b9f59e64bab90de34406d22e5bdfec8a4c23374a60ec6174d73fc80194f). I was about to build something similar while adapting `ParameterInputBox` when I noticed this file and that its `filterChildProps` function was very close to what I needed. I saw it wasn't being used anywhere in the project, so I took the liberty of adapting it to a slightly more general purpose, renaming it to just `filterProps` (though it can still be used for filtering child props specifically). Hope that's cool. 👍🏼


components/parameter-sentence/parameter-input.jsx
```diff
@@ -8,14 +14,17 @@ export const ParameterInputBox = React.forwardRef((props, ref) => {
...
 		onFocus,
 		onBlur,
-		...inputProps
+		...otherProps
 	} = props;
+	const [styleProps, inputProps] = filterProps(otherProps, {
+		...layoutPropTypes,
+		...typographyPropTypes,
+	});
+	const { width, fontSize, theme } = styleProps;
```
RobertBolender, line 26: `theme` is not going to be matched by `layoutPropTypes` or `typographyPropTypes`, will it?

components/parameter-sentence/styled.jsx
```diff
@@ -1,67 +1,37 @@
...
 import { resetStyles } from '../utils';
 
 const selectStyling = css`
 	white-space: nowrap;
 	min-height: fit-content;
-	font-size: ${({ styleOverrides }) => styleOverrides.fontSize || '16px'};
-	width: ${({ styleOverrides }) => styleOverrides.width};
-	border-bottom: dashed ${thickness.two} ${colors.blueBase};
+	/* font-size: ${({ styleOverrides }) => styleOverrides.fontSize || '16px'}; */
+	/* width: ${({ styleOverrides }) => styleOverrides.width}; */
```
RobertBolender, line 10: I know this is still a work-in-progress draft, but don't forget to remove these comments :)

components/parameter-sentence/styled.jsx
```diff
@@ -1,67 +1,37 @@
...
+		isOpen ? themeGet('colors.parameterSentence.active') : themeGet('colors.gray66')};
 	font-family: inherit;
 	border-radius: 0;
 	line-height: 1;
 	padding-bottom: 2px;
 
 	&:hover {
 		&:not(:focus) {
-			color: ${colors.blueBase};
+			color: ${themeGet('colors.blue4')};
```
RobertBolender, line 22: Friendly reminder to use semantic theme colors rather than specific color shades, to allow for easier theming.

components/utils/filter-props.js
```diff
@@ -1,15 +1,27 @@
...
+		if (filterList.includes(propName)) {
+			filteredProps[propName] = propValue;
 		} else {
-			parentProps[propName] = prop;
+			remainingProps[propName] = propValue;
 		}
 	}
 
-	return [parentProps, childProps];
+	return [filteredProps, remainingProps];
```
RobertBolender, line 26: Tiniest of nit-picks: the word "filtered" to me can have two meanings: either ones that *match* a filter or ones that have been *removed* by a filter. Now, the presence of the word "remaining" does a tremendous help in interpreting these variable names, but could we make it just a step easier and call these "matchingProps" and "remainingProps"?

A similar nit-pick would be for the call-site of this function above, where it's not *immediately* apparent at the callsite whether `styleProps` or `inputProps` are the ones that *match* the provided object.

RobertBolender: Also: hearty approval for refactoring/repurposing this unused utility method. It has been sitting unused since July 2019 😄

components/parameter-sentence/parameter-input.jsx
tywmick: It [won't](https://styled-system.com/table/#typography), but I have realized that I probably can't use object destructuring in imports like I tried to do at the top of this file, and @styled-system/prop-types doesn't have exports named `layout` or `typography`, so I'll need to fix that as well...

```diff
-import {
-  layout as layoutPropTypes,
-  typography as typographyPropTypes,
-} from '@styled-system/prop-types';
+import styledSystemPropTypes from '@styled-system/prop-types';

...

   const [styleProps, inputProps] = filterProps(otherProps, {
-    ...layoutPropTypes,
+    ...styledSystemPropTypes.layout,
-    ...typographyPropTypes,
+    ...styledSystemPropTypes.typography,
   });
```

tywmick: Oh! I missed the entire point of your question: that's not the right place to get `theme`. ☝🏼 I should just use `DefaultThemeProvider` instead of the `theme` prop so I don't have to pass it down the tree manually.

components/utils/filter-props.js
tywmick: Ooh good point on the variable names! 👌🏼 Thanks!

<code>style<b>dSystem</b>Props</code>/`inputProps` at the call-site, perhaps?

components/parameter-sentence/styled.jsx
tywmick: Ope good call! 😂

components/utils/filter-props.js
RobertBolender: This is a step forward, I'm happy with it, and would be happy to merge it. 👍 

I can now clearly tell at a glance at the call-site that I intended for the matching props to be put into the first variable rather than the second. But there is still the possibility that I might be unfamiliar with `filterProps` and not know which variable is returned first..

What would you think of returning an *object* with two properties: `matchingProps` and `unmatchingProps` which could be renamed at the call-site such as:

```
const { matchingProps: styledSystemProps, unmatchingProps: inputProps } = filterProps(otherProps, {
  ...styledSystemPropTypes.layout,
  ...styledSystemPropTypes.typography,
});
```

Again, I'm happy with what we already have, just throwing this out there as a non-blocking suggestion.

tywmick: Ooh I like it! Returning an object will also give the added benefit of being able to just capture one property or the other if you don't need both 👌🏼

Maybe name that second property `remainingProps` instead of `unmatchingProps`? Or `nonMatchingProps`.

RobertBolender: That works for me 👍 
Quick nit: the JSDoc comment still says that `filterProps` returns an array

tywmick: Oop good catch; thanks!

tywmick:
I've updated `ParameterSelect` now with the `Listbox` changes I'm expecting, just passing Styled System props instead of the old `styledOverrides` prop. Once @korbinancell's v6 `Listbox` is ready (see #411), I'll make one last sweep here to make sure `ParameterSelect` is compatible with the new API, then I'll update the Catalog documentation, and then this PR should be ready to go. 👍🏼

RobertBolender:
> I've updated `ParameterSelect` now with the `Listbox` changes I'm expecting, just passing Styled System props instead of the old `styledOverrides` prop. Once @korbinancell's v6 `Listbox` is ready (see #411), I'll make one last sweep here to make sure `ParameterSelect` is compatible with the new API, then I'll update the Catalog documentation, and then this PR should be ready to go. 👍🏼

I think that we can go ahead and get this PR merged first, then follow-up with any necessary changes to the v6 ParameterSelect once the Listbox changes are in.
Without objection, I'll merge this tomorrow.

tywmick:
Sounds good!

PR #412, opened 2020-12-14
Create v6 Checkbox with global Styled System theme and child config components

tywmick:
Continuing #378.

v5's `Checkbox` had already enabled _optional_ use of the global Styled System theme, so this v6 version just removes the legacy `theme` prop and its related `themeOverrides` functionality.

While I was at it, I also removed the explicit `className` prop from `Checkbox`, since it seemed unnecessary in this version, and I created an `...otherProps` object to pass to `Styled.CheckboxContainer` as well. On that note, should I enable any Styled System categories for this component? None are enabled currently.


tywmick:
Just added some child config components, `Checkbox.Box` and `Checkbox.Label`! Those implementations look good? Should I add some additional Styled System categories to either?

Here's a [new section](https://deploy-preview-412--faithlife-styled-ui.netlify.app/#/checkbox/variations-v6?a=custom-boxlabel-props) on the v6 Checkbox Variations page that shows them in action.

tywmick:
Didn't add the new child config components to the upgrade guide since they're a non-breaking change, but let me know if we want non-breaking changes in this guide. 👍🏼

RobertBolender (approved):

PR #415, opened 2020-12-17
Create v6 LoadingSpinner with Styled System

tywmick:
A little more #378.

I removed the legacy `theme` prop that had been used in v5 and added Styled System categories [`layout`](https://styled-system.com/table#layout) and [`position`](https://styled-system.com/table#position). `layout` because v5 `LoadingSpinner` used a `height` prop and `position` because v6 `Button` is [trying to use](https://github.com/Faithlife/styled-ui/blob/f52061da4c35f4be53e2e524cf840b6f6a148b84/components/button/Button.js#L109) a `position` prop on `LoadingSpinner` even though it isn't enabled.

Should I add any other Styled System prop categories? While I'm asking, is there a standard set of Styled System prop categories that we want all components to have?


components/loading-spinner/styled.jsx
```diff
@@ -41,12 +42,17 @@ export const Spinner = styled(Box)`
...
 
-	${props =>
-		props.height &&
-		css`
-			width: ${props.height}px;
-			height: ${props.height}px;
-		`}
+	${layout}
+
+	/* Square up dimensions in case it's necessary */
```
RobertBolender, line 47: Clever, I like it 👍

PR #416, opened 2020-12-17
Create v6 Radio with global Styled System theme and child config components

tywmick:
Another stop on #378.

I replaced the legacy `theme` prop with our global Styled System theme. While I was at it, I also converted it from a class component to a function component, since this one was a pretty simple switch.

Any general Styled System categories I should add to this one?


tywmick:
Just added some child config components, `Radio.Icon` and `Radio.Label` (much like the ones I added to `Checkbox` in #412). Here's [how they look in action](https://deploy-preview-416--faithlife-styled-ui.netlify.app/#/radio/variations-v6?a=custom-iconlabel-props).

tywmick:
Didn't add the new child config components to the upgrade guide since they're a non-breaking change, but let me know if we want non-breaking changes in the upgrade guide.

RobertBolender (approved):

PR #418, opened 2020-12-18
Rename "text-input-v2" exports

tywmick:
For #378.

Renames `text-input` to `legacy-text-input` and `text-input-v2` to `text-input`. Everything's building and testing correctly, so I'm pretty sure I caught everything. 👍🏼 This a good solution? Are we wanting to just remove everything under the legacy `text-input` export, since everything in it is deprecated?


RobertBolender:
> Are we wanting to just remove everything under the legacy text-input export, since everything in it is deprecated?

Yep! I already had that removal in my WIP PR for v6, so I went ahead and cherry-picked the other commits from this PR onto that one, so the renaming change will be merged when we are ready to ship v6.

https://github.com/Faithlife/styled-ui/pull/413

PR #420, opened 2020-12-18
Upgrade to styled-components v5

tywmick:
Continuing #378.

Both the [v5 migration FAQ](https://styled-components.com/docs/faqs#what-do-i-need-to-do-to-migrate-to-v5) and the [v5 announcement post](https://medium.com/styled-components/announcing-styled-components-v5-beast-mode-389747abd987) say that the only breaking change is as follows:

> Note: The subfunction object-form `.attrs({ prop: props => {} })` syntax that was deprecated in v4 is removed in v5. Use function-form attrs instead `.attrs(props => ({}))` (you should have been seeing console warnings to make this update ahead of time.)

I've checked every occurrence of `.attrs` in Styled UI, and none of them use the deprecated form—they're all either props factory functions or simple props objects with no subfunction values, both of which are correct, though the [`attrs` API docs](https://styled-components.com/docs/api#attrs) are a bit ambiguous there (styled-components/styled-components-website#748). So I don't think anything else needs to be changed—everything's building and testing correctly on my end. 👍🏼

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,7 +119,7 @@
 		"rimraf": "^2.6.2",
 		"sass-loader": "^7.0.3",
 		"style-loader": "^0.21.0",
-		"styled-components": "4.3.1",
+		"styled-components": "^5.2.1",
 		"stylelint": "^10.1.0",
 		"stylelint-config-styled-components": "^0.1.1",
 		"stylelint-processor-styled-components": "^1.8.0",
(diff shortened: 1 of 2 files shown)


RobertBolender (approved):
Thanks for checking the `.attrs` usages within this repo.
I want to test upgrading the `styled-components` package within the Faithlife.com project first before merging this PR, that project is our largest and primary consumer and I want to be certain that `styled-components` v5 doesn't cause any issues with SSR or anything.

PR #422, opened 2020-12-22
Add initial snapshot tests based on Catalog "variants" pages

tywmick:
Seeing the options proposed in #378 for testing, I first looked at [Percy](https://percy.io/), but it required creating an account. I then tried out [jest-image-snapshot](https://github.com/americanexpress/jest-image-snapshot), but that proved difficult to implement, because it would require starting the Catalog dev server during test setup, and [Catalog](https://github.com/interactivethings/catalog/tree/master) doesn't include a JavaScript API for managing its dev serer. It wouldn't be impossible (I thought about calling `catalog-build` in the script and then using [serve-handler](https://github.com/vercel/serve-handler) to create the server from the static build)—just difficult.

What may be quicker to implement is normal tree-based [snapshot testing](https://jestjs.io/docs/en/snapshot-testing) with [Jest](https://jestjs.io/en/) and [Enzyme](https://enzymejs.github.io/enzyme/), just copying over the component trees from each component's "variants" page on the Catalog site for a first draft. To that end, how does this setup and these `Accordion` tests look for starters?

I placed a new `test` folder inside `components/accordion` to house test files and snapshots, but I can move component tests into the centralized `tests` directory if that would be preferred. I figured putting tests next to the source of the components themselves would remind folks to keep those tests up to date, but either way. 👍🏼

One advantage of tree-based snapshot tests over image snapshots, by the way, is that if you're analyzing the HTML tree, you'll see if any accessibility features are broken, something image snapshots wouldn't catch.


RobertBolender:
We're still having some internal discussions about how we want to unify our approach to testing, and we're going to postpone further investigation into snapshot testing until after our v6 launch.

I appreciate the effort you've put into this, but we're not going to accept any snapshot testing PRs at this time.

I have updated our Roadmap to indicate this, and replaced the short-term v6 goal with having basic Jest unit tests using [React-Testing-Library](https://github.com/testing-library/react-testing-library) rather than snapshot tests.

PR #424, opened 2020-12-28
Check date periods for duplicate display names in DatePeriodPicker

### LINKED ISSUE #423: DatePeriodPicker: Filter duplicates from date periods list
Following up on #389—just saving this here so I don't forget about it. More context in [this discussion](https://github.com/Faithlife/styled-ui/pull/389#discussion_r548043519) with @rmjohnson and @RobertBolender, but here's the key quote:

> How would you feel about a filtering the list for duplicates, keeping only the first for each display string, and logging a warning/error for any duplicates? (Not blocking for this PR, it can be a follow-up)

tywmick:
Resolves #423.

I added a new method called `getUniqueDatePeriods` that returns only the first date period of each display name, retaining the original array index in a new `index` property. It also sends a console warning in development environments when duplicates are passed. The `render` and `getDatePeriodIndex` methods now call this method instead of using `this.props.datePeriods`.

While I was at it, I also made `dateRange` and its `start` and `end` properties required in the prop type for `datePeriods`. I figured that was the intent already (that all passed date periods have date ranges, not just names), but if that's not the case I can change it back. 👍🏼

### DIFF
--- a/components/date-period-picker/component.jsx
+++ b/components/date-period-picker/component.jsx
@@ -120,7 +120,7 @@ export class DatePeriodPicker extends PureComponent {
 	 * Gets the index of a passed date period that happens to match a manually selected date range, if
 	 * one matches.
 	 * @param {{ start: Date, end: Date }} dateRange - The manually selected start and end dates.
-	 * @return {number | null} The index of a matching date period or `null` if no period matches.
+	 * @return {number | null} The original array index of a matching date period or `null` if no period matches.
 	 */
 	getDatePeriodIndex = ({ start, end }) => {
 		if (start === undefined || end === undefined || start === null || end === null) {
@@ -132,7 +132,7 @@ export class DatePeriodPicker extends PureComponent {
 				isSameDay(start, datePeriod.dateRange.start) &&
 				isSameDay(end, datePeriod.dateRange.end)
 			) {
-				return datePeriod.index;
+				return datePeriod.originalIndex;
 			}
 		}
 
@@ -150,9 +150,9 @@ export class DatePeriodPicker extends PureComponent {
 
 	/**
 	 * Returns only the first date period of each display name, retaining the original array index in
-	 * a new `index` property. Also sends a console warning in development environments when
+	 * a new `originalIndex` property. Also sends a console warning in development environments when
 	 * duplicates are passed.
-	 * @returns {Array<{ index: number; displayName: string; dateRange: { start: Date; end: Date } }}
+	 * @returns {Array<{ displayName: string; dateRange: { start: Date; end: Date }; originalIndex: number}}
 	 */
 	getUniqueDatePeriods = () => {
 		const uniqueDatePeriods = [];
@@ -172,7 +172,7 @@ export class DatePeriodPicker extends PureComponent {
 				}
 			} else {
 				// Display name is unique: add period to return array (and save original index)
-				uniqueDatePeriods.push({ ...currentPeriod, index: i });
+				uniqueDatePeriods.push({ ...currentPeriod, originalIndex: i });
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


RobertBolender:
@rmjohnson do you know of any uses of this component where the date range was intended to be open-ended?
Is it possible to select a date range that with a start date but no end, or end date with no start?

rmjohnson:
@RobertBolender, @tywmick, we do have an "All time" use case which has `null` for both start and end. I'm not sure if there is a use case where we provide just one though (as we usually will set end to today instead of leaving it `null`).

tywmick:
Hmm, looks like PropTypes [doesn't have a specific null type](https://stackoverflow.com/questions/40729634/how-to-specify-null-prop-type-in-reactjs), so the only way to allow a null value is by taking off `.isRequired`. Just taking `.isRequired` off `start` and `end` but leaving it on `dateRange` sound good, then?

Oh, should I add `isRequired` to the `PropTypes.shape` so the `datePeriods` array can't contain any `null` or `undefined` elements?

```diff
datePeriods: PropTypes.arrayOf(
  PropTypes.shape({
    displayName: PropTypes.string.isRequired,
    dateRange: PropTypes.shape({
-     start: PropTypes.instanceOf(Date).isRequired,
+     start: PropTypes.instanceOf(Date),
-     end: PropTypes.instanceOf(Date),
+     end: PropTypes.instanceOf(Date),
    }).isRequired,
- }),
+ }).isRequired,
).isRequired,
```

components/date-period-picker/component.jsx
```diff
@@ -113,43 +116,84 @@ export class DatePeriodPicker extends PureComponent {
...
+			if (uniqueDisplayNames.includes(currentPeriod.displayName)) {
+				if (process.env.NODE_ENV !== 'production') {
+					console.warn(
+						`A \`DatePeriodPicker\` has been passed multiple date periods named "${
+							currentPeriod.displayName
+						}". Only the first "${currentPeriod.displayName}" period has been passed.`,
+					);
+				}
+			} else {
+				uniqueDatePeriods.push(Object.assign({ index: i }, currentPeriod));
```
RobertBolender, line 173: Does passing `{ index: i }` here mean that in the case of a duplicate, I'll end up with an array with `undefined` gaps?
Will we need to filter for undefined/null values later when we render the `Styled.DatePeriod`s?

tywmick: I was meaning to add a new property `index` to each unique element of the `datePeriods` array, so for instance, if `this.props.datePeriods` is

```js
[
  {
    displayName: 'Last 7 Days',
    dateRange: { start: dateFunctions.addDays(new Date(), -7), end: new Date() },
  },
  {
    displayName: 'Last 7 Days',
    dateRange: { start: dateFunctions.addDays(new Date(), -14), end: new Date() },
  },
  {
    displayName: 'Last 30 Days',
    dateRange: { start: dateFunctions.addDays(new Date(), -30), end: new Date() },
  },
]
```

...then `this.getUniqueDatePeriods()` should return...

```js
[
  {
    index: 0,
    displayName: 'Last 7 Days',
    dateRange: { start: dateFunctions.addDays(new Date(), -7), end: new Date() },
  },
  {
    index: 2,
    displayName: 'Last 30 Days',
    dateRange: { start: dateFunctions.addDays(new Date(), -30), end: new Date() },
  },
]
```

Then the component sends the value of that `index` property to `setSelectedDate` instead of the array index within `this.getUniqueDatePeriods()` so that the dev still gets the index they expect from the original `datePeriods` array they passed.

RobertBolender: Oh! Got it, got it. I had mentally conflated the `index` property here with an array index, and missed that the `.map` function below now destructures `index` from the first object parameter instead of using the second "index" parameter.

I liked the earlier version with `.includes` better than the newer `.find(...) === undefined`.

I won't pick this nit too strongly, but perhaps a name other than `index` would be helpful here? `datePeriodIndex` or something else?

tywmick: Mm, that is a good idea—`originalIndex` perhaps? Since the consumer will never have to see it.

RobertBolender: That sounds good to me 👍

RobertBolender (approved):

PR #426, opened 2021-01-04
Chms 1328 Add select checkbox option

randallknutson:
In CHMS we have the ability to do a multi-select from the dropdown. This means each option has a checkbox and multiple can be selected at the same time. This PR adds the functionality of having a configuration option that allows having that checkbox.


components/text-input-v2/styled.jsx
```diff
@@ -14,8 +14,12 @@ export const AvatarOptionImage = styled.img`
 	width: 40px;
 `;
 
+export const AvatarOptionCheckbox = styled.div`
+	padding-right: ${props => props.theme.space[3]};
+`;
+
 export const AvatarOptionText = styled.div`
-	padding: ${props => props.theme.space[2]};
+	padding: ${props => props.theme.space[3]};
```
RobertBolender, line 22: @randallknutson, would you have a link to a spec for the spacing on this component?

randallknutson: I'm not sure where the original spec is but I made this change to match the existing spacing (8px) in chms for the person record selector that we are replacing. At a 2 it is half the width of the existing component.

components/text-input-v2/select.jsx
```diff
@@ -156,13 +157,19 @@ const defaultComponents = {
...
 	const theme = useTheme();
+	const { avatar, label, secondaryLabel } = data;
 
 	return (
 		<Styled.AvatarOption {...props}>
 			{avatar ? (
 				<>
+					{!allowSelect ? null : (
+						<Styled.AvatarOptionCheckbox theme={theme}>
+							<Checkbox onClick={() => {}} isChecked={isOptionChecked(data)} disableAutoBlur />
```
RobertBolender, line 170: @randallknutson these new props should be documented, at minimum with propTypes.
`isOptionChecked` sounds like a boolean variable name, not a function name.
This `onClick` handler that does nothing also looks suspicious.

randallknutson: @RobertBolender would ```shouldOptionBeChecked``` be a better name? It should return a boolean, obviously, so it should have some type of boolean sounding name.

The onClick does seem a little strange at first but this is actually correct and the way the current ChMS component does this. Because it is inside of the Select component it triggers a secondary event of "select-option" which we handle elsewhere. This is the event we actually care about, not the click event. This is because clicking on the label or anywhere on the option should trigger the event to select/deselect it which is what the "select-option" event does.

I'll add some documentation for the new props.

RobertBolender: `shouldOptionBeChecked` would still be a boolean-sounding name.
For a function that returns a boolean, I would expect something like `getShouldOptionBeChecked` or `getIsOptionChecked`.

But.. since the consumer already controls both the `data` and the function that takes `data` and returns a boolean, could we instead have the consumer just pass in an `isChecked` boolean, rather than a function?

randallknutson: The prop exists on the `Select` component, not the `Option`. Therefore we can't just pass in a boolean. We have to pass in the function which can compare the current `Option` data to the list of selected options. You'll notice the `isOptionChecked` is from `selectProps` which is a top level prop.

RobertBolender: Thanks for helping me understand.
Does the `isSelected` prop not do what we need here?
https://github.com/JedWatson/react-select/blob/5bee3f40e74c165117a6765588ae906f07ad76f9/packages/react-select/src/components/Option.js#L81

If not, I would still be in favor of renaming the function to have a more function-like name.

randallknutson: The issue is that there is a "pending selected" stage for the checkbox. That is, the options have been selected "to be selected" but are not actually selected until the done button is clicked. For example, you check five items, their checkboxes are checked and then you click done. `isSelected` doesn't apply until the done button is pressed. We have to maintain a separate pendingSelected state.

randallknutson: Has been renamed and documentation added.

RobertBolender:
@randallknutson, since this `allowSelect` prop is only used by the `AvatarOption` component, what would you think about adding it to the option `data` object itself, rather than a top-level prop?

image


I would also suggest a name such as `showCheckbox`, rather than `allowSelect`, as the current name doesn't communicate to me how it's modifying the behavior of a "Select" component.

catalog/input/select.md
```diff
@@ -139,6 +139,34 @@ state: { tags: [] }
...
+		options={[
+			{ value: "washington", label: "Washington" },
+			{ value: "california", label: "California" },
+			{ value: "Texas", label: "Texas" }
+		]}
+		allowSelect={true}
+		getIsOptionChecked={(data) => {
+			return data.value && !!state.pendingSelectedValues.find((x) => x.value === data.value);
+		}}
+		onKeyDown={handleKeyDown}
```
RobertBolender, line 164: FYI: `handleKeyDown` is not defined and this crashes the example.

randallknutson:
> @randallknutson, since this `allowSelect` prop is only used by the `AvatarOption` component, what would you think about adding it to the option `data` object itself, rather than a top-level prop?
> 
> I would also suggest a name such as `showCheckbox`, rather than `allowSelect`, as the current name doesn't communicate to me how it's modifying the behavior of a "Select" component.

I agree that `showCheckboxes` makes more sense but adding a prop like variable to data doesn't make sense. Data is supposed to be the option data. While it would make sense to do something like that if you wanted to be able to turn checkboxes on and off per option, the configuration is designed to be more of the whole select. For example, you would set `showCheckboxes={true}` on the select and that would cause all the checkboxes to show. It would be very awkward to add them to each item in the data array that turns into the `Options`.

RobertBolender:
> For example, you would set showCheckboxes={true} on the select and that would cause all the checkboxes to show.

My only remaining (minor) concern is that this is dependent on the `<Option>` component to implement, and that we only show the checkbox in the `AvatarOption` component if an `avatar` is provided.

PR #429, opened 2021-01-06
Add common/typography Styled System props to v6 components

tywmick:
I'm running through and adding [`common`](https://github.com/Faithlife/styled-ui/blob/1e0c8465be657e65e3541fb895507c4b8ea85b5e/theme/system.js#L16-L34) and (when appropriate) [`typography`](https://github.com/Faithlife/styled-ui/blob/1e0c8465be657e65e3541fb895507c4b8ea85b5e/theme/system.js#L36-L50) Styled System props to components as per [this comment](https://github.com/Faithlife/styled-ui/issues/378#issuecomment-749862700) on #378. Started with Accordion and I'm going to continue down alphabetically. 👍🏼

### DIFF
--- a/components/accordion/accordion-header.jsx
+++ b/components/accordion/accordion-header.jsx
@@ -5,6 +5,7 @@ import { Text } from '../Text';
 import { Box } from '../Box';
 import { UtilityButton } from '../button';
 import { ChevronRight, ChevronDown } from '../icons/12px';
+import { typography } from '../../theme/system';
 import { useAccordionContext, useAccordionItemContext } from './accordion-util';
 
 export function AccordionHeader({
@@ -56,7 +57,7 @@ export function AccordionHeader({
 	}, [headerId, focusableChildList]);
 
 	return (
-		<Box
+		<HeaderBox
 			display="flex"
 			gridArea="header"
 			borderTop={1}
@@ -117,7 +118,7 @@ export function AccordionHeader({
 					{renderCustomIndicator({ isExpanded, onExpansion: handleExpansion })}
 				</Box>
 			) : null}
-		</Box>
+		</HeaderBox>
 	);
 }
 
@@ -131,7 +132,10 @@ AccordionHeader.propTypes = {
 	/** Receives an isExpanded boolean value. */
 	renderCustomIndicator: PropTypes.func,
 	...Box.propTypes,
+	...typography.propTypes,
 };
+
+const HeaderBox = styled(Box)(typography);
 
 const Heading = styled.header.attrs(({ ariaLevel }) => ({
 	role: 'heading',
--- a/components/accordion/accordion-item.jsx
+++ b/components/accordion/accordion-item.jsx
@@ -1,10 +1,12 @@
 import React, { useCallback, useMemo } from 'react';
+import styled from 'styled-components';
 import PropTypes from 'prop-types';
 import { useId } from '../shared-hooks';
 import { Box } from '../Box';
+import { typography } from '../../theme/system';
 import { useAccordionContext, AccordionItemContextProvider } from './accordion-util';
 
-export function AccordionItem({ children, index, pinned: isPinned }) {
+export function AccordionItem({ children, index, pinned: isPinned, ...otherProps }) {
 	const { expandedSections, onExpansion } = useAccordionContext();
 
 	const isExpanded = expandedSections.includes(index);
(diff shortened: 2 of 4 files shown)


tywmick:
Gave `DatePicker` a `div` wrapper in the process, to allow style props to apply to the whole thing. It was only "wrapped" in a `React.Fragment` before.

tywmick:
`DatePicker` was also previously using colors defined in [`shared-styles`](https://github.com/Faithlife/styled-ui/blob/1e0c8465be657e65e3541fb895507c4b8ea85b5e/components/shared-styles/index.js) instead of using the global Styled System theme, so I changed that while I was at it.

RobertBolender:
> Gave `DatePicker` a `div` wrapper in the process, to allow style props to apply to the whole thing. It was only "wrapped" in a `React.Fragment` before.

This is a kind of breaking change, please move this to a v6 version of the component.

components/accordion/accordion-header.jsx
```diff
@@ -131,8 +132,11 @@ AccordionHeader.propTypes = {
 	/** Receives an isExpanded boolean value. */
 	renderCustomIndicator: PropTypes.func,
 	...Box.propTypes,
+	...typography.propTypes,
 };
 
+const HeaderBox = styled(Box)(typography);
```
RobertBolender, line 138: The `HeaderBox` is unlikely to benefit from the typography props, because the actual text content is rendered inside a descendent component `<HoverableText>` with its own style props.

Upon further reflection, let's not move forward with adding `typography` to any new components that don't already have it, we can add them later on an as-needed basis.

tywmick: Sounds good. 👍🏼 I thought it _might_ be handy sometimes to be able to change a font family or text overflow or something for all children of a component instead of having to duplicate it on each child, but yeah, that scenario's probably not nearly common enough to be worth adding `typography` everywhere.

tywmick:
One change in c76ca22c93693a8328c3752c7f7728a94c20b4c0 (in [`date-period-picker/styled.jsx`](https://github.com/Faithlife/styled-ui/commit/c76ca22c93693a8328c3752c7f7728a94c20b4c0#diff-1a53cf3f003c44ba6f7935c85984c2ada3662479622fa570a53cea23d2e820bd)) caused a visual change in the component that it seems like was the original intent, just coded incorrectly before. Here's the change:

```diff
-import { colors, thickness, fonts } from '../shared-styles';
+import { themeGet } from '@styled-system/theme-get';
 
 ...
 
 export const DatePeriod = styled.div`
-	font: ${fonts.ui14};
+	${themeGet('textStyles.ui.14')}
 	...
 `;
 
 ...
 
 export const Label = styled.label`
 	...
-	font: ${fonts.ui14};
+	${themeGet('textStyles.ui.14')}
 	...
 `;
```

For context, here's `fonts.ui14`:

https://github.com/Faithlife/styled-ui/blob/1e0c8465be657e65e3541fb895507c4b8ea85b5e/components/shared-styles/index.js#L18-L22

And here's `textStyles.ui.14`:

https://github.com/Faithlife/styled-ui/blob/1e0c8465be657e65e3541fb895507c4b8ea85b5e/theme/textStyles.js#L89-L93

It's the same set of styles, but since the first version was written `font: ${fonts.ui14};` instead of just `${fonts.ui14}`, `font-weight` and `line-height` were set correctly, but not `font-size`, since that line ended up in the compiled CSS as `font: font-size: 14px;`.

Here's a before/after screenshot:

Screen Shot 2021-01-08 at 5 44 54 PM

All this to say, should I leave this at the smaller `ui.14` style which seems to have been the original intent, or should I increase it to `ui.16` to match the previous _actual_ font size (but also increase the line height by 2 pixels)?

RobertBolender:
> One change in c76ca22 (in date-period-picker/styled.jsx) caused a visual change in the component that it seems like was the original intent, just coded incorrectly before. Here's the change:

Solid catch.

> All this to say, should I leave this at the smaller ui.14 style which seems to have been the original intent, or should I increase it to ui.16 to match the previous actual font size (but also increase the line height by 2 pixels)?

I'll check with our design team, but it can stay at `ui.14` for now.

tywmick:
I added `propTypes` to all the `Dropdown` components in my intent to add `common` style props to all of them, but once I did so, I found that they already each accepted `common` props by passing them to `Box` or `Text` or something similar.

One probably unexpected addition in baa4444, though, is a new utility function:

https://github.com/Faithlife/styled-ui/blob/5e50f0d6c3f6b6b24d2998597ecf91efd2d3e2fa/components/utils/prop-types.js#L1-L15

I made it thinking it'd be nice to check component types of children on `SegmentedButtonGroup` and `DropdownMenu`. Here's how I'm using it:

https://github.com/Faithlife/styled-ui/blob/5e50f0d6c3f6b6b24d2998597ecf91efd2d3e2fa/components/button/Button.js#L176-L177

https://github.com/Faithlife/styled-ui/blob/5e50f0d6c3f6b6b24d2998597ecf91efd2d3e2fa/components/dropdown/dropdown-menu.jsx#L69-L81

Is that overkill and I should just make those `PropTypes.node` and call it good?

RobertBolender:
> Is that overkill and I should just make those `PropTypes.node` and call it good?

No objection to this.

PR #432, opened 2021-01-11
Menu

korbinancell:


components/listbox/listbox-toggle.jsx
```diff
@@ -1,53 +1,24 @@
-import React, { useCallback } from 'react';
-import PropTypes from 'prop-types';
+import React from 'react';
+import { Menu } from '../../index-v6';
```
RobertBolender, line 2: Nit: I would suggest importing from `../menu/` instead (here and elsewhere in this PR). In general I prefer to avoid reaching up from a file to an index that *also* imports that same file. Webpack is *probably* smart enough to avoid recursive/infinite strangeness here, but we can be ever-so-slightly safer while also making the code ever-so-slightly more clear and precise.

korbinancell: Makes sense to me. I was in the habit of doing it that way in the docs I guess.

Issue #431, opened 2021-01-08
[RFD] ImageWell


zachsnoek:
# ImageWell

* Specifically describe the problem you'd like the new component to solve:

There are a couple of instances of an "image well" (or "image selector") component. When clicked, the component will launch an AssetPicker and display a preview of the selected image. The component is not being exported from a shared library, so the styles are being copied across different projects. We would like to add this to `tv-management` but don't want to duplicate the component again.

* Describe the proposed API:

```js
ImageWell.propTypes = {
  size: PropTypes.string,
  onClick: PropTypes.function,
  disabled: PropTypes.boolean,
  previewUrl: PropTypes.string,
  fitOverlayToSquare: PropTypes.boolean,
}
```

Example usage:

This would show the default "+ Add Image" text and camera icon that are shown if there is no `previewUrl`.

```jsx
const handleClick = async () => {
  const image = await pickImage();
  setSelectedImage(image);
}

<ImageWell
  onClick={handleClick}
  previewUrl={selectedImage?.url}
/>
```

The default content shown when `previewUrl` is not specified can be changed by providing children.

```jsx
const handleClick = async () => {
  const roomLayout = await pickRoomLayout();
  setSelectedLayout(roomLayout);
}

<ImageWell
  size="small"
  onClick={handleClick}
  previewUrl={selectedLayout?.url}
>
  <RoomIcon />
  {"+ Add room layout"}
</ImageWell>
```

* Are there existing specs or designs?

    * [ChurchManagement](https://git.faithlife.dev/Logos/ChurchManagement/blob/master/chms-tool/src/components/EditRecordWizard/FormComponents/Image/ImageSelector.tsx)
    * [ProclaimDigitalSignage](https://git.faithlife.dev/Logos/ProclaimDigitalSignage/blob/master/src/components/signage-faithlife-tab/image-picker/index.jsx)

No preview image specified:

![Screen Shot 2021-01-08 at 9 47 55 AM](https://user-images.githubusercontent.com/26049962/104047949-58ffc200-5197-11eb-9fc8-44927f1bfa97.png)

Preview image specified:

![Screen Shot 2021-01-08 at 9 48 55 AM](https://user-images.githubusercontent.com/26049962/104047964-5e5d0c80-5197-11eb-91f1-b4f3d57ce591.png)

RobertBolender:
Excellent write-up 👍 
This looks like a great candidate for Styled-UI.
A couple of notes:
[...]

PR #433, opened 2021-01-11
Add ImageWell component

zachsnoek:
This PR adds an ImageWell component. #431


zachsnoek:
The current upper-right corner icon is based off of the icon in the Facilities/Resources tab:

Screen Shot 2021-01-12 at 11 40 30 AM

ChMS also uses an image well with a camera icon instead:

Screen Shot 2021-01-12 at 11 40 41 AM

Should we default to the X icon but allow that icon to be changed with an `icon` prop?

components/image-well/component.jsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { X } from '../icons/12px';
+import { Camera } from '../icons/18px';
+import * as Styled from './styled';
```
bryanrsmith, line 5: Please use styled system components and avoid the pattern of grouping styled components into a single file.

components/image-well/component.jsx
```diff
@@ -0,0 +1,70 @@
...
+	children,
+	previewUrl,
+	fitPreviewToSquare,
+	onSelectImage,
+	onRemoveImage,
+	localizedResources,
+}) => {
+	if (previewUrl) {
+		return (
+			<ImageWellBase previewUrl={previewUrl}>
```
bryanrsmith, line 17: Seems like this will render a button with no click handler, which is undesirable.

components/image-well/component.jsx
```diff
@@ -0,0 +1,70 @@
...
+				<Styled.WellPreview
+					style={{
+						backgroundImage: `url(${previewUrl})`,
+						backgroundSize: fitPreviewToSquare ? 'cover' : 'contain',
+						backgroundPosition: 'center',
+						backgroundRepeat: 'no-repeat',
+					}}
+				>
+					<Styled.RemoveIconContainer onClick={() => onRemoveImage?.()}>
```
bryanrsmith, line 26: Click handlers need to go on buttons, and we should not render a button that does nothing (so, either conditionally render the button or make the handler required)

components/image-well/component.jsx
bryanrsmith: Also, shouldn't you be able to click the image again to choose a different one?

components/image-well/component.jsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { X } from '../icons/12px';
+import { Camera } from '../icons/18px';
+import * as Styled from './styled';
+
+const ImageWell = ({
+	children,
+	previewUrl,
+	fitPreviewToSquare,
```
bryanrsmith, line 10: I'm a little surprised that we want this to be configurable. Has Design confirmed that the inconsistency is desirable?

components/image-well/component.jsx
```diff
@@ -0,0 +1,70 @@
...
+				<Styled.WellPreview
+					style={{
+						backgroundImage: `url(${previewUrl})`,
+						backgroundSize: fitPreviewToSquare ? 'cover' : 'contain',
+						backgroundPosition: 'center',
+						backgroundRepeat: 'no-repeat',
+					}}
+				>
+					<Styled.RemoveIconContainer onClick={() => onRemoveImage?.()}>
+						<X className="image-well_remove-icon" />
```
bryanrsmith, line 27: Prefer styled components APIs or use a selector

components/image-well/styled.jsx
```diff
@@ -0,0 +1,71 @@
+import styled from 'styled-components';
+import { colors, thickness, fonts } from '../shared-styles';
```
bryanrsmith, line 2: use theme. shared-styles is pending removal.

components/image-well/styled.jsx
```diff
@@ -0,0 +1,71 @@
...
+
+const fixedContainerWidth = '116px';
+const fixedContainerHeight = '116px';
+
+export const WellContainer = styled.div`
+	width: ${fixedContainerWidth};
+	height: ${fixedContainerHeight};
+	box-sizing: content-box;
```
bryanrsmith, line 10: Why?

components/image-well/styled.jsx
```diff
@@ -0,0 +1,71 @@
...
+`;
+
+export const WellButton = styled.button`
+	margin: 0;
+	padding: 0;
+	width: ${fixedContainerWidth};
+	height: ${fixedContainerHeight};
```
bryanrsmith, line 17: This seems fragile; would 100% or flex be more appropriate?

bryanrsmith (changes requested):
> Should we default to the X icon but allow that icon to be changed with an icon prop?

We should ask design which use cases this should accommodate and make sure the happy path is right.

components/image-well/component.jsx
zachsnoek: [Confirmed this with design](https://www.flowdock.com/app/logos/styled-ui/threads/BbgtEItR72govOHEZB5UYnVDFb9).

components/image-well/component.jsx
zachsnoek: One version of the ImageWell in ChMS shows the X icon but displays the image in a light box when the preview is clicked (and removes the image when the X is clicked). Another version shows the camera icon and prompts you to choose a new image regardless of where you clicked the image well.

I added a [commit](https://github.com/Faithlife/styled-ui/pull/433/commits/4793ae37eb0de087fbe1d0005c9f6215c5ee4b97) to add this behavior in and documented it; let me know if this behavior isn't what we want or if it should be implemented some other way.

zachsnoek:
I've added `ImageWell.PreviewContent` so that we can support custom controls over the preview image as shown in [this spec](https://app.zeplin.io/project/5b10273d413156335e1b6743/screen/5cc8b19387477603a84fa4c0). Custom content shown when there is no `previewUrl` is now wrapped in `ImageWell.SelectContent`.

components/image-well/image-well-icons.jsx
```diff
@@ -0,0 +1,68 @@
...
+	return (
+		<RemoveIconContainer onClick={handleClick}>
+			<X />
+		</RemoveIconContainer>
+	);
+};
+
+export const CameraIcon = ({ onClick = () => {} }) => {
+	const handleClick = e => {
+		e.stopPropagation();
```
RobertBolender, line 59: For both of these icons, is it ever appropriate for the consumer to *not* supply an onClick handler? Won't that result in the click being swallowed by this stopPropagation and not bubble up to the ImageWell click handler?

If the function is required, it should be marked as a required propType, and we shouldn't have the default `() => {}`

zachsnoek: Not supplying an `onClick` handler could be useful if you want to show the camera icon and have that call `onSelectImage` when clicked, but also provide some other icon with a custom `onClick`. Not providing the camera's `onClick` handler should just cause the event to bubble up:

```jsx
<ImageWell
    onSelectImage={() => alert("image well clicked")}
>
    <ImageWell.PreviewContent>
        <ImageWell.CameraIcon /> {/* clicking this calls onSelectImage */}
        <StarIcon onClick={(e) => {
            e.stopPropagation();
            alert("star clicked");
        }}/>
    </ImageWell.PreviewContent>
</ImageWell>
```

However, we do want to prevent bubbling on the default icons.

components/image-well/image-well-icons.jsx
```diff
@@ -0,0 +1,54 @@
...
+	}
+`;
+
+export const RemoveIcon = ({ onClick = () => {} }) => (
+	<RemoveIconContainer onClick={onClick}>
+		<X />
+	</RemoveIconContainer>
+);
+
+export const CameraIcon = ({ onClick = () => {} }) => (
```
RobertBolender, line 50: My previous comment still stands. These two icon components are exported, and they *must* always be passed an onClick handler, so they should have it defined as required on their propTypes and not have a default `() => {}`.

```js
CameraIcon.propTypes = {
  onClick: PropTypes.func.isRequired
}
```

components/image-well/component.jsx
```diff
@@ -0,0 +1,98 @@
...
+const ImageWell = ({
+	children,
+	previewUrl,
+	fitPreviewToSquare,
+	onSelectImage,
+	onRemoveImage,
+	localizedResources,
+	...props
+}) => {
+	const handleIconClick = (e, cb) => {
```
RobertBolender, line 18: Naming nit: Faithlife guidelines say to avoid abbreviations:
https://wiki.lrscorp.net/Avoid_Abbreviations

components/image-well/image-well-base.jsx
```diff
@@ -0,0 +1,43 @@
+import styled from 'styled-components';
+import { border, color, typography } from 'styled-system';
+
+export const fixedContainerWidth = '116px';
+export const fixedContainerHeight = '116px';
+
+export const WellContainer = styled.div`
+	width: ${fixedContainerWidth};
+	height: ${fixedContainerHeight};
```
RobertBolender, line 9: Two more questions:

1. Why 116px in particular?
2. Could this `div` instead be a `Box` so that it will accept styled-system layout props, allowing for overriding these values?

zachsnoek: 1. I took those styles from [ChurchManagement](https://git.faithlife.dev/Logos/ChurchManagement/blob/master/chms-tool/src/components/EditRecordWizard/FormComponents/Image/styles.less#L15-L16); otherwise there is no particular reason.
2. I'll make that change :+1:

RobertBolender: > otherwise there is no particular reason.

Also, did I miss somewhere in this PR where that exported const is used, or can it be removed?

zachsnoek: > Also, did I miss somewhere in this PR where that exported const is used, or can it be removed?

They're also being used in [`image-well-preview.jsx`](https://github.com/Faithlife/styled-ui/pull/433/files#diff-0c759ef95ed80b40c6265308877d51f3d9ac956dd5b738c5b5d4e2aac8ab38eeR7-R8).

RobertBolender: Thanks.
Strong recommendation: have that component *also* extend `Box` instead of `div`, and have the top-level component pass along it's width/height props to that component as well. You'll also be able to use styled-system props for what is currently in the `style` prop.

zachsnoek:
I added a "primary" variant to match [this](https://app.zeplin.io/project/5b10273d413156335e1b6743/screen/5cc8b19387477603a84fa4c0) spec and what's currently in ChMS. If this isn't an appropriate use of a variant I can remove it.

RobertBolender:
> I added a "primary" variant to match [this](https://app.zeplin.io/project/5b10273d413156335e1b6743/screen/5cc8b19387477603a84fa4c0) spec and what's currently in ChMS. If this isn't an appropriate use of a variant I can remove it.

The concept of a `variant` assumes that there is more than one. We currently only have one style, so I would say that adding a `variant` prop is premature.

zachsnoek:
The dashed border is now being created with this tool: https://kovart.github.io/dashed-border-generator/. If this seems reasonable for now, I can try and make it work better with the `border` `styled-system` prop.

PR #435, opened 2021-01-18
Select all text on focus by default in NumberInput

tywmick:
Resolves [FLCOM-6239](https://faithlife.atlassian.net/browse/FLCOM-6239), where it was requested that all number inputs select all text on focus by default. More discussion with @AuresaNyctea in the comments in Jira.

Seems like a breaking change, but... should it be considered a "bug fix" since we want this functionality across the entire product line? ¯\_(ツ)_/¯

### DIFF
--- a/components/input/Input.jsx
+++ b/components/input/Input.jsx
@@ -100,6 +100,10 @@ Input.propTypes = {
 	/** Large variation (deprecated in favor of the variant prop) */
 	large: PropTypes.bool,
 	size: PropTypes.number,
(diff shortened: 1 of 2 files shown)


tywmick:
Here's the [demo](https://deploy-preview-435--faithlife-styled-ui.netlify.app/#/text-input/input?a=numberinput) of the updated `NumberInput`, by the way.

RobertBolender (approved):

PR #436, opened 2021-01-19
Move styling into styled-ui.

randallknutson:
This moves styling into styled-ui from the Lookup component.


randallknutson:
No longer moving to styled-ui.

PR #437, opened 2021-01-21
Fix padding for AvatarOption that gets a larger default from react-select

randallknutson:


components/text-input-v2/select.jsx
```diff
@@ -162,7 +162,7 @@ const AvatarOption = ({ data, selectProps: { showCheckboxes, getIsOptionChecked
 	const { avatar, label, secondaryLabel } = data;
 
 	return (
-		<Styled.AvatarOption {...props}>
+		<Styled.AvatarOption {...props} theme={theme}>
```
randallknutson: I'm using `styled` here instead of `theme` since react-select components are already passing in a `theme` prop which will interfere.

components/text-input-v2/select.jsx
```diff
@@ -162,7 +162,7 @@ const AvatarOption = ({ data, selectProps: { showCheckboxes, getIsOptionChecked
 	const { avatar, label, secondaryLabel } = data;
 
 	return (
-		<Styled.AvatarOption {...props}>
+		<Styled.AvatarOption {...props} styled={theme}>
```
randallknutson, line 165: I'm using `styled` here instead of `theme` since react-select components are already passing in a `theme` prop which will interfere.

RobertBolender: That would make for a good explanatory code comment as well 👍

PR #439, opened 2021-02-02
More Flexible date period picker

bryankenote:
I based this on a commit pretty far back (like 60 commits or something) so that this could be tested locally in the catalog.

The main changes needed here for chms are
1. Ability to disable the calendar input (`DatePicker`). I just added the ability to toggle the other inputs as well just incase.
2. Ability to distinguish between when the user selects an enumerated date period (`last 7 days`, `this month`, etc...) and when they enter a custom date range. This is so that when saving these ranges to report records, we know whether to save the date range as an explicit range, or recalculate the range next time they open the report. For example, if they create a report for records with birthdays `this month`, they are likely wanting to see birthdays for the current month each time they open the report. If they select a specific range, they probably don't want the range to change the next time they open the report.

### DIFF
--- a/components/date-period-picker/component.jsx
+++ b/components/date-period-picker/component.jsx
@@ -40,6 +40,7 @@ export class DatePeriodPicker extends PureComponent {
 		dateFunctions: dateFunctionProps,
 		/** Debounce value for date inputs. Defaults to 500ms */
 		debounce: PropTypes.number,
+		disableCustomRange: PropTypes.bool,
 		disableCalendar: PropTypes.bool,
 	};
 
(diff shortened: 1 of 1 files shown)


bryankenote:
@RobertBolender Is anyone free to review this. https://github.com/Faithlife/styled-ui/pull/439/commits/89768a66a6043685191865c438d8b757b5689d1c (2 in my original description) Is really the important piece here. If we don't want to allow the separate inputs to be disabled, that's not too big of a deal.

RobertBolender:
I can take a look at this early next week 👍

PR #442, opened 2021-02-19
Add v6 upgrade notes for Button and Select

tywmick:
A few extra breaking changes I found in the process of [FaithlifeEquipment#255](https://git.faithlife.dev/Logos/FaithlifeEquipment/pull/255).

Is the new Unicode table a bit overkill? I found that Catalog could handle neither [GFM tables](https://github.github.com/gfm/#tables-extension-) nor HTML tables, so this was the best I could do. If it's too weird I could just summarize in some additional multilevel bullet points.


RobertBolender:
Love the table 👍

PR #446, opened 2021-02-25
Add a few more notes to the Upgrade Guide

tywmick:
Clearing up the `/v6` import confusion and also adding a breaking change in `Popover` and `Tooltip` I found during [FaithlifeEquipment#265](https://git.faithlife.dev/Logos/FaithlifeEquipment/pull/265).

### DIFF
--- a/UpgradeGuide.md
+++ b/UpgradeGuide.md
@@ -46,6 +46,8 @@
 - `ParameterSentence`
   - Subcomponents have been renamed: `ParameterSelect` to `ParameterSentence.Select`, `ParameterInputBox` to `ParameterSentence.Input`.
   - The `styleOverrides` prop has been removed in favor of Styled System props.
+- `Popover`
+  - The `styleOverrides` prop has been removed in favor of Styled System props.
 - `Radio`
   - The old `theme` prop functionality has been replaced by the [global theme object](https://faithlife.github.io/styled-ui/#/theme/usage) and Styled System props. Use the [`ThemeProvider` component](https://faithlife.github.io/styled-ui/#/theme/customization) to customize the theme.
 - `SimpleToast`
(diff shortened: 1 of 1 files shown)


UpgradeGuide.md
```diff
@@ -4,6 +4,8 @@
 
 - Installation
   - Styled-UI v6 should be installed as a **direct** dependency, not a peerDependency anymore.
+- Importing
+  - Importing from `'@faithlife/styled-iu/v6'` has been deprecated—you can import v6 components from `'@faithlife/styled-ui'` now.
```
RobertBolender, line 8: Let's word this more strongly, "deprecated" can have the meaning "it gives a console.warning", but in this case the whole entrypoint has been *removed* and attempts to import from `/v6` will throw.

tywmick: Oops, you're right! I wrote "When v6 is released, the `/v6` entrypoint will continue to be supported with a deprecation warning until v7 is released" so many times [in the v5 documentation](https://v5--faithlife-styled-ui.netlify.app/#/modal/documentation/v6) that I didn't think to check if `/v6` was still there.

RobertBolender: Ah right, and I forgot to explicitly tell you that we *weren't* doing that. We should also update the v5 documentation (deployed from the `v5` branch) to clarify this as well.

UpgradeGuide.md
```diff
@@ -4,6 +4,8 @@
 
 - Installation
   - Styled-UI v6 should be installed as a **direct** dependency, not a peerDependency anymore.
+- Importing
+  - Importing from `'@faithlife/styled-iu/v6'` has been removed—import v6 components from `'@faithlife/styled-ui'` now.
```
RobertBolender, line 8:
```suggestion
  - Importing from `'@faithlife/styled-ui/v6'` has been removed—import v6 components from `'@faithlife/styled-ui'` now.
```

tywmick: 😂🙏🏼

PR #457, opened 2021-03-20
Add menuItem text container config component.

korbinancell:
Not sure if I super like this solution. With the config children pattern we are kinda boxed in though.

### DIFF
--- a/components/utils/get-config-props.js
+++ b/components/utils/get-config-props.js
@@ -9,7 +9,8 @@ import React from 'react';
  */
 export function getConfigProps(children, name) {
 	const childrenList = React.Children.toArray(children);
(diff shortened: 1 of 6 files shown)


components/utils/get-config-props.js
```diff
@@ -9,7 +9,8 @@ import React from 'react';
  */
 export function getConfigProps(children, name) {
 	const childrenList = React.Children.toArray(children);
-	return childrenList.find(child => child.type.childConfigComponent === name)?.props ?? {};
+	console.log(childrenList);
```
mdombrowski1997, line 12: Was this `console.log` intentionally left in?

PR #464, opened 2021-04-08
Add MultiButton implementation to replace SegmentedButtonGroup

EthanRutherford:


theme/brandOverrides/proclaim.js
```diff
@@ -36,6 +36,12 @@ export const theme = {
 
 			segmentedButtonGroupBorder: '#494949',
 			segmentedButtonGroupBackground: '#7a7a7a',
+
+			multi: {
+				border: '#494949',
+				selectedBackground: '#0097FF',
+				selectedForeground: 'foregroundPrimary',
```
EthanRutherford, line 43: I don't know how to test the proclaim theme to verify these colors are correct.

components/menu/menu-toggle.jsx
```diff
@@ -28,14 +29,15 @@ export function MenuToggle({ hideCarrot, size, variant, disabled, children, ...b
...
 			{React.cloneElement(actionChild, {
 				defaultSize: size,
 				defaultVariant: variant,
 				defaultDisabled: disabled,
+				isMulti: true,
 			})}
-			<Button
-				ref={splitRef}
+			<MultiButton
+				ref={toggleRef}
```
EthanRutherford, line 40: this change (splitRef -> toggleRef) fixes a bug when closing the dropdown, because `toggleRef.current` was null.

EthanRutherford: It also would appear no one is using this codepath, else we would have definitely heard about it by now.

RobertBolender: We're now attaching `toggleRef` to two different components? That seems incorrect at first glance 🤔

EthanRutherford: Oh, weird, I totally didn't notice it was being set on `SegmentedButtonGroup` before.. but the ref was definitely null during event handlers before. I'll look at this again.

RobertBolender (approved):

components/menu/menu-toggle.jsx
RobertBolender: I'm not reproing any crashes on the existing MenuToggle docs, where were you able to repro?
https://faithlife.github.io/styled-ui/#/menu/variations

EthanRutherford: Very strange, I can no longer reproduce anywhere either.. However, after an edit to only use the ref on the actual button, the behavior is still identical to what we expect it to be.

I can change it back if you'd like, but this does simplify things a little bit.

RobertBolender: LGTM 👍

RobertBolender (approved):

catalog/button/variations.md
```diff
@@ -175,7 +175,49 @@ state: { loading: false }
...
+## Button Select
+
+A common use of MultiButtons is for a "select" like group of buttons serving as options
+
+```react
+showSource: true
+state: { value: 0 }
+---
+<ButtonDemo>
+	<ButtonSelect
```
korbinancell, line 203: If we are going to re-write the button groups we should use the correct underlying tags. These are acting as a radio/ checkbox group so they should either be that element, or at the very least have the correct roles to identify them.

korbinancell (changes requested):

korbinancell (approved):
Rad, thanks.

PR #468, opened 2021-04-12
MultiButtons (v5)

RobertBolender:


catalog/button/variations-v6.md
```diff
@@ -167,22 +167,63 @@ state: { loading: false }
...
+
+## Button Groups (deprecated)
+
+Prefer using `MultiButton` or `ButtonSelect`
 
 ```react
 showSource: true
 ---
 <ButtonDemo>
-	<SegmentedButtonGroup>
```
EthanRutherford, line 176: You lost the SegmentedButtonGroup here? Any reason why?

EthanRutherford: Wait, all of it changed.. What happened here?

RobertBolender: Ah, this got auto-merged with an example from v5, fixing.

PR #473, opened 2021-04-16
Add `onStep` handler to `NumberInput` (v6)

tywmick:
A v6 version of #472. Not urgent since I only need the v5 version for my current case. Just a callback prop that runs when either step button is pressed.


RobertBolender:
Closing this in favor of merging the v5 one back into master.

PR #475, opened 2021-04-26
Rename "carrot" to "caret" in `Menu.Toggle`

tywmick:
### [FLCOM-7075](https://faithlife.atlassian.net/browse/FLCOM-7075)

Just a minor spelling error that bugged me while I was working in this space. `Menu.Toggle`'s `hideCarrot` prop may confuse some people, though, so I've deprecated it in favor of a new `hideCaret` prop. But it won't break anything because it'll only give a console warning for `hideCarrot`.


korbinancell (approved):
🥕

RobertBolender (approved):

PR #476, opened 2021-04-26
Send style props to `Menu.Toggle` container when appropriate

tywmick:
### [FLCOM-7017](https://faithlife.atlassian.net/browse/FLCOM-7017) part 1

When `Menu.ActionButton` is being used, a lot of `Menu.Toggle` props are currently being sent to the caret button that ought to be sent to the container. For example, one would expect `<Menu.Toggle width="100%">` to only affect the container's width, but right now that would just make the caret button on the right comically wide.

Adding prop types to several components in the first commit helped me determine which style props needed to be filtered.

### DIFF
--- a/components/menu/menu-toggle.jsx
+++ b/components/menu/menu-toggle.jsx
@@ -1,7 +1,7 @@
 import React, { useMemo } from 'react';
 import PropTypes from 'prop-types';
 import styledSystemPropTypes from '@styled-system/prop-types';
-import { deprecateProp, getConfigChild } from '../utils';
+import { deprecateProp, filterProps, getConfigChild } from '../utils';
 import { Button, MultiButton } from '../button';
 import { Box } from '../Box';
 import { useDropdownContext, useKeyboardActivate } from './utils';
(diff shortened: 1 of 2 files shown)


components/menu/menu-toggle.jsx
```diff
@@ -87,6 +104,7 @@ export function MenuActionButton({
 			size={defaultSize}
 			variant={defaultVariant}
 			disabled={defaultDisabled}
+			flexGrow={1}
```
tywmick, line 107: Regarding this line and giving `Box` `display="inline-flex"` in line 45, setting the width of `Menu.Toggle` does nothing if `Menu.ActionButton` isn't using `flex-grow`. This can be configured manually without this change, but I think this is a more sensible default. For example, one would expect

```jsx
<Menu.Toggle width="200px">
  <Menu.ActionButton>Send</Menu.ActionButton>
</Menu.Toggle>
```

to visually span 200 pixels, but without this change, you'd have to write

```jsx
<Menu.Toggle width="200px" display="inline-flex">
  <Menu.ActionButton flexGrow={1}>Send</Menu.ActionButton>
</Menu.Toggle>
```

josephdangerstewart (approved):

korbinancell (approved):
LGTM. Thanks for taking the time to add the prop types.

tywmick:
Gotta just rebase for #475 first...

PR #478, opened 2021-05-03
Make `Select` components clearable by default

tywmick:
### [FLCOM-6739](https://faithlife.atlassian.net/browse/FLCOM-6739)

It was noted as confusing that clearing a `Select` component leaves an old value in the input. After investigating the [React Select docs](https://react-select.com), I'm guessing that such components weren't clearable at all, since [its `Select` components have](https://react-select.com/props#select-props) an `isClearable` prop that is falsy by default.

This PR adds an explicit `isClearable` prop to all Styled UI `Select` components and gives it a default value of `true`.

Probably the easiest way to test this would be to ~check out the branch, then `yarn && yarn build && yarn catalog-start`, then go to http://localhost:4000/#/text-input/select~ _go to https://deploy-preview-478--faithlife-styled-ui.netlify.app/#/text-input/select_.

### DIFF
--- a/components/text-input/select.jsx
+++ b/components/text-input/select.jsx
@@ -5,6 +5,7 @@ import { colors } from '../shared-styles';
 import { ChevronDown } from '../icons/12px';
 import { Checkbox } from '../../components/check-box';
 import { DebouncedSelectAsync, DebouncedSelectAsyncCreatable } from './debounced-async';
+import { useClearableSelectValue } from './use-clearable-select-value';
 import * as Styled from './styled';
 import { ThemedBox } from '../ThemedBox';
 import { useTheme } from '../../theme';
@@ -281,14 +282,36 @@ export function useCommonSelectProps(props, ref) {
 		components,
 		disabled,
 		isDisabled,
-		isClearable = true,
 		styles,
 		onKeyDown,
+		value,
+		defaultValue,
+		onChange,
+		inputValue,
+		defaultInputValue,
+		onInputChange,
+		isClearable = true,
+		isMulti,
 		...otherProps
 	} = props;
 
 	const body = useBody();
 	const theme = useTheme();
+	const {
+		innerValue,
+		innerOnChange,
+		innerInputValue,
+		innerOnInputChange,
+	} = useClearableSelectValue({
+		value,
+		defaultValue,
+		onChange,
+		inputValue,
+		defaultInputValue,
+		onInputChange,
+		isClearable,
+		isMulti,
+	});
 
 	return {
 		ref: ref,
@@ -298,9 +321,14 @@ export function useCommonSelectProps(props, ref) {
 		noOptionsMessage: noOptionsMessage,
 		menuPortalTarget: body,
 		isDisabled: disabled || isDisabled,
-		isClearable: isClearable,
 		styles: selectStyles(props, theme),
 		onKeyDown: e => handleKeyDown(e, onKeyDown),
+		value: innerValue,
+		onChange: innerOnChange,
+		inputValue: innerInputValue,
+		onInputChange: innerOnInputChange,
+		isClearable: isClearable,
+		isMulti: isMulti,
 		...otherProps,
 	};
 }
--- a/components/text-input/select.jsx
+++ b/components/text-input/select.jsx
@@ -242,111 +242,68 @@ function handleKeyDown(e, onConsumerKeyDown) {
 export { reactSelectComponents };
 
 /** Autocomplete control based on react-select */
-export const Select = React.forwardRef(
-	({ components = {}, disabled, isDisabled, isClearable = true, ...props }, ref) => {
-		const body = useBody();
-		const onConsumerKeyDown = props.onKeyDown;
-
-		const theme = useTheme();
-
-		return (
-			<ReactSelect
-				ref={ref}
-				classNamePrefix="fl-select"
-				theme={selectTheme}
-				components={{ ...defaultComponents, ...components }}
-				noOptionsMessage={noOptionsMessage}
-				menuPortalTarget={body}
-				isDisabled={disabled || isDisabled}
-				isClearable={isClearable}
-				{...props}
-				styles={selectStyles(props, theme)}
-				onKeyDown={e => handleKeyDown(e, onConsumerKeyDown)}
-			/>
-		);
-	},
-);
+export const Select = React.forwardRef((props, ref) => {
+	const transformedProps = useCommonSelectProps(props, ref);
+	return <ReactSelect {...transformedProps} />;
+});
 
 /** The same as `Select`, but allows new entries. */
-export const CreatableSelect = React.forwardRef(
-	({ components = {}, disabled, isDisabled, isClearable = true, ...props }, ref) => {
-		const body = useBody();
-		const onConsumerKeyDown = props.onKeyDown;
-
-		const theme = useTheme();
-
-		return (
-			<ReactSelectCreatable
-				ref={ref}
-				classNamePrefix="fl-select"
-				theme={selectTheme}
-				formatCreateLabel={node => <span>New entry: {node}</span>}
-				components={{ ...defaultComponents, ...components }}
-				noOptionsMessage={noOptionsMessage}
-				menuPortalTarget={body}
-				isDisabled={disabled || isDisabled}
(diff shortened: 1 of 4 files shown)


components/text-input/select.jsx
```diff
@@ -360,16 +356,3 @@ function useBody() {
...
-function useLegacyChangeHandler(onChange, isMulti) {
-	return useCallback(
-		(values, change) => {
-			if (onChange) {
-				return onChange(isMulti ? values || [] : values, change);
-			}
-		},
-		[isMulti, onChange],
-	);
-}
```
tywmick, line 375: This Hook reverted a breaking change introduced in React Select v3, which it looks like we upgraded to while Styled UI was still in v5. There's a TODO note on it to remove the Hook in v6, and we're already warning about this breaking change in the [v6 upgrade guide](https://faithlife.github.io/styled-ui/#/upgrade-guide), so I think it's best to remove it now. This isn't essential to this PR, though, so I can revert this commit if desired. I do think it's a worthwhile change, though (see the explanation in the [React Select v3 upgrade guide](https://github.com/JedWatson/react-select/issues/3585)).

If this commit is merged, I'll post a special notice in the Styled UI flow.

RobertBolender:
> Probably the easiest way to test this

FYI, we have a PR build environment 😄 
You can find the link under "Show all checks"
https://deploy-preview-478--faithlife-styled-ui.netlify.app/

tywmick:
Oh that's much better! 😅

tywmick:
I've added functionality to clear the `Select` value when the user is typing (for clearable, single-value `Select`s). Probably ventured into "too clever" territory by the end of it, but I was never stuck, and it always felt like I was nearly there... But at least this was just one day of work this time. I'll use this to continue training my forecasting skills—ideally I should've pushed back about this extra functionality being necessary in a spring-cleaning sprint. And stopped at some point to take the time to figure out how to explain what my plan was.

tywmick:
This is ready for another look, by the way, @RobertBolender 👍🏼

RobertBolender (approved):
I'm not *totally* convinced that we want to support the optionally-controlled-state pattern/hook long-term, but I'm a big fan of the `useCommonSelectProps` cleanup and am very happy with the UX improvement with clearing the Single Select when typing.

Let's be sure to merge this with Rebase, not Squash, to preserve the individual commits.

components/text-input/use-clearable-select-value.js
```diff
@@ -0,0 +1,49 @@
...
+
+			setInputValue(newInputValue);
+			if (onInputChange) {
+				onInputChange(newInputValue, meta);
+			}
+		},
+		[innerOnSelectChange, isClearable, isMulti, onInputChange, setInputValue],
+	);
+
+	return [selectValue, innerOnSelectChange, inputValue, innerOnInputChange];
```
josephdangerstewart, line 48: Nit (non-blocking): I realize this hook is pretty tightly coupled to the selects so it's not a huge deal although can we return an object here? My general rule of thumb is to prefer returning objects from hooks when

1. There are more than two values being returned
2. The hook is likely to be called once and only once in a component

My reasoning for this is so that A) consumers are encouraged to use consistent variable names and B) consumers don't have to worry about getting the order correct when destructuring

components/utils/use-optionally-controlled-state.js
```diff
@@ -0,0 +1,25 @@
...
+ * @param {any} defaultValue - An optional default value to initialize the state with when the value
+ * is not externally controlled.
+ * @returns {[any, (newValue?: any) => void]} An array of the state value and either a setter
+ * function or (if the value is externally controlled) an empty function.
+ */
+export function useOptionallyControlledState(controlledValue, defaultValue) {
+	const [value, setValue] = useState(
+		controlledValue !== undefined ? controlledValue : defaultValue,
+	);
+	useEffect(() => {
```
josephdangerstewart, line 18: Nit (non-blocking): I think this hook could be simplified a bit

```js
const isControlled = controlledValue !== undefined;

const [value, setValue] = useState(isControlled ? controlledValue : defaultValue);

if (isControlled) {
    return [controlledValue, EMPTY_FUNCTION];
}

return [value, setValue];
```

josephdangerstewart: This way saves you a `useEffect` call and a few lines of code

josephdangerstewart: Nice utility hook by the way!

josephdangerstewart (approved):

components/text-input/use-clearable-select-value.js
tywmick: Good call! 👌🏼

components/utils/use-optionally-controlled-state.js
tywmick: Oh yeah, I suppose that `useEffect` is a bit useless there. 😅 Thanks! I think I'll simplify it a bit more to

```js
const [value, setValue] = useState(defaultValue);

if (controlledValue !== undefined) {
  return [controlledValue, EMPTY_FUNCTION];
} else {
  return [value, setValue];
}
```

then, since `value` and `setValue` won't be used if `controlledValue`'s defined.

tywmick:
As we'll be rebasing instead of squashing for this one, I'll edit the last two commits instead of adding more so the git history's more useful.

components/text-input/use-clearable-select-value.js
```diff
@@ -0,0 +1,54 @@
...
+	onSelectChange,
+	controlledInputValue,
+	onInputChange,
+	{ defaultValue: defaultSelectValue, defaultInputValue, isClearable, isMulti },
+) {
+	const [selectValue, trySetSelectValue] = useOptionallyControlledState(
+		controlledSelectValue,
+		defaultSelectValue,
+	);
+	const [inputValue, trySetInputValue] = useOptionallyControlledState(
```
tywmick, line 19: @RobertBolender, I renamed these setter functions to `trySetValue` instead of just `setValue`, since they don't actually set the value if it's controlled. That makes me feel a lot better about `useOptionallyControlledState` as a whole.

Reason I think I need the optionally-controlled-state pattern is that I couldn't get clearing the select value when typing to work if I didn't at least internally control both the `innerValue` and `value` (the [Action Meta](https://react-select.com/advanced#action-meta) example from the React Select docs helped me with this solution at some point). And if I only allow internal control of those values, it'll break any consumer's attempts to control externally with those two props, which was previously allowed.

Do you think this Hook would be better placed in the `text-input` folder rather than `utils`?

RobertBolender: I don't feel strongly enough about it to suggest that you move it to `text-input`, but if you felt inclined to do so I would also not object. 😄

components/text-input/select.jsx
```diff
@@ -242,115 +243,91 @@ function handleKeyDown(e, onConsumerKeyDown) {
...
+	} = props;
+
+	const body = useBody();
+	const theme = useTheme();
+	const {
+		innerValue,
+		innerOnChange,
+		innerInputValue,
+		innerOnInputChange,
+	} = useClearableSelectValue(value, onChange, inputValue, onInputChange, {
```
RobertBolender, line 305: Super nit: how would you feel about updating this hook to take a single object parameter, rather than 4 positional parameters and then an object parameter?

I'm not immediately seeing what value we get from having the 5 parameters instead of 1.

tywmick: I was thinking of the object parameter as a configuration object and the first four parameters as the "important" ones since they're the ones being transformed (when you look at the return object). But since this Hook is so specialized, it's not like any of the "configuration options" actually optional, so I'll go make the change. 👍🏼

components/text-input/use-clearable-select-value.js
```diff
@@ -0,0 +1,57 @@
...
+export function useClearableSelectValue({
+	value: controlledSelectValue,
+	defaultValue: defaultSelectValue,
+	onChange: onSelectChange,
+	inputValue: controlledInputValue,
+	defaultInputValue,
+	onInputChange,
+	isClearable,
+	isMulti,
+}) {
```
tywmick, line 17: @RobertBolender, new object parameter with name changes that make more sense inside the Hook. 👍🏼

Issue #467, opened 2021-04-10
Tooltip closes when mouse moves over tooltip content itself


tywmick:
### _Original description under the title "Add `delay` prop to v6 `Popover`":_

Not crucial, but the v5 `Popover` had a [`delay` prop](https://github.com/Faithlife/styled-ui/blob/e75309a44de6eac24c1b9aeef30790eac02014b1/components/popover/popover-base.jsx#L47-L48) that allowed the consumer to specify numbers of milliseconds to delay showing and/or hiding the popover. I'm having to lose this feature as I'm updating Tasks to Styled UI v6 ([Tasks#187](https://git.faithlife.dev/Logos/Tasks/pull/187)).

It looks like `Tooltip` still has a built-in 200ms delay, using a `useDebouncedCallback` hook, so it would probably be fairly simple to adapt that pattern for the new `Popover`:

https://github.com/Faithlife/styled-ui/blob/5dfe1dfcaa866f2309c08a53980b01cda5230bc8/components/popover/tooltip.jsx#L14-L26

EthanRutherford:
The new popover is "lower level" than the old one. The consumer controls how and when to show the popover, and can achieve a delay by simply debouncing their own show/hide state transitions.

However, I'd also question what use case there is for a popover with a delayed open, other than a tooltip? If the user clicks a button to open the popover, it should be immediate.

From reading your comment in the linked tasks PR, it sounds like what you really are after *is* a tooltip, it's just that if the mouse enters the tooltip content itself, the tooltip closes. 
I might opine that that's actually a bugfix/improvement to tooltips that we should make, rather than adding any state management to popovers.

tywmick:
Excellent analysis, that's a much better solution! Should I create a new issue and close this one or just change the title of this issue?

EthanRutherford:
IMO, you can keep this one and change the title. Don't know if @RobertBolender has a different opinion, though.

RobertBolender:
To ensure I'm understanding the updated recommendation: We want to add a `delay` prop to the `Tooltip` component, rather than to `Popovers`? No objection to renaming and re-using this issue 👍

EthanRutherford:
[...]

Issue #450, opened 2021-03-12
Tooltip is missing from v6 docs


tywmick:
Don't have the time at the moment, but here's where it lives [in the v5 docs](https://v5--faithlife-styled-ui.netlify.app/#/popover/variations?a=tooltip).

PR #481, opened 2021-05-06
Tooltip improvements

EthanRutherford:
Addresses #450 and #467. Also adds delay prop to tooltip.

### DIFF
--- a/components/popover/tooltip.jsx
+++ b/components/popover/tooltip.jsx
@@ -1,40 +1,55 @@
-import React, { useState, useEffect, useRef } from 'react';
+import React, { useState, useEffect, useRef, useCallback } from 'react';
 import PropTypes from 'prop-types';
-import { useDebouncedCallback } from '../shared-hooks'; // Copied from react-ui. Import from there when made available.
 import { Popover } from './popover';
 import { Box } from '../Box';
 
+function useDelayedHoverState(delay) {
+	const [isHovered, setIsHovered] = useState(false);
+	const timeouts = useRef({ isHovered: false });
+
+	const cancel = useCallback(() => {
+		clearTimeout(timeouts.current.enterTimeout);
+		clearTimeout(timeouts.current.leaveTimeout);
+	}, []);
+
+	const onMouseEnter = useCallback(() => {
+		cancel();
+		if (!isHovered) {
+			timeouts.current.enterTimeout = setTimeout(() => setIsHovered(true), delay);
+		}
+	}, [cancel, delay, isHovered]);
+
+	const onMouseLeave = useCallback(() => {
+		cancel();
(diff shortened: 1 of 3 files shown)


components/popover/tooltip.jsx
```diff
@@ -62,6 +82,8 @@ Tooltip.propTypes = {
 	...Popover.propTypes,
 
 	children: PropTypes.oneOfType([PropTypes.node, PropTypes.func]).isRequired,
+	/** Hover delay in milliseconds */
+	delay: PropTypes.number,
```
RobertBolender, line 86: Should we set a default for this prop (ideally in the destructuring of props, not defaultProps), or does the setTimeout do the intelligent thing with a nullish delay argument?

components/popover/tooltip.jsx
```diff
@@ -1,40 +1,55 @@
...
+
+	const cancel = useCallback(() => {
+		clearTimeout(timeouts.current.enterTimeout);
+		clearTimeout(timeouts.current.leaveTimeout);
+	}, []);
+
+	const onMouseEnter = useCallback(() => {
+		cancel();
+		if (!isHovered) {
+			timeouts.current.enterTimeout = setTimeout(() => setIsHovered(true), delay);
```
RobertBolender, line 18: Will this `setIsHovered` setState cause a react warning if the timeout finishes after the component has unmounted? Do we need to call `cancel` in unmount somewhere to avoid that leak?

components/popover/tooltip.jsx
EthanRutherford: I believe nullish would be treated as zero, but I don't mind making that explicit. Also, open question, should we use a *non*zero default?

components/popover/tooltip.jsx
EthanRutherford: We do! see the useEffect below.

RobertBolender: Ah perfect, I had missed that 👍

components/popover/tooltip.jsx
RobertBolender: Explicit zero sounds fine to me for now 👍

tywmick: Current delay is [200ms](https://github.com/Faithlife/styled-ui/blob/3d7257f52c3d941845634906deb7b33d0c52dca7/components/popover/tooltip.jsx#L21), yeah? I think that feels like a nice default. 👍🏼

EthanRutherford: That delay is actually only on close, and is there to make sure we don't close if the mouse (briefly) leaves then reenters the hover area. Current behavior is immediate.

PR #485, opened 2021-05-14
Jest!

korbinancell:
Will probably need some configuration in netlify as well. Not sure.

There are quite a few todos in here. This pr is already way too big, so I left aXe compliance and some component reconfiguring for a later pr.

If I couldn't test that the component worked I at least tested that it existed.


RobertBolender:
FYI: `render`, `axe`, and `screen` are not exported from `test-utils`
https://app.netlify.com/sites/faithlife-styled-ui/deploys/609ef25624017600081d8fa4

korbinancell:
Whoops, yea that was me being lazy. I just need to remove that file entirely and import from the different packages directly.

components/__tests__/import-tests.test.js
```diff
@@ -0,0 +1,7 @@
+describe('components', () => {
+	it.skip('should not crash when imported', () => {
```
rmjohnson, line 2: Do we really need this test?

components/__tests__/layout-primitives.test.js
```diff
@@ -0,0 +1,13 @@
+import { Box } from '../Box';
+import { Stack } from '../Stack';
+import { Text } from '../Text';
+
+describe('Layout Primitives', () => {
```
rmjohnson, line 5: You might consider moving these to their own test files to maintain `component-name.test.js(x)` pattern to make the tests easier to find.

components/__tests__/switch.test.jsx
```diff
@@ -0,0 +1,28 @@
...
+
+		const callback = jest.fn();
+		render(<Switch onClick={callback} isChecked={false} />);
+
+		userEvent.click(screen.getByRole('switch'));
+
+		expect(callback).toHaveBeenCalledTimes(1);
+	});
+
+	// TODO: make aXe compatible
```
rmjohnson, line 18: nit: you might consider making a github issue for this.

components/accordion/__tests__/accordion.test.jsx
```diff
@@ -0,0 +1,76 @@
...
+			for (const section of [0, 1, 2]) {
+				if (expandedSections.includes(section)) {
+					expect(screen.queryByText(`Section ${section + 1} Text`)).toBeVisible();
+				} else {
+					expect(screen.queryByText(`Section ${section + 1} Text`)).not.toBeVisible();
+				}
+			}
+
+			userEvent.click(screen.getByText(clickedSection));
+			expect(onExpansion.mock.calls[0][0]).toEqual(expectedExpandedSections);
```
rmjohnson, line 48: `calls[0][0]` feels a bit mysterious here to me. Are you looking for `toHaveBeenCalledWith`?

components/check-box/__tests__/check-box.test.jsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import { axe } from 'jest-axe';
+import userEvent from '@testing-library/user-event';
+
+import { Checkbox } from '../';
+
+describe('Checkbox', () => {
+	it('should respond to user click', () => {
```
rmjohnson, line 9: Another couple of easy tests might be setting `isChecked` and asserting the underlying checkboxes are indeed checked / unchecked.

components/help-box/component.jsx
```diff
@@ -51,7 +51,7 @@ export function HelpBox({
 					))}
 				<Styled.HelpBoxContent>{children}</Styled.HelpBoxContent>
 				{(handleClose && (
-					<Styled.CloseButton onClick={handleClose}>
+					<Styled.CloseButton aria-label="Close" onClick={handleClose}>
```
rmjohnson, line 54: :golfclap:

components/loading-spinner/__tests__/loading-spinner.test.jsx
```diff
@@ -0,0 +1,9 @@
+import { LoadingSpinner } from '../';
+
+describe('LoadingSpinner', () => {
+	it('is truthy', () => {
```
rmjohnson, line 4: What? No interaction tests? :trollface:

components/parameter-sentence/__tests__/parameter-sentence.test.jsx
```diff
@@ -0,0 +1,119 @@
...
+		expect.hasAssertions();
+
+		const callback = jest.fn();
+		render(
+			<ParameterSentence accessibilityFormLabel="Tithe Calculator">
+				<ParameterSentence.Input
+					defaultValue="10"
+					value={''}
+					onChange={event => callback(event.target.value)}
+					// formatValue={val => `${val}%`}
```
rmjohnson, line 36: commented out code

components/__tests__/switch.test.jsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { render, axe, screen, userEvent } from '../../test-utils';
```
rmjohnson, line 2: Missed this one =)

components/parameter-sentence/__tests__/parameter-sentence.test.jsx
```diff
@@ -0,0 +1,119 @@
...
+describe('ParameterSentence', () => {
+	it('and its children are truthy', () => {
+		expect.hasAssertions();
+
+		expect(ParameterSentence).toBeTruthy();
+		expect(ParameterSentence.Input).toBeTruthy();
+		expect(ParameterSentence.Select).toBeTruthy();
+	});
+
+	it('s Input is interact-able', () => {
```
rmjohnson, line 26: I think I see what you're going for here but unfortunately since the test output doesn't seem to include "it" it looks a little weird / confusing.

image

Maybe just `Input is..`?

components/parameter-sentence/__tests__/parameter-sentence.test.jsx
```diff
@@ -0,0 +1,119 @@
...
+		const callback = jest.fn(val => `${val}%`);
+		render(
+			<ParameterSentence accessibilityFormLabel="Tithe Calculator">
+				<ParameterSentence.Input
+					defaultValue="10"
+					value={''}
+					onChange={() => {}}
+					formatValue={callback}
+					width="30px"
+					accessibilityLabel={'Percent of income to tithe'}
```
rmjohnson, line 60: Would be good to include this label in the `getByRole` call to add some coverage for it.

components/parameter-sentence/__tests__/parameter-sentence.test.jsx
```diff
@@ -0,0 +1,119 @@
...
+				<ParameterSentence.Select
+					selectedId={'annual'}
+					onItemSelect={callback}
+					options={scheduleOptions}
+					accessibilityLabel={'Pay schedule of income'}
+				/>
+			</ParameterSentence>,
+		);
+
+		userEvent.click(screen.getByText('annual'));
```
rmjohnson, line 84: Do we not have an accessible role on this?

components/popover/__tests__/tooltip.test.jsx
```diff
@@ -0,0 +1,54 @@
...
+		expect(Tooltip).toBeTruthy();
+	});
+
+	it.skip('shows children', async () => {
+		expect.hasAssertions();
+
+		render(<Tooltip content="hovered!">Hover me!</Tooltip>);
+
+		userEvent.hover(screen.getByText('Hover me!'));
+		await waitFor(() => expect(screen.getByText('hovered!')).toBeVisible());
```
rmjohnson, line 31: I'd expect (heh...) we could use `expect(await screen.findByText(...` to achieve the same thing.

components/simple-toast/__tests__/simple-toast.test.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import { axe } from 'jest-axe';
+
+import { SimpleToast } from '../';
+
+describe('SimpleToast', () => {
```
rmjohnson, line 7: Consider adding a test using https://jestjs.io/docs/timer-mocks#advance-timers-by-time to test out the `showTime`.

components/slider/__tests__/slider.test.jsx
```diff
@@ -0,0 +1,23 @@
...
+import { Slider } from '../';
+
+// TODO: add interaction tests. It will require some work on the component itself.
+describe('Slider', () => {
+	it('is truthy', () => {
+		expect.hasAssertions();
+
+		expect(Slider).toBeTruthy();
+	});
+	// TODO: this is sort of an accessiblity nightmare currently
```
rmjohnson, line 14: 😬

jest.config.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+	testEnvironment: 'jest-environment-jsdom',
+	moduleNameMapper: {
+		// '\\.(css|less)$': require.resolve('./__mocks__/styleMock.js'),
```
rmjohnson, line 4: Commented out code.

package.json
```diff
@@ -14,13 +14,16 @@
 	"scripts": {
 		"deploy-site": "npm run catalog-build && storybook-to-ghpages --ci --existing-output-dir=catalog/build",
 		"eslint": "eslint ./components ./catalog --ext .js --ext .jsx",
-		"test": "npm run build && npm run unit-test && npm run eslint && npm run stylelint",
+		"test": "npm run build && npm run test:ci && npm run eslint && npm run stylelint",
```
rmjohnson, line 17: I'll defer to you but it's a little weird to me that the `test` command runs the `build` and lint.

rmjohnson (changes requested):

package.json
mdombrowski1997: ☝🏽 I believe Jest builds the code for you, so building before testing doesn't change anything

components/collapse/__tests__/collapse.test.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+
+import { Collapse } from '../';
+
+describe('Collapse', () => {
+	it('should respond to user click', () => {
```
mdombrowski1997, line 7: Is this test name misleading?  We're not testing user input or clicking, we're testing that the `isOpen` prop shows/hides the children.  Should we update the name to match what is being tested?

components/date-picker-input/__tests__/date-picker-input.test.jsx
```diff
@@ -0,0 +1,71 @@
...
+import * as dateFunctions from 'date-fns';
+import { render, screen } from '@testing-library/react';
+import { axe } from 'jest-axe';
+import userEvent from '@testing-library/user-event';
+
+import { DatePickerInput } from '../';
+
+describe('DatePeriodPicker', () => {
+	it('should parse user input', () => {
+		expect.hasAssertions();
```
mdombrowski1997, line 11: This [expect.hasAssertions()](https://jestjs.io/docs/expect#expecthasassertions) is in a lot of these tests.  It appears to offer safety when we're delegating our assertions to functions called within the test function.  However, in this (and I believe all other) tests, we call `expect` directly, so this ever-present call feels more like clutter than a help.  Should we drop these calls to `expect.hasAssertions()`?

components/date-picker/__tests__/date-picker.test.jsx
```diff
@@ -0,0 +1,92 @@
...
+				parseDate={dateFunctions.parse}
+			/>,
+		);
+
+		const firstOfMonth = dateFunctions.startOfMonth(new Date());
+		userEvent.click(screen.getAllByText('1')[0]);
+		expect(callback).toHaveBeenLastCalledWith(firstOfMonth);
+	});
+
+	// eslint-disable-next-line no-unsanitized/method
```
mdombrowski1997, line 29: This lint rule seems to be unhelpful in tests.  Calling "unsafe" code is expected, and is 0 risk since they are tests.  Should we disable this rule for all `.test.jsx` files for ease of use?

components/popover/__tests__/tooltip.test.jsx
mdombrowski1997: Does `findByText` assert that the element is visible though?  I would image `findByText` would find the element (which appears to be always present in the DOM since `getByText` isn't throwing), then if it isn't immediately visible, would fail.  I actually think I find this current version more clear as to what we're asserting (assuming I'm correct here)

components/share-dialog/__tests__/share-dialog.test.jsx
```diff
@@ -0,0 +1,39 @@
...
+			disconnect: jest.fn(),
+			observe: jest.fn(),
+			unobserve: jest.fn(),
+		}));
+	});
+
+	it('is truthy', () => {
+		expect.hasAssertions();
+
+		expect(ShareDialog).toBeTruthy();
```
mdombrowski1997, line 20: Are we asserting here that the `import` was successful?  If so, why?

mdombrowski1997: I see the same for `SequencedTab` and `Tab`.  Are we ensuring that we don't change the name or shape of our `export`s?

components/share-dialog/__tests__/share-dialog.test.jsx
```diff
@@ -0,0 +1,39 @@
...
+	beforeAll(() => {
+		window.ResizeObserver = jest.fn(() => ({
+			disconnect: jest.fn(),
+			observe: jest.fn(),
+			unobserve: jest.fn(),
+		}));
+	});
+
+	it('is truthy', () => {
+		expect.hasAssertions();
```
mdombrowski1997, line 18: This feels redundant here in particular where the entire test *is* a single assertion

components/accordion/__tests__/accordion.test.jsx
korbinancell: I had originally experimented with using snapshots to test this data vs hard coding the expected data, looks like I left this around.

components/parameter-sentence/__tests__/parameter-sentence.test.jsx
korbinancell: It appears we don't from the output in the test. This component might need to be revisited now that it has been updated to v6 🤔

components/popover/__tests__/tooltip.test.jsx
korbinancell: These tests are skipped b/c something about popper.js does not play nicely with jest. I might just have to mock popper and go from there.

components/simple-toast/__tests__/simple-toast.test.jsx
korbinancell: These are not tested b/c of some weirdness with refs in jest testing. B/c they are not rendered with react-dom the refs are never set. (Which is why I think that popovers are having issues). So I kinda just skipped this. However, now that I think about it we are using useImperativeHandle to tack on all of the methods, so maybe that will still all happen even without the working ref?

components/collapse/__tests__/collapse.test.jsx
korbinancell: copy paste got me again

components/date-picker-input/__tests__/date-picker-input.test.jsx
korbinancell: This one is kinda a force of habit. Our linking in giving land requires it (and actually found a couple of tests that weren't actually checking anything when we first added it).

components/share-dialog/__tests__/share-dialog.test.jsx
korbinancell: It is mainly to make sure that we don't lose any children of these new Component.ChildComponent style components along the way. It might be overkill normally, but for something like styled-ui with so many varied users making sure we don't lose components unexpectantly is always nice to have.

components/check-box/__tests__/check-box.test.jsx
korbinancell: There actually isn't an underlying checkbox element, it is all buttons and svgs and aria roles

rmjohnson (approved):
Love it! :shipit:

PR #492, opened 2021-06-03
Fix build package before publishing

lawrencesin:
The jenkins job and the package.json to publish `styled-ui` was missing the `npm run build` command; which resulted in the published package missing the `dist` directory.  Added it to the `prepublishOnly` step so that the package would be built properly before packing it for publishing.


RobertBolender:
For now, we've added the `yarn build` step directly in the Jenkins config.

PR #494, opened 2021-06-16
🐛 Addresses an issue in macOS Safari where stepUp/stepDown on an input would crash

JacobHaitsma:
Addresses an issue in Safari lts affecting NumberInput where stepUp/stepDown on an input would crash

[Example (affects macOS 14.0.3 (at least) - lts)](https://codesandbox.io/s/great-mccarthy-s2ybz
)
Treat empty input values as 0 when handling stepUp/stepDown


korbinancell (approved):
LGTM

PR #498, opened 2021-08-13
Fix Share Dialog CSS and add Modal feature

ELanning:


components/share-dialog/styled.jsx
```diff
@@ -45,6 +45,6 @@ export const CopyContainer = styled.div`
 export const Copied = styled.div`
 	position: absolute;
 	top: 100%;
-	right: 0;
+	right: -3px;
```
ELanning, line 48: Happy to change this to `left: 201px` if the reviewer has a grudge against negative number ;P

components/share-dialog/component.jsx
```diff
@@ -21,22 +21,37 @@ export class ShareDialog extends React.Component {
...
+			<Modal
+				isOpen={isOpen}
+				onClose={onClose}
+				container="body"
+				contentPadding={6}
+				scrollbarEnabled={scrollbarEnabled}
+			>
 				<Modal.Header title={modalTitle} textStyle="h.24" />
-				<Modal.Content>
+				<Modal.Content overflowY="visible">
```
ELanning, line 54: This fixes the `Copied!` text from being hidden.

components/modal-backdrop/component.jsx
```diff
@@ -7,22 +7,28 @@ import { Box } from '../Box';
...
 	static propTypes = {
-		/** Callback function to tell modal to close due to clicking the backdrop  */
+		/** Callback function to tell modal to close due to clicking the backdrop */
 		onClose: PropTypes.func.isRequired,
 		/** The z-index of the backdrop */
 		zIndex: PropTypes.number,
 		/** Contents of the modal */
 		children: PropTypes.node.isRequired,
+		/** Whether page's active scrollbar should be hidden when modal is active */
+		scrollbarEnabled: PropTypes.bool,
```
ELanning, line 17: On some pages, hiding the scrollbar actually makes a large portion of the page content move. This can be jarring, so this flag was added as a way around it. (At the cost of the scrollbar being visible while the modal is open).

RobertBolender: The effect of this boolean is not only that the background scrollbar is visible, but that background scrolling is enabled. Would you consider instead a prop that was `disableBackgroundScrolling`?

ELanning: `disableBackgroundScrolling` would be a breaking change I believe? Meaning already existing consumers would have to update their code and add `disableBackgroundScrolling={true}`

ELanning: Not saying we _shouldn't_ make a breaking change + major version update, but that's why I didn't go with a `disableFoo` paradigm

RobertBolender: We could default the prop to `true` (which I don't *love*) to avoid the breaking change (which I *would* like to avoid), or else use `allowBackgroundScrolling` defaulting to `false` (the default default)

ELanning: `allowBackgroundScrolling` defaulting to `false` sounds good to me 👍

RobertBolender (approved):

PR #510, opened 2021-09-08
Fix prop types

### LINKED ISSUE #509: Console errors when using ModalFooterButtons
Screen Shot 2021-09-07 at 1 52 27 PM
These errors appear in the console when opening a Modal that contains a ModalFooterButtons component. Other than the messages appearing, it doesn't seem to affect functionality. 

I believe the errors are because [the Button component expects the size prop to be oneOf small, medium, or large](https://github.com/Faithlife/styled-ui/blob/master/components/button/Button.js#L93), but [the ModalFooterButtons component passes in an array containing small and medium](https://github.com/Faithlife/styled-ui/blob/master/components/modal/modal-footer-buttons.jsx#L23).

RobertBolender:
Fixed #509.

### DIFF
--- a/components/button/Button.js
+++ b/components/button/Button.js
@@ -98,6 +98,7 @@ ButtonCore.propTypes = {
 		'primary',
 		'secondary',
 		'minor',
+		'transparent',
 		'minorTransparent',
 		'link',
 		'danger',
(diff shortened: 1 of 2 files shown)


korbinancell (approved):
That was fast. Thanks!

PR #512, opened 2021-09-15
Export everything from the underlying react-select

MitchelMyersFl:
Make everything in react-select available to consumers of this package.

### DIFF
--- a/components/text-input/select.jsx
+++ b/components/text-input/select.jsx
@@ -124,7 +124,7 @@ const selectStyles = (props, theme) => {
 				return ourStyle;
 			},
 		}),
-		{}
+		{},
 	);
 
 	return { ...customStyles, ...ourWrappedStyles };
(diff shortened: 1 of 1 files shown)


components/text-input/select.jsx
```diff
@@ -1,5 +1,6 @@
 import React, { useState, useEffect } from 'react';
 import ReactSelect, { components as reactSelectComponents } from 'react-select';
+import * as reactSelect from 'react-select';
```
korbinancell, line 3:
```suggestion
import ReactSelect, { components as reactSelectComponents, ...reactSelectFunctions } from 'react-select';
```

components/text-input/select.jsx
```diff
@@ -240,7 +241,7 @@ function handleKeyDown(e, onConsumerKeyDown) {
 	}
 }
 
-export { reactSelectComponents };
+export { reactSelect, reactSelectComponents };
```
korbinancell, line 244:
```suggestion
export { reactSelectFunctions, reactSelectComponents };
```

PR #515, opened 2021-09-28
Add DatePicker Footer support

ELanning:
Adds a new component, `DatePicker.Footer`. When passed as a child to the `DatePicker` component, it is slotted into the bottom of the `DatePicker`. This allows custom buttons and other things.

### DIFF
--- a/catalog/date-picker-input/variations.md
+++ b/catalog/date-picker-input/variations.md
@@ -8,14 +8,14 @@ noSource: true
 </React.Fragment>
 ```
 
-## Default Date Picker
+## Default Date Picker Input
 
 ```react
 showSource: true
 state: { selectedDate: null }
 ---
 <div>
-	<span>The selected date is {dateFunctions.format(state.selectedDate, 'MM-dd-yyyy')}</span>
+	<span>The selected date is {state.selectedDate && dateFunctions.format(state.selectedDate, 'MM-dd-yyyy')}</span>
 	<DatePickerInput
 		defaultSelectedDate={state.selectedDate || new Date()}
 		onChange={(date) => setState({ selectedDate: date })}
@@ -26,14 +26,14 @@ state: { selectedDate: null }
 </div>
 ```
 
-## Customized placement of Date Picker
+## Customized placement of Date Picker Input
 
 ```react
 showSource: true
 state: { selectedDate: null }
 ---
 <div>
-	<span>The selected date is {dateFunctions.format(state.selectedDate, 'MM-dd-yyyy')}</span>
+	<span>The selected date is {state.selectedDate && dateFunctions.format(state.selectedDate, 'MM-dd-yyyy')}</span>
 	<DatePickerInput
 		defaultSelectedDate={state.selectedDate || new Date()}
 		onChange={(date) => setState({ selectedDate: date })}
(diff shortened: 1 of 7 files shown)


catalog/date-picker-input/variations.md
```diff
@@ -8,14 +8,14 @@ noSource: true
...
-## Default Date Picker
+## Default Date Picker Input
 
 ```react
 showSource: true
 state: { selectedDate: null }
 ---
 <div>
-	<span>The selected date is {dateFunctions.format(state.selectedDate, 'MM-dd-yyyy')}</span>
+	<span>The selected date is {state.selectedDate && dateFunctions.format(state.selectedDate, 'MM-dd-yyyy')}</span>
```
ELanning, line 18: This seems to work without the `state.selectedDate &&` part on production, but crashes locally, so figured I'd add it regardless.

components/date-picker-input/component.jsx
```diff
@@ -152,7 +152,9 @@ export function DatePickerInput({
 								dateFunctions={dateFunctions}
 								minDate={minDate}
 								maxDate={maxDate}
-							/>
+							>
+								{children}
```
ELanning, line 156: There is an important design consideration here. Namely, I chose to use `DatePicker.Footer` instead of creating `DatePicker.Footer` _and_ `DatePickerInput.Footer`.

The latter has the benefit of more encapsulation at the cost of more boilerplate. Because these components are so closely related, I didn't feel it worth the extra indirection.

components/date-picker/component.jsx
```diff
@@ -122,9 +132,12 @@ export class DatePicker extends Component {
 			validate,
 			minDate,
 			maxDate,
+			children,
 		} = this.props;
 		const { currentMonth, weeks } = this.state;
 
+		const footer = React.Children.toArray(children).filter(child => child.type === Footer);
```
ELanning, line 139: I saw there was a util that did stuff similar to this, eg:
https://github.com/Faithlife/styled-ui/blob/master/components/utils/get-config-props.js#L22

But it didn't seem necessary here.

ELanning: I'm not sure this utility will scale without a linter as there's already [inconsistencies popping up](https://github.com/Faithlife/styled-ui/blob/f750db8eda9ea52fbaa915f313456f8ad88203e1/components/image-well/component.jsx#L24). Which may point to this functionality not being worth "standardizing", but no objections to changing it.

ELanning: Thinking about it more; I wonder if it's better to view `getConfigChild` as a utility for pulling out child components from the other children. Rather than "the single source of truth for the [smart children pattern](https://www.reactguide.dev/#smart-children)".

More evidence the former is preferred- my usage does not even use the second return value:
```
const [footer] = getConfigChild(...);
```

korbinancell: I don't have an issue with the second value not being used, I think that is a common enough pattern. I do like the idea of a linter though.

korbinancell (approved):
Perhaps we should add a default of 8px padding to the footer component I leave that up to you since you have an actual use case in mind. Other than that LGTM.

ELanning:
I've found "helpful margin/padding" defaults are usually a hit or a miss. They can sometimes result in things like this (copied from an actual project).
```
	/* FormRenderer adds hardcoded margins. These offset it. */
	margin-bottom: -20px;
	margin-top: -6px;
```

korbinancell:
Adding it as a prop to the underlying Box that is overwritten by the prop spread would make this a non-issue, but I see how this could be annoying. :shipit:

components/date-picker/component.jsx
ELanning: Could be a good [checkr rule](https://marketplace.visualstudio.com/items?itemName=checkr.checkr). EG:
```
[
	function preferGetConfigChild({ underline, code }) {
		underline(
			code`$a = React.Children.toArray(children).filter(child =>child.type === $$$)`,
			"💬 Consider using `getConfigChild` when selecting child components.",
			"info");
	}
]
```

PR #524, opened 2022-01-07
Fix date-related tests.

MitchelMyersFl:


components/date-picker-input/__tests__/date-picker-input.test.jsx
```diff
@@ -25,7 +25,7 @@ describe('DatePeriodPicker', () => {
 			/>,
 		);
 
-		userEvent.type(screen.getByDisplayValue('August 26'), 'March 14');
+		userEvent.type(screen.getByDisplayValue(/August 26/), 'March 14');
```
MitchelMyersFl, line 28: The main thing is that the date is `2021-08-26` which now in 2022 renders as `August 26, 2021` and therefore is not found (not exact match). Using a regex is one way of letting us do a "contains" match. There may be other solutions. @korbinancell, thoughts?

korbinancell: Smart. That sounds good to me.

PatrickNausha: > There may be other solutions. @korbinancell, thoughts?

Faithlife uses [jest-date-mock](https://www.npmjs.com/package/jest-date-mock), e.g. https://git.faithlife.dev/Logos/Faithlife/blob/9c8ab1e5843b1de9c39667192ee61a9ed5da060f/src/Faithlife.Web/Scripts/tests/calendar-item.test.jsx to mock the current time. I've seen essentially this same bug come up so frequently I wonder if we should default to mocking the current time in every test.

korbinancell: This keeps coming back to bite us yea. Looking through I think I like jest-date-mock, we can take it a setup further and automatically set up the date to a deterministic time before each test it looks like.

PR #525, opened 2022-02-07
Use @babel/plugin-transform-runtime and usedExports.

PatrickNausha:
`@babel/plugin-transform-runtime` shrinks the `dist/main.js` bundle from 819KB to 757KB. `usedExports` reduces it further to 747KB and lets consumers optimize away another 5KB or more of code (`usedExports` leaves `unused harmony export` comments in the bundle that appear to help consumers' webpack builds eliminate some dead code).

This amounts to double digit savings in consumers.

styled-ui-shrinkage

Note: The GZIP savings %age win in consumers is greater than the %age win "819KB to 757KB." This may be a result of Babel's injected helper code previously being repeated many times, but then being minified. The minifier mangles the repeated verbatim helper code such that the repetition is not verbatim anymore and redundancy reduces slightly after minification.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -16,9 +16,12 @@
 				"ssr": true
 			}
 		],
-		["@babel/plugin-transform-runtime", {
-			"version": "^7.17.0"
-		}],
+		[
+			"@babel/plugin-transform-runtime",
+			{
+				"version": "^7.17.0"
+			}
+		],
 		"react-docgen",
 		"@babel/plugin-proposal-class-properties",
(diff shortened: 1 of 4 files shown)


.babelrc
```diff
@@ -16,6 +16,9 @@
 				"ssr": true
 			}
 		],
+		["@babel/plugin-transform-runtime", {
```
ddunkin, line 19: nit: formatting

ddunkin (approved):

korbinancell (approved):
:shipit:

PR #526, opened 2022-02-07
Updated popperjs/core dependency version to 2.11.2 CMRC-19517

SaltyPineapple:


korbinancell (approved):
LGTM. Looks like selects still work fine.

